PR #13, opened 2015-01-05
Adding new Cisco TelePresence banners

mdobrska-r7:
This work is related to this task: CONTENT-6924
recog_verify.rb tests pass for all 3 edited files.
Nexpose DTest passing as well.

### DIFF
--- a/xml/ftp_banners.xml
+++ b/xml/ftp_banners.xml
@@ -808,4 +808,17 @@ against these patterns to fingerprint FTP servers.
     <param pos="1" name="host.name"/>
     <param pos="2" name="service.version"/>
   </fingerprint>
+  <fingerprint pattern="^Welcome to the (?:Cisco )?(?:Tele[Pp]resence) ([a-zA-Z\s]*?) ((?:MSE )?\d+), version (\d+.\d+\(\d+.\d+\)).*?">
+    <description>Cisco TelePresence</description>
+    <example os.device = "AM GW" os.version="1.1(1.34)" hw.series="3610">Welcome to the Cisco TelePresence AM GW 3610, version 1.1(1.34) </example>
+    <example os.device = "IP GW" os.version="2.0(3.32)" hw.series="3510">Welcome to the Cisco TelePresence IP GW 3510, version 2.0(3.32) </example>
+    <example os.device = "MCU" os.version="4.3(2.18)" hw.series="4210">Welcome to the Cisco TelePresence MCU 4210, version 4.3(2.18) </example>
+    <example os.device = "MCU" os.version="4.3(1.68)" hw.series="MSE 8420">Welcome to the Cisco TelePresence MCU MSE 8420, version 4.3(1.68) </example>
+    <example os.device = "Server" os.version="2.1(1.18)" hw.series="7010">Welcome to the Telepresence Server 7010, version 2.1(1.18) </example>
+    <param pos="0" name="os.vendor" value="Cisco"/>
+    <param pos="0" name="os.product" value="TelePresence"/>
+    <param pos="1" name="os.device"/>
+    <param pos="2" name="hw.series"/>
+    <param pos="3" name="os.version"/>
+  </fingerprint>
 </fingerprints>
--- a/xml/ftp_banners.xml
+++ b/xml/ftp_banners.xml
@@ -808,13 +808,13 @@ against these patterns to fingerprint FTP servers.
     <param pos="1" name="host.name"/>
     <param pos="2" name="service.version"/>
   </fingerprint>
-  <fingerprint pattern="^Welcome to the (?:Cisco )?(?:Tele[Pp]resence) ([a-zA-Z\s]*?) ((?:MSE )?\d+), version (\d+.\d+\(\d+.\d+\)).*?">
+  <fingerprint pattern="^Welcome to the (?:Cisco )?(?:TelePresence) ([a-zA-Z\s]*?) ((?:MSE )?\d+), version (\d+.\d+\(\d+.\d+\)).*?" flags="REG_ICASE">
     <description>Cisco TelePresence</description>
-    <example os.device = "AM GW" os.version="1.1(1.34)" hw.series="3610">Welcome to the Cisco TelePresence AM GW 3610, version 1.1(1.34) </example>
-    <example os.device = "IP GW" os.version="2.0(3.32)" hw.series="3510">Welcome to the Cisco TelePresence IP GW 3510, version 2.0(3.32) </example>
-    <example os.device = "MCU" os.version="4.3(2.18)" hw.series="4210">Welcome to the Cisco TelePresence MCU 4210, version 4.3(2.18) </example>
-    <example os.device = "MCU" os.version="4.3(1.68)" hw.series="MSE 8420">Welcome to the Cisco TelePresence MCU MSE 8420, version 4.3(1.68) </example>
-    <example os.device = "Server" os.version="2.1(1.18)" hw.series="7010">Welcome to the Telepresence Server 7010, version 2.1(1.18) </example>
+    <example os.device="AM GW" os.version="1.1(1.34)" hw.series="3610">Welcome to the Cisco TelePresence AM GW 3610, version 1.1(1.34) </example>
+    <example os.device="IP GW" os.version="2.0(3.32)" hw.series="3510">Welcome to the Cisco TelePresence IP GW 3510, version 2.0(3.32) </example>
+    <example os.device="MCU" os.version="4.3(2.18)" hw.series="4210">Welcome to the Cisco TelePresence MCU 4210, version 4.3(2.18) </example>
+    <example os.device="MCU" os.version="4.3(1.68)" hw.series="MSE 8420">Welcome to the Cisco TelePresence MCU MSE 8420, version 4.3(1.68) </example>
+    <example os.device="Server" os.version="2.1(1.18)" hw.series="7010">Welcome to the Telepresence Server 7010, version 2.1(1.18) </example>
     <param pos="0" name="os.vendor" value="Cisco"/>
     <param pos="0" name="os.product" value="TelePresence"/>
     <param pos="1" name="os.device"/>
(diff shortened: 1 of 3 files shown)


xml/ftp_banners.xml
```diff
@@ -808,4 +808,17 @@ against these patterns to fingerprint FTP servers.
     <param pos="1" name="host.name"/>
     <param pos="2" name="service.version"/>
   </fingerprint>
+  <fingerprint pattern="^Welcome to the (?:Cisco )?(?:Tele[Pp]resence) ([a-zA-Z\s]*?) ((?:MSE )?\d+), version (\d+.\d+\(\d+.\d+\)).*?">
```
jhart-r7: Rather than [Pp] and [a-zA-Z], you can either make the entire fingerprint case insensitive (with a ```flags``` attribute). Alternatively, you can make just some parts of it case insensitive, like:

```
(?i:Telepresence)
```

xml/ftp_banners.xml
```diff
@@ -808,4 +808,17 @@ against these patterns to fingerprint FTP servers.
...
+    <description>Cisco TelePresence</description>
+    <example os.device = "AM GW" os.version="1.1(1.34)" hw.series="3610">Welcome to the Cisco TelePresence AM GW 3610, version 1.1(1.34) </example>
```
jhart-r7: Minor style nit, but no spaces before after the ```=```.

xml/sip_banners.xml
```diff
@@ -11,4 +10,15 @@ SIP Server header values are matched against these patterns to fingerprint SIP d
       <param pos="0" name="os.product" value="IOS"/>
       <param pos="1" name="os.version"/>
    </fingerprint>
+   <fingerprint pattern="^TANDBERG\/\d+ \(([a-zA-Z]+\d+(?:\.\d+)+)(?:.*)\)">
```
jhart-r7: No need to group that last .*.  This should be sufficient:

````
<fingerprint pattern="^TANDBERG/\d+ \(([a-zA-Z]+\d+(?:\.\d+)+).*\)">
```

jhart-r7:
I've made a few minor suggestions that I'd like to see fixed before landing, but otherwise this looks almost OK to land.

In the future, please read and follow https://github.com/rapid7/recog/blob/master/CONTRIBUTING.md -- specifically the part about forking and then doing your work in your own personal fork.  Doing it this way keeps the process similar to our other Github repositories (most notably metasploit) and makes it easier for others to add code to your PR to help it land.

xml/ftp_banners.xml
mdobrska-r7: (?i:Telepresence) will not work with the library used in Nexpose.

jhart-r7: OK -- please use ```flags="REG_ICASE"``` instead and file a Nexpose ticket to get that fixed.  While we (surprisingly) don't use that syntax right now, I can definitely see it being used at some point in the future.

xml/ftp_banners.xml
mdobrska-r7: I'll make these changes

xml/sip_banners.xml
mdobrska-r7: True, I'll fix that

xml/snmp_sysdescr.xml
```diff
@@ -1527,28 +1527,50 @@
...
+      <example os.version="TCNC5.1.0.280662" hw.series="SX20">Cisco Codec SoftW: TCNC5.1.0.280662 MCU: Cisco TelePresence SX20 Date: 2012-02-14 S/N: FTT161300A0 BootSW: Board: 101790-6 [A0]</example>
+      <example os.version="F8.1" hw.series="6000MXP">TANDBERG Codec SoftW: F8.1 PAL MCU: TANDBERG Profile 6000MXP Date: 2009-09-07 S/N: 25A46936 BootSW: Rev. 1.15, 2008-01-04 Board: 100670 rev. 07</example>
+      <example os.version="F9.1" hw.series="6000MXP">TANDBERG Codec SoftW: F9.1 NTSC MCU: TANDBERG 6000MXP PORTABLE Date: 2011-06-08 S/N: 25A07029 BootSW: Rev. 1.16, 2009-01-21 Board: 100670 rev. 07</example>
+      <example os.version="TC4.2.0.260857" hw.series="C40">TANDBERG Codec SoftW: TC4.2.0.260857 MCU: TANDBERG Codec C40 Date: 2011-07-11 S/N: B1AV02D00182 BootSW: Board: 101681-1 [04]</example>
+      <example os.version="TC4.2.1.265253" hw.series="EX60">TANDBERG Codec SoftW: TC4.2.1.265253 MCU: TANDBERG EX60 Date: 2011-09-06 S/N: A1AZ02E00212 BootSW: Board: 101620-6 [22]</example>
+      <example os.version="TE1.0.0.175435" hw.series="E20">TANDBERG Codec SoftW: TE1.0.0.175435 MCU: TANDBERG E20 Date: 2009-02-09 S/N: A1AA11B00658 BootSW: Board: 101390-6</example>
```
jhart-r7: To some extent you are just following what is already here, but why are we calling Tandberg Cisco now?  Yes, Tandberg is owned by Cisco as of 2010, but everything before that was absolutely a Tandberg product and not Cisco.  For devices never than 2010 there are obviously some things that aren't clear.

I think this is probably fine so long as every reference to anything Tandberg is updated (you missed a few)

mdobrska-r7: I'm not sure if I'm understanding what you mean  by 
'every reference to anything Tandberg is updated (you missed a few)'

Should I keep Cisco as vendor here? What did I miss?

jhart-r7: Sorry for the confusion.  I'll to clarify.  There are two points.

First is that as your PR stands right now, some Tandberg devices will be fingerprinted as Tandberg and others will be fingerprinted as Cisco.  There are two other fingerprints in this file that are for Tandberg devices and the vendor there is still Tandberg. In general, any time you are changing a vendor or product name, you must change all references.   There are probably some exceptions (for example, when, as part of an acquisition, the acquired product becomes distinctly different and isn't just a rebranding -- in that case it may make sense to call pre-acquisition products one thing and post-acquisition products another).

Secondly, in cases where the vendor name changes (as part of an acquisition, for example), this needs to be a conscious, product/consumer-wide change.  In other words, some devices previously fingerprinted as Tandberg will now be fingerprinted as Cisco.  If, for example, Metasploit and Nexpose have existing coverage that expects a Tandberg fingerprint, that coverage will stop working.  Fortunately, it doesn't look like there is any affected Tandberg coverage in Metasploit or Nexpose, so the change should be OK.

mdobrska-r7: Thanks for clarification. I'm intending to change the old Tandberg fingerprints os.vendor to Cisco and move them into this section of the file. Let me know if it's ok.

xml/snmp_sysdescr.xml
```diff
@@ -1527,28 +1527,65 @@
...
-      <example hw.product="MCU 4205">Cisco TelePresence MCU 4205</example>
-      <example hw.product="MCU MSE 8420">Cisco TelePresence MCU MSE 8420</example>
+      <example os.device = "MCU" hw.series="4205">Cisco TelePresence MCU 4205</example>
+      <example os.device = "MCU" hw.series="MSE 8420">Cisco TelePresence MCU MSE 8420</example>
+      <example os.device = "HD MCU" hw.series="4520">Cisco TelePresence HD MCU 4520</example>
       <param pos="0" name="os.certainty" value="0.85"/>
       <param pos="0" name="os.vendor" value="Cisco"/>
       <param pos="0" name="os.product" value="TelePresence"/>
-      <param pos="1" name="hw.product"/>
+      <param pos="1" name="os.device"/>
```
jhart-r7: Something isn't correct/consistent here -- in the fingerprint immediately before this one, you hardcode ```os.device``` to ```Codec```, but here, with very similar patterns you extract ```os.device``` from the banner, getting things like "MCU", etc.  This is unlikely by design, but if it is, please document why.

xml/snmp_sysdescr.xml
```diff
@@ -1527,28 +1527,65 @@
...
+      <param pos="2" name="hw.series"/>
    </fingerprint>
 
-   <fingerprint pattern="^Cisco TelePresence (Conductor|MCU (?:MSE )?\d+)">
+   <fingerprint pattern="^Cisco TelePresence ((?:HD )?MCU) ((?:MSE )?\d+)">
       <description>Cisco TelePresence</description>
-      <example hw.product="Conductor">Cisco TelePresence Conductor</example>
-      <example hw.product="MCU 4205">Cisco TelePresence MCU 4205</example>
-      <example hw.product="MCU MSE 8420">Cisco TelePresence MCU MSE 8420</example>
+      <example os.device = "MCU" hw.series="4205">Cisco TelePresence MCU 4205</example>
```
jhart-r7: Here and elsewhere, no spaces around the ```=``` in attribute values.

xml/snmp_sysdescr.xml
```diff
@@ -1527,28 +1527,65 @@
...
+      <param pos="0" name="os.vendor" value="Cisco"/>
+      <param pos="0" name="os.device" value="Web cam"/>
+   </fingerprint>
+
+   <fingerprint pattern="^TANDBERG MPS-MCU$">
+      <description>Tandberg videoconferencing bridge</description>
+      <example>TANDBERG MPS-MCU</example>
+      <param pos="0" name="os.vendor" value="Cisco"/>
+      <param pos="0" name="os.product" value="MPS-MCU"/>
+      <param pos="0" name="os.device" value="VOIP"/>
```
jhart-r7: Both in this fingerprint and the one before it, ```os.device``` is using a different pattern than what you are doing elsewhere.

xml/snmp_sysdescr.xml
mdobrska-r7: Sorry, about the inconsistency, I have a lot of banners from Sonar, but I'm not always sure what sort of devices they come from. For this fingerprint os.device should be Bridge. Will update.

xml/snmp_sysdescr.xml
mdobrska-r7: The first fingerprint should be consistent with other Codecs: 
os.vendor"Cisco", os.product="TelePresence", os.device="Codec"

The second one is not listed as TelePresence device as far as I can see so I'd rather leave it as it is.

PR #16, opened 2015-01-29
Add fingerprints for Python HTTP servers

inokii:
Fingerprints for two additional Python based HTTP servers: 1. Tornado Python web framework and 2. SimpleHTTP


hmoore-r7:
Thanks! We should be able to review and merge soon.

hmoore-r7:
Released in 1.0.11

PR #18, opened 2015-02-03
Move all capture validation to rspec

jhart-r7:
This simply takes the capture validation (too many captures, too few) logic from the functional recog code and moves it to the specs instead.  In doing this, it will flag 8 fingerprints as having incorrect groups (which I'll also fix in this PR but want to show that rspec will flag them)


hmoore-r7:
Test is failing correctly, merging =)

PR #30, opened 2015-02-18
Add initial MySQL banner support

TomSellers:
Add initial MySQL banner support

### DIFF
--- a/xml/mysql_banners.xml
+++ b/xml/mysql_banners.xml
@@ -865,6 +865,19 @@
     <param pos="0" name="os.version" value="10.04"/>
   </fingerprint>
 
+  <fingerprint pattern="^(\d{1,2}\.\d{1,3}\.\h{1,3})(?:-\d{1,2}\.\d{1,3}\.\h{1,4})?-MariaDB.+~hardy(?:-log)?$">
+    <example service.version="5.5.34">5.5.34-MariaDB-mariadb122~hardy-log</example>
+    <description>MariaDB MariaDB on Ubuntu 8.04 LTS (Hardy Heron)</description>
+    <param pos="1" name="service.version"/>
+    <param pos="0" name="service.vendor" value="MariaDB"/>
+    <param pos="0" name="service.family" value="MySQL"/>
+    <param pos="0" name="service.product" value="MariaDB"/>
+    <param pos="0" name="os.vendor" value="Ubuntu"/>
+    <param pos="0" name="os.family" value="Linux"/>
+    <param pos="0" name="os.product" value="Linux"/>
+    <param pos="0" name="os.version" value="8.04"/>
+  </fingerprint>
+
   <fingerprint pattern="^(\d{1,2}\.\d{1,3}\.\h{1,3})(?:-\d{1,2}\.\d{1,3}\.\h{1,4})?-[Mm]ariaDB.+~trusty-wsrep(?:-log)?$">
     <example service.version="5.5.5">5.5.5-10.1.1-MariaDB-1~trusty-wsrep-log</example>
     <description>MariaDB MariaDB with Galera Cluster on Ubuntu 14.04 (Trusty Tahr)</description>
@@ -877,6 +890,20 @@
     <param pos="0" name="os.family" value="Linux"/>
     <param pos="0" name="os.product" value="Linux"/>
     <param pos="0" name="os.version" value="14.04"/>
+  </fingerprint>
+
+  <fingerprint pattern="^(\d{1,2}\.\d{1,3}\.\h{1,3})(?:-\d{1,2}\.\d{1,3}\.\h{1,4})?-MariaDB.+~saucy-wsrep(?:-log)?$">
+    <example service.version="5.5.5">5.5.5-10.0.10-MariaDB-1~saucy-wsrep-log</example>
+    <description>MariaDB MariaDB with Galera Cluster on Ubuntu 13.10 (Saucy Salamander)</description>
+    <param pos="1" name="service.version"/>
+    <param pos="0" name="service.vendor" value="MariaDB"/>
+    <param pos="0" name="service.family" value="MySQL"/>
+    <param pos="0" name="service.product" value="MariaDB"/>
+    <param pos="0" name="service.edition" value="Galera Cluster"/>
+    <param pos="0" name="os.vendor" value="Ubuntu"/>
+    <param pos="0" name="os.family" value="Linux"/>
+    <param pos="0" name="os.product" value="Linux"/>
+    <param pos="0" name="os.version" value="13.10"/>
   </fingerprint>
 
   <fingerprint pattern="^(\d{1,2}\.\d{1,3}\.\h{1,3})(?:-\d{1,2}\.\d{1,3}\.\h{1,4})?-MariaDB.+~precise-wsrep(?:-log)?$">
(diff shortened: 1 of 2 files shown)


xml/mysql_banners.xml
```diff
@@ -0,0 +1,850 @@
...
+    <param pos="0" name="service.family" value="MySQL"/>
+    <param pos="0" name="service.product" value="MySQL"/>
+    <param pos="0" name="service.edition" value="Community Edition"/>
+    <param pos="0" name="os.vendor" value="Microsoft"/>
+    <param pos="0" name="os.device" value="General"/>
+    <param pos="0" name="os.family" value="Windows"/>
+    <param pos="0" name="os.product" value="Windows"/>
+  </fingerprint>
+
+  <fingerprint pattern="^(\d\.\d{1,2}\.\h{1,3})(?:-m\d)?(?:-rc)?(?:-alpha)?(?:-beta)?-[Cc]ommunity(?:-max)?(?:-debug)?(?:-log)?(?:-debug)?$">
```
jhart-r7: Here you are searching for Community or community, but in the previous one you are just looking for community.  IMO, make both of these (and perhaps all of them) case insensitive

xml/mysql_banners.xml
```diff
@@ -0,0 +1,850 @@
...
+    <param pos="0" name="service.family" value="MySQL"/>
+    <param pos="0" name="service.product" value="MySQL"/>
+    <param pos="0" name="service.edition" value="Community Edition"/>
+  </fingerprint>
+
+  <fingerprint pattern="^(\d\.\d{1,2}\.\h{1,3})(?:-alpha)?(?:-beta)?(?:-rc)?(?:-gamma)?(?:-max)?-nt(?:-max)?(?:-log)?$">
+    <example>5.0.18-nt</example>
+    <example>4.1.12a-nt-max-log</example>
+    <example>3.23.58-max-nt</example>
+    <example>5.1.6-alpha-nt-max</example>
```
jhart-r7: test version extraction

xml/mysql_banners.xml
```diff
@@ -0,0 +1,850 @@
...
+    <description>Oracle MySQL on Ubuntu</description>
+    <param pos="1" name="service.version"/>
+    <param pos="0" name="service.vendor" value="Oracle"/>
+    <param pos="0" name="service.family" value="MySQL"/>
+    <param pos="0" name="service.product" value="MySQL"/>
+    <param pos="0" name="os.vendor" value="Ubuntu"/>
+    <param pos="0" name="os.device" value="General"/>
+    <param pos="0" name="os.family" value="Linux"/>
+    <param pos="0" name="os.product" value="Linux"/>
+    <param pos="2" name="os.version"/>
```
jhart-r7: test both extractions here (and everywhere else it was missed)

xml/mysql_banners.xml
```diff
@@ -0,0 +1,850 @@
...
+    <example service.version="5.1.58">5.1.58-1ubuntu1</example>
+    <example service.version="4.0.24">4.0.24_Debian-10ubuntu2-log</example>
+    <example service.version="5.6.19">5.6.19-1~exp1ubuntu2-log</example>
+    <description>Oracle MySQL on Ubuntu where trailing digits aren't OS version</description>
+    <param pos="1" name="service.version"/>
+    <param pos="0" name="service.vendor" value="Oracle"/>
+    <param pos="0" name="service.family" value="MySQL"/>
+    <param pos="0" name="service.product" value="MySQL"/>
+    <param pos="0" name="os.vendor" value="Ubuntu"/>
+    <param pos="0" name="os.device" value="General"/>
```
jhart-r7: os.device is totally valid, but IMO there is little value in using it in mysql fingerprints unless you could somehow determine something more specific.

xml/mysql_banners.xml
```diff
@@ -0,0 +1,850 @@
...
+    <description>Oracle MySQL Commercial Edition with Named Pipes (Windows)</description>
+    <param pos="1" name="service.version"/>
+    <param pos="0" name="service.vendor" value="Oracle"/>
+    <param pos="0" name="service.family" value="MySQL"/>
+    <param pos="0" name="service.product" value="MySQL"/>
+    <param pos="0" name="service.edition" value="Commercial Edition"/>
+    <param pos="0" name="os.vendor" value="Microsoft"/>
+    <param pos="0" name="os.device" value="General"/>
+    <param pos="0" name="os.family" value="Windows"/>
+    <param pos="0" name="os.product" value="Windows"/>
```
jhart-r7: os.family alone is sufficient in this case.  If you could tell more specifically what version of Windows, great, by Windows alone as a product is too vague.

TomSellers: This was mimicing what was done n ftp_banners.xml

xml/mysql_banners.xml
TomSellers: If i'm not mistaken there was a limited number of instances of Community, something around 40 or in the data set.  I will try to mine the data and also see if there is a performance impact for doing case insensitivity.

jhart-r7:
This is looking pretty good.  To your pending commit comments:

- Community vs community metrics -- would be nice to know, but my suggestion for this wasn't necessarily that I think there may be a banner like this today, but that there may be one _some day_. 
- Base generic regex for MySQL versions not matched so far -- the trouble with this is that it is then hard for you to tell if you have a new/unmatched/unknown banner because this generic regex would catch all mysql versions you hadn't seen before and might want to handle differently for whatever reason.  I'm not totally opposed as I see some value in this. Perhaps post a sample of what you were thinking?
- fingerprint optimization based on sample data 
- reviewing use of trailing terminator for regex

xml/mysql_banners.xml
```diff
@@ -0,0 +1,812 @@
+<?xml version="1.0"?>
+<fingerprints>
```
jhart-r7: You should add a `matches` attribute here to indicate what it should be matching.  

That brings up something I think we discussed in email but I don't see addressed here.  This single file matches against two different types of data -- version banners and error messages.  IMO these should go in separate files and use different `match` attributes.  Then, anyone/anything using these files can use the right file depending on the type of message received.

TomSellers: Thanks, I'll address both as well as all of the outstanding issues. I hope to have the final PR for review by the end of this coming weekend.

TomSellers:
@jhart-r7

* I have left out the generic match.  I had been leaning towards leaving it out, but wasn't sure of the best fit for your use case. Your comments cleared that up for me.

* Fingerprint optimization was a reference to changing the order of the fingerprints with the goal of having the most common earlier in the file so that matches occurred faster.  The downside of this is that similar fingerprints would no longer be grouped which would complicate the effort of maintaining fingerprints.   With 1 exception, testing indicate that the benefit across the 5.2 million banner test set was not significant enough to change the FP order.  That change, in the mysql_errors, has been made.

* The review of the regex trailing terminator was at the request of HD.  I prefer to use the $ anchor where possible when I am fairly certain that I have all the string accounted for.  This reduces the chance of inadvertently matching and loosing an opportunity to gather more information about a target service.  
  * The end of line anchor needs to be added to the 'MariaDB MariaDB' fingerprint around line 595 and retested.  That is next on my list.  It will allow us to add a few fingerprints for MariaDB on certain flavors of Ubuntu and Debian.

Current match metrics against the 2015.02.01 dataset:

MySQL records           5,258,834
MySQL match count   5,255,535    99.937%
Unmatched:                       3,229

Additional metrics can be seen in this gits:
  https://gist.github.com/TomSellers/d1749e7b993a1270a526

TomSellers:
That last commit should say that removed 'a' fingerprint for Percona, There are still multiple fingerprints in place that are reliable for the related banners.

xml/mysql_banners.xml
```diff
@@ -0,0 +1,736 @@
+<?xml version="1.0"?>
+<fingerprints matches="mysql.banners">
+
+   <fingerprint pattern="^(\d\.\d{1,2}\.\h{1,3}(?:[.-]\d{1,2})?(?:[.-]\d{1})?)(?:-m\d{1,2})?(?:-rc)?(?:-alpha)?(?:-beta)?(?:-gamma)?(?:-?[Mm]ax)?(?:-rs)?(?:-modified)?(?:-debug)?(?:-log)?$">
```
jhart-r7: Here and everywhere else where you are trying to match the numeric part of the version, would it make sense to make this more flexible such that version parts of arbitrary length are matched?  For example, would 10.0.0 or 5.100.9999 ever be a valid version?

gschneider-r7: Not sure about MySQL proper, but MariaDB uses 10.x.x versions as they diverge from Oracle's implementation.

TomSellers: Perhaps, but isn't in our data set yet.  MariaDB does use 10.x.x and that is accounted for in the matches starting at line 568.  The 'MariaDB MariaDB on Debian Lenny' fingerprint is missing this.  

I can adjust the regex as you requested.  I keep falling back to my 'unknown service' mindset and forgetting that the assumption is that we already know its MySQL and just want to fingerprint it.

TomSellers:
@jhart-r7 This morning's updates contain the changes to the regex to better handle future strings such as 10.xxx.xxx.   There are also quite additions that identify the target's OS.

TomSellers:
@jhart-r7 @gschneider-r7  The previous is likely my last commit tweaking fingerprints. Anything that follows will be bug fixes for issues that someone discovers.

jhart-r7:
Thanks, @TomSellers!  I should be able to land this today.  My only last bit of feedback, which I'll address while landing, is to add some comments in these two files to indicate how these are to be used (IOW, what parts of the protocol are being fingerprinted).

PR #33, opened 2015-02-19
Set-Cookie fingerprint improvements

jhart-r7:
Adds an explicit fingerprint for HP System Management Homepage (SMH) as well as adds ignored fingerprints for common generic cookies that can't be used to pinpoint anything.


jhart-r7:
@hmoore-r7 , done

Issue #20, opened 2015-02-05
Strict or Flexible Version Matching?


inokii:
The HTTP server header fingerprints file (xml/http_servers.xml) currently contains 19 patterns that use `((?:\d+\.)+\d+` to match the service version. A successful match of the pattern requires, at a minimum, a major.minor style version number. A major only style version number will fail to match. For example, using the CherryPy pattern we can see the version portion of the pattern requires, at a minimum, a version number that contains two digits separated by a period.

```
$ echo 'CherryPy/3.1.2' | bin/recog_match.rb xml/http_servers.xml -
MATCH: {"service.vendor"=>"CherryPy", "service.product"=>"CherryPy", "service.family"=>"CherryPy", "service.version"=>"3.1.2", "data"=>"CherryPy/3.1.2"}

$ echo 'CherryPy/3.1' | bin/recog_match.rb xml/http_servers.xml -
MATCH: {"matched"=>"Web server component of CherryPy web application framework.", "service.vendor"=>"CherryPy", "service.product"=>"CherryPy", "service.family"=>"CherryPy", "service.version"=>"3.1", "data"=>"CherryPy/3.1"}

$ echo 'CherryPy/3' | bin/recog_match.rb xml/http_servers.xml -
FAIL: CherryPy/3
```

I have not fully investigated all of the HTTP server header fingerprints or other fingerprint files, but I think this is an important topic to discuss and determine a best practice for the project. The patterns I introduced in #16 use `((?:\d+\.)*\d+)` which allows for flexible version number matching. I prefer more flexible patterns as eases maintenance of the fingerprints to some degree.

jhart-r7:
In general I'd agree with what you've suggested.  If it is possible to write a fewer number of fingerprints (ideally just 1) for each product with careful grouping, then absolutely do it if you have ```<examples>``` showing that the regex works and extractions happen as expected.

jhart-r7:
Also, I think https://github.com/rapid7/recog/blob/master/README.md could/should be updated to describe our preferences around the regular expressions.   @inokii, if you'd like to take a pass at updating it via a PR that would be great!

inokii:
Added Best Practices section to README.md per your suggestion. Opened PR #34. I'm hoping to find some free time over the next week to make updates to the 19 fingerprints I mentioned, and will submit PR after testing.

jhart-r7:
Landed #34.  Thanks for the contributions!

PR #34, opened 2015-02-19
Added Best Practices section

inokii:
The Best Practices section notes the importance of simplifiying fingerprints and flexible version number matching. Change made per suggestion of @jhart-r7 in #20.


jhart-r7:
Excellent.  This looks like a great start.  

As an aside, I was thinking earlier that over time the documentation for the XML files themselves is going to get a bit large and keeping it crammed in the README may be too much.  We can cross that bridge later.

PR #35, opened 2015-02-25
EOL 1.9.3 support, fixes #26

### LINKED ISSUE #26: EOL 1.9.3 support
2/23/2015, 1.9.3 will be EOL.  Metasploit is also making this change (or maybe they did already -- I don't know for sure).

Recog should drop 1.9.3 support too.

jhart-r7:
Remove 1.9.3 from Travis
Set minimum ruby version for the gem to 2.1+
Bump recog version

I can still build the gem on 1.9.3 but it won't install:

```
$  gem build recog.gemspec && gem install recog-1.0.18.gem 
  Successfully built RubyGem
  Name: recog
  Version: 1.0.18
  File: recog-1.0.18.gem
ERROR:  Error installing recog-1.0.18.gem:
	recog requires Ruby version >= 2.1.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: ruby
 cache: bundler
 rvm:
   - 2.1.5
-  - 1.9.3
   - jruby
 before_install:
   - "echo 'gem: --no-ri --no-rdoc' > ~/.gemrc"
--- a/lib/recog/version.rb
+++ b/lib/recog/version.rb
@@ -1,3 +1,3 @@
 module Recog
-  VERSION = '1.0.18'
+  VERSION = '2.0.0'
 end
--- a/lib/recog/version.rb
+++ b/lib/recog/version.rb
@@ -1,3 +1,3 @@
 module Recog
-  VERSION = '1.0.17'
+  VERSION = '1.0.18'
 end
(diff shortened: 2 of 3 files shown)


gschneider-r7:
Shouldn't this be more than a "patch" version change? Anyone relying on semantic versioning will be surprised if they don't have Ruby 2.1+ installed and they pull this in. Should at least be a "minor" version increase in my opinion.

limhoff-r7:
Post 1.0.0, an incompatible change requires incrementing the MAJOR version number, so that means the correct version for this change is 2.0.0.

limhoff-r7:
For reference: http://semver.org/

jhart-r7:
Agreed -- 2.0.0 is more appropriate.

jhart-r7:
AFAIK, this is good to go.  If there are other reasons we shouldn't land this now, please speak up.

hmoore-r7:
No problem here, we can start bumping Gemfile requirements once this is released

jhart-r7:
@hmoore-r7, could you land this?  I'll follow up with the Nexpose side of the house to ensure that the new major version is fetched if you could coordinate similarly with Metasploit.  Thanks!

hmoore-r7:
Sure thing!

PR #36, opened 2015-02-25
HTTP banners for TippingPoint and HP OpenView

alynn71:
Add HP TippingPoint and OpenView HTTP server banners

### DIFF
--- a/xml/http_servers.xml
+++ b/xml/http_servers.xml
@@ -1891,9 +1891,9 @@
 
    <fingerprint pattern = "^(?:BBC \d+\.\d+\.\d+\.?\d*; )?(?:com.hp.openview.)?[c|C]oda (\d+\.\d+\.\d+\.?\d*)$">
       <description>HP Openview Coda (Communications Daemon)</description>
-      <example>com.hp.openview.Coda 0.0.1</example>
-      <example>BBC 05.20.050; coda 10.00.123</example>
-      <example>BBC 2.6.0.7; com.hp.openview.Coda 0.0.1</example>
+      <example service.component.version="0.0.1">com.hp.openview.Coda 0.0.1</example>
+      <example service.component.version="10.00.123">BBC 05.20.050; coda 10.00.123</example>
+      <example service.component.version="0.0.1">BBC 2.6.0.7; com.hp.openview.Coda 0.0.1</example>
       <param pos="0" name="service.vendor" value="HP"/>
       <param pos="0" name="service.family" value="OpenView"/>
       <param pos="0" name="service.component.vendor" value="HP"/>
@@ -1903,8 +1903,8 @@
    </fingerprint>
 
    <fingerprint pattern="^BBC \d+\.\d+\.\d+\.?\d*; ovbbcrcp (\d+\.\d+\.\d+\.?\d*)$">
-      <example>BBC 11.00.044; ovbbcrcp 11.00.044</example>
       <description>OpenView Reverse Channel Proxy (RCP)</description>
+      <example service.component.version="11.00.044">BBC 11.00.044; ovbbcrcp 11.00.044</example>
       <param pos="0" name="service.vendor" value="HP"/>
       <param pos="0" name="service.family" value="OpenView"/>
       <param pos="0" name="service.component.vendor" value="HP"/>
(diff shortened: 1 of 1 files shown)


xml/http_servers.xml
```diff
@@ -1907,6 +1907,27 @@
       <param pos="1" name="service.version"/>
    </fingerprint>
 
+   <fingerprint pattern="^BBC (\d+\.\d+\.\d+); ovbbccb \d+\.\d+\.\d+$">
+      <example>BBC 06.00.083; ovbbccb 06.00.083</example>
+      <example>BBC 06.20.050; ovbbccb 06.20.050</example>
+      <example>BBC 11.10.035; ovbbccb 11.10.035</example>
+      <description>OpenView ovbbccb </description>
+      <param pos="0" name="service.vendor" value="HP"/>
+      <param pos="0" name="service.product" value="OpenView ovbbccb"/>
```
jhart-r7: Is it correct and important to be calling this ovbbccb?  I have a similar change that I haven't put up for a PR yet that looks like:

```
+   <!-- fingerprints various services exposed by HP OpenView components.  For
+        now, the versions and components are not extracted because it is
+        unclear if these are relevant/accurate/useful.
+
+        BBC == Black Box Communications
+        CB == Communications Broker
+        CODA == COmmunications DAemon
+        LLB == Local Location Broker
+        RCP == Reverse Channel Proxy
+   -->
+   <fingerprint pattern = "^BBC (?:[\d\.]+); (?:\S+) (?:[\d\.]+|unknown version)$">
+      <description>HP Openview</description>
+      <example>BBC 06.20.101; ovbbccb 06.20.101</example>
+      <example>BBC 11.13.007; ovbbccb unknown version</example>
+      <example>BBC 05.20.050; coda 10.00.123</example>
+      <example>BBC 2.6.0.7; com.hp.openview.Coda 0.0.1</example>
+      <example>BBC 2.6.0.7; com.hp.openview.bbc.LLBServer 2.6.0.7</example>
+      <example>BBC 11.00.044; ovbbcrcp 11.00.044</example>
+      <param pos="0" name="service.vendor" value="HP"/>
+      <param pos="0" name="service.family" value="OpenView"/>
+      <param pos="0" name="service.product" value="OpenView"/>
    </fingerprint>
```

If splitting it out by the different part makes sense (based on what you did here, it does), then should we also extract the product specific version?  IOW, there are two versions in this string which may not be the same -- is one the OpenView version and the other the specific component's version?

alynn71: Based on your examples I would interpret the first as version number as the BBC version and the second as the component version. I think we could accurately fingerprint these components but probably a specific product version.

 OpenView is more of a product family than something we can tie a version to.  For example: OpenView Operations Agent and Openview Performance Agent can both utilize these components which they call  subagents but we cannot determine the which of these agents is installed based on the banners since it is only the subagent versions that are identified. 

I'm not sure how useful subagent details alone would be without the product. I could add a parameter for BBC version to my PR or we could just use your method.

jhart-r7: Using just two examples:

```
<example>BBC 06.20.101; ovbbccb 06.20.101</example>
<example>BBC 05.20.050; coda 10.00.123</example>
```

Based on what you've said, I think we should have one fingerprint per component (CB, CODA, LLB, RCP), and for each we'd assert:

```
<param pos="0" name="service.vendor" value="HP"/>
<param pos="0" name="service.family" value="OpenView"/>
<param pos="0" name="service.component.vendor" value="HP"/>
<param pos="0" name="service.component.family" value="OpenView"/>
<param pos="0" name="service.component.product" value="<That component name>"/>
<param pos="1" name="service.component.version"/>
```

alynn71: Sounds good. I can add them. It looks like there are already individual fingerprints for LBB and Coda that don't include version that can be tweaked to fit this format

xml/http_servers.xml
```diff
@@ -1889,22 +1889,73 @@
...
+   </fingerprint>
+
+   <fingerprint pattern="^UOS$">
+      <example>UOS</example>
+      <description>HTTP Server that appears unique to Managment Console on HP TippingPoint IPS Devices</description>
+      <param pos="0" name="service.vendor" value="HP"/>
+      <param pos="0" name="service.product" value="HTTP"/>
+      <param pos="0" name="service.family" value="TippingPoint"/>
+      <param pos="0" name="os.vendor" value="HP"/>
+      <param pos="0" name="os.family" value="TippingPoint"/>
```
jhart-r7: I am assuming you didn't put a product here because you can't know any more than "yes, this is a TippingPoint device of some sort".  In that case, maybe adding an {{os.device}} of {{IPS}} might be good.  I'll get that and other cleanups to you in a PR in a moment.

jhart-r7:
@alynn71 -- I made a few small cleanups to your PR in https://github.com/alynn71/recog/issues/1

If it all looks good to you, merge it and the changes will show up here for landing by me.

jhart-r7:
Thanks for the contribution, @alynn71.  I've landed this and released 1.0.18.

PR #39, opened 2015-03-06
Add Comware SSH banner

alynn71:
added Comware SSH banner and successfully tested in Nexpose

### DIFF
--- a/xml/ssh_banners.xml
+++ b/xml/ssh_banners.xml
@@ -777,7 +777,18 @@ fingerprint SSH servers.
       <param pos="0" name="os.family" value="Windows"/>
       <param pos="0" name="os.product" value="Windows"/>
    </fingerprint>
-
+   
+   <fingerprint pattern="^Comware-(\d+\.?\d*\.?\d*)$">
+      <description>SSH on H3C Comware</description>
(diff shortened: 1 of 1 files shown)


xml/ssh_banners.xml
```diff
@@ -777,7 +777,18 @@ fingerprint SSH servers.
...
       <param pos="0" name="os.product" value="Windows"/>
    </fingerprint>
-
+   
+   <fingerprint pattern="^Comware-(\d+\.?\d*\.?\d*)$">
+      <description>SSH on H3C Comware</description>
+      <example os.version="5.20.105">Comware-5.20.105</example>
+      <example os.version="5.20">Comware-5.20</example>
+      <param pos="0" name="service.vendor" value="H3C"/>
+      <param pos="0" name="service.product" value="SSH"/>
```
jhart-r7: I'll leave this here, but in general we try not to list protocols where it should really be a product.  Put another way, unless we know for sure that H3C makes their own SSH server (rather than, say, just rebuilding OpenSSH), then we might not want to put SSH here.  Based on my quick look at what H3C does, though, keeping SSH here makes more sense than not.

jhart-r7:
Thanks again, @alynn71 .

PR #43, opened 2015-03-19
Comware and OpenVMS ssh

alynn71:
I wasn't sure what would be the best way to classify the OpenVMS SSH service versions.  According to the documentation the SSH server is based on SSH2 software from SSH Communication Security.
Is using this as the service.component.version appropriate? Would it be better to use either the SSH Secure Shell version or the TCP/IP Services for openVMS version as the service.version parameter instead?

### DIFF
--- a/xml/ssh_banners.xml
+++ b/xml/ssh_banners.xml
@@ -790,7 +790,7 @@ fingerprint SSH servers.
       <param pos="1" name="os.version"/>
    </fingerprint>
 
-   <fingerprint pattern="^(\d\.\d\.\d) SSH Secure Shell OpenVMS V\d\.\d$">
+   <fingerprint pattern="^(\d\.\d+\.\d+) SSH Secure Shell OpenVMS V\d\.\d$">
       <description>SSH for OpenVMS </description>
       <example service.component.version="3.2.0">3.2.0 SSH Secure Shell OpenVMS V5.5</example>
       <!--V5.5 refers to TCP/IP Services for OpenVMS version -->
@@ -807,7 +807,7 @@ fingerprint SSH servers.
       <param pos="0" name="os.certainty" value="0.75"/>
    </fingerprint>
 
-   <fingerprint pattern="^(\d\.\d\.\d) SSH (?:Secure Shell )?OpenVMS V\d\.\d VMS_sftp_version (\d)$">
+   <fingerprint pattern="^(\d\.\d+\.\d+) SSH (?:Secure Shell )?OpenVMS V\d\.\d VMS_sftp_version (\d)$">
       <description>SSH for OpenVMS sftp</description>
       <example service.component.version="3.2.0" service.version="3">3.2.0 SSH Secure Shell OpenVMS V5.5 VMS_sftp_version 3</example>
       <example service.component.version="3.2.0" service.version="3">3.2.0 SSH OpenVMS V5.5 VMS_sftp_version 3</example>


xml/ssh_banners.xml
```diff
@@ -778,10 +778,11 @@ fingerprint SSH servers.
       <param pos="0" name="os.product" value="Windows"/>
    </fingerprint>
    
-   <fingerprint pattern="^Comware-(\d+\.?\d*\.?\d*)$">
+   <fingerprint pattern="^(?:SSH-[1|2]\.\d{1,2}-)?Comware-(\d+\.?\d*\.?\d*)$">
```
jhart-r7: Odd... do the real banners include double SSH- in them?  This file (for better or worse) is meant to be matched against everything after the SSH-<version>. IOW, if the full banner was:

```
SSH-2.0-OpenSSH_5.9p1 Debian-5ubuntu1.4
```

What recog would be matching against is:

```
OpenSSH_5.9p1 Debian-5ubuntu1.4
```

alynn71: I think I see what I did. I pulled some banner examples from the console which displays the full banner and others from scan logs which display just display the match. I'll test it and resubmit.

xml/pop_banners.xml
```diff
@@ -225,6 +225,20 @@ matched against these patterns to fingerprint POP3 servers.
       <param pos="0" name="os.certainty" value="0.5"/>
    </fingerprint>
 
+   <fingerprint pattern="^TCPIP POP server V\d\.\d\S-\S{3}, OpenVMS V(\d\.\d-\d) .*$">
+      <description>TCP/IP Services for OpenVMS POP server</description>
+      <example os.version="7.3-2">TCPIP POP server V5.4J-15A, OpenVMS V7.3-2 Alpha at example.com, up since 2015-02-12 08:44:53 20400434.2</example>
```
jhart-r7: Everything after the version -- is that consistently there?  If so, you should extract the hostname as `host.name` or domain as `host.domain` depending on what example.com is actually representing.

xml/ssh_banners.xml
```diff
@@ -789,6 +789,43 @@ fingerprint SSH servers.
...
+      <param pos="0" name="os.vendor" value="HP"/>
+      <param pos="0" name="os.device" value="General"/>
+      <param pos="0" name="os.family" value="OpenVMS"/>
+      <param pos="0" name="os.certainty" value="0.75"/>
+   </fingerprint>
+
+   <fingerprint pattern="^(\d\.\d\.\d) SSH (?:Secure Shell )?OpenVMS V\d\.\d VMS_sftp_version (\d)$">
+      <description>SSH for OpenVMS sftp</description>
+      <example service.component.version="3.2.0">3.2.0 SSH Secure Shell OpenVMS V5.5 VMS_sftp_version 3</example>
+      <example service.component.version="3.2.0" >3.2.0 SSH OpenVMS V5.5 VMS_sftp_version 3</example>
```
jhart-r7: Minor, but you may as well test the extraction of service.version

jhart-r7:
This looks good.  I can make my minor suggested improvements when landing.

xml/ssh_banners.xml
```diff
@@ -789,6 +789,43 @@ fingerprint SSH servers.
...
+      <param pos="0" name="service.vendor" value="HP"/>
+      <param pos="0" name="service.family" value="OpenVMS"/>
+      <param pos="0" name="service.product" value="SSH Server"/>
+      <param pos="0" name="os.vendor" value="HP"/>
+      <param pos="0" name="os.device" value="General"/>
+      <param pos="0" name="os.family" value="OpenVMS"/>
+      <param pos="0" name="os.certainty" value="0.75"/>
+   </fingerprint>
+
+   <fingerprint pattern="^(\d\.\d\.\d) SSH (?:Secure Shell )?OpenVMS V\d\.\d VMS_sftp_version (\d)$">
```
jhart-r7: Here and in the others you've added, it may make sense to make the version regular expressions a little more flexible unless there is good reason not to.  For example, this won't match if the component version is 3.11.0.

jhart-r7:
I've addressed most of my feedback in https://github.com/alynn71/recog/pull/3 -- it would be nice to address the version regex too.

PR #45, opened 2015-04-06
Correct FreeBSD SNMP fingerprint

### LINKED ISSUE #44: FreeBSD SNMP banners incorrectly assert Linux fingerprints; regex is too loose
https://github.com/rapid7/recog/blob/master/xml/snmp_sysdescr.xml#L2489-2512

Those should be FreeBSD, not Linux.  

Also, that second fingerprint is just too loose -- while it is likely correct that anything beginning with FreeBSD is in fact FreeBSD, there is almost certainly more we could get from the fingerprint provided this overly loose regex wasn't there and the match didn't occur.

jhart-r7:
This fixes #44.  `rspec` passes before and after this change.

### DIFF
--- a/xml/snmp_sysdescr.xml
+++ b/xml/snmp_sysdescr.xml
@@ -2502,15 +2502,6 @@ Copyright (c) 1995-2005 by Cisco Systems
       <param pos="3" name="os.arch"/>
    </fingerprint>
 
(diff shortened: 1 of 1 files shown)


jhart-r7:
Restarting the build.  Looks like travis was wedged during `bundle install`

Issue #47, opened 2015-04-21
warning pops up regularly from recog in the context of metasploit framework


bcook-r7:
This warning has begun regularly occurring within metasploit framework's msfconsole:

recog-1.0.27/lib/recog/fingerprint/regexp_factory.rb:33: warning: nested repeat operator '+' and '?' was replaced with '*' in regular expression

It seems to appear randomly (not necessarily in response to any particular action being done in framework), so it is not clear what to do to reproduce it yet.

bcook-r7:
Running the specs, I see the same warning. It appears that this pattern from ./xml/mysql_banners.xml causes the warning to popup, though there may be others:

```
<fingerprint pattern="^(\d{1,2}\.\d{1,3}\.\h{1,3})(?:-\d{1,2}\.\d{1,3}\.\h{1,4})?-MariaDB.+~sid-wsrep(?:-log)?$">
```

jhart-r7:
I (unfortunately) admit that I've seen those too and didn't do anything. Any idea which `+` or `?` in particular is the problem?  In the provided example they seem fine.

themetric:
Haven't figured out the one above yet, but found another one that popped up from running the specs: 
`/^(\d{1,2}\.\d{1,3}\.\h{1,3})(?:[^+]+)?\+tld\d/`
and the problem portion seems to be `[^+]+` which appears in many of the mysql patterns.  I think the redundancy is the double `+` which is saying "one or more of, one or more of."  If you remove the second `+` the warning goes away.  Haven't tested yet whether all the matches still work after a change like that.

bcook-r7:
For /^(\d{1,2}\.\d{1,3}\.\h{1,3})(?:[^+]+)?\+tld\d/ removing the second + causes it to fail the first test case. I think I tried something like /^(\d{1,2}\.\d{1,3}\.\h{1,3})(.*)tld\d/ which IIRC seemed to match the version string in both test cases, but caused more results than desired to come back.

pdeardorff-r7:
[...]

PR #48, opened 2015-04-30
Change redundant ? to * fixes rapid7/recog#47

pdeardorff-r7:
Change to a regex that was causing this warning 
`recog-1.0.27/lib/recog/fingerprint/regexp_factory.rb:33: warning: nested repeat operator '+' and '?' was replaced with '*' in regular expression`

#47

### DIFF
--- a/xml/mysql_banners.xml
+++ b/xml/mysql_banners.xml
@@ -973,7 +973,7 @@
     <param pos="0" name="os.version" value="6.0"/>
   </fingerprint>
 
-  <fingerprint pattern="^(\d{1,2}\.\d{1,3}\.\h{1,3})(?:-\d{1,2}\.\d{1,3}\.\h{1,4})?-MariaDB.+~sid-wsrep(?:-log)?$">
+  <fingerprint pattern="^(\d{1,2}\.\d{1,3}\.\h{1,3})(-\d{1,2}\.\d{1,3}\.\h{1,4})*-MariaDB.+~sid-wsrep(-log)*$">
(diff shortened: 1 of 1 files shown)


xml/mysql_banners.xml
```diff
@@ -1071,7 +1071,7 @@
     <param pos="0" name="service.product" value="MySQL"/>
   </fingerprint>
 
-  <fingerprint pattern="^(\d{1,2}\.\d{1,3}\.\h{1,3})(?:[^+]+)?\+tld\d">
+  <fingerprint pattern="^(\d{1,2}\.\d{1,3}\.\h{1,3})([^+]+)?\+tld\d">
```
jhart-r7: Removing the `?:` breaks the unit tests. The unit tests require that the number of capturing regular expressions match the number of uses of those extractions in the params.  This is because in the past we've had problems with over capturing causing memory problems in extreme cases (`(.*)$`, for example), or under capturing.

pdeardorff-r7: Ah ok, I was worried about the captures, but didn't quite know how the change would affect them.  I'll keep playing with different ways of doing it and make sure to run the unit tests as well as the verify script.  I want to try and find a way of determining what the resulting regexp is after the replacement so we can just use that and the warning will go away.

pdeardorff-r7:
Should be good to go now, I added back the `?:` so it won't capture the subsequent groups.  Out of curiosity I found the warning is coming from here: https://github.com/ruby/ruby/blob/trunk/regparse.c#L5420 but I couldn't find out how it's able to remove the redundancy and replace with `*`.

PR #50, opened 2015-05-01
Correct generic AIX / HP-UX FTP fingerprint

### LINKED ISSUE #49: HP-UX and AIX use identical FTP banners; recog incorrectly asserts AIX fingerprint
This is from a known HP-UX system:

```
host.example.com FTP server (Version 1.1.214.4 Wed Aug 23 03:38:25 GMT 2000) ready.
```

Unfortunately it is identical in format to one found on an AIX system:

```
host.example.com FTP server (Version 4.1 Sat Sep 7 14:31:53 CDT 2002) ready.
```

It isn't currently clear what the product here actually is, but asserting an OS fingerprint from this is definitely not correct.

jhart-r7:
Fixes #49

### DIFF
--- a/bin/recog_match
+++ b/bin/recog_match
@@ -9,12 +9,12 @@ require 'recog/matcher_factory'
 options = OpenStruct.new(color: false, detail: false, fail_fast: false)
 
 option_parser = OptionParser.new do |opts|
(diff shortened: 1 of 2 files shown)


hmoore-r7:
Agree that "-" for stdin is more common, but making this use stdin by default is sane too.

PR #51, opened 2015-05-08
Initial fingerprints for mDNS: Avahi, Mac OS X, AirPort

gwiseman-r7:
This PR contains 2 new XML fingerprint files which each correspond to TXT records that can be retrieved from mDNS endpoints.

The TXT record for _device-info._tcp.local. (which sometimes needs the host name prepended for the query) may contain information that can be used to fingerprint certain systems, in particular Mac OS X.

The TXT record for _workstation._tcp.local. (ditto w/r/t prepending the hostname) may contain information that can be used to fingerprint the mDNS software, in particular Avahi.

Note that due to the nature of mDNS the service may need to be queried from the same subnet as the host.

### DIFF
--- a/xml/mdns_device-info_txt.xml
+++ b/xml/mdns_device-info_txt.xml
@@ -1,0 +1,134 @@
+<?xml version="1.0"?>
+<!--
+  Fingerprint definitions that are matched against the string values in the
+  TXT record returned by an mDNS endpoint for the '_device-info._tcp.local.'
+  domain name.
+-->
+<fingerprints matches="mdns.device-info.txt">
+
+   <!--
+     OS X versions:
+     The number specified after osxvers= is equivalent to the major
+     version of OS X plus 4. E.g. osxvers=13 means OS X 10.9
+   -->
+   <fingerprint pattern="^osxvers=14">
+      <description>OS X 10.10 (Yosemite)</description>
+      <example>osxvers=14</example>
+      <param pos="0" name="os.certainty" value="0.85"/>
+      <param pos="0" name="os.vendor" value="Apple"/>
+      <param pos="0" name="os.family" value="Mac OS X"/>
+      <param pos="0" name="os.product" value="Mac OS X"/>
+      <param pos="0" name="os.version" value="10.10"/>
+   </fingerprint>
+
+   <fingerprint pattern="^osxvers=13">
+      <description>OS X 10.9 (Mavericks)</description>
+      <example>osxvers=13</example>
+      <param pos="0" name="os.certainty" value="0.85"/>
+      <param pos="0" name="os.vendor" value="Apple"/>
+      <param pos="0" name="os.family" value="Mac OS X"/>
+      <param pos="0" name="os.product" value="Mac OS X"/>
+      <param pos="0" name="os.version" value="10.9"/>
+   </fingerprint>
+
+   <fingerprint pattern="^osxvers=12">
+      <description>OS X 10.8 (Mountain Lion)</description>
+      <example>osxvers=12</example>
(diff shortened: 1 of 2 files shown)


xml/mdns_device-info_txt.xml
```diff
@@ -0,0 +1,134 @@
+<?xml version="1.0"?>
+<!--
+  Fingerprint definitions that are matched against the string values in the
+  TXT record returned by an mDNS endpoint for the '_device-info._tcp.local.'
+  domain name.
```
jhart-r7: Here and in the other file, it may be helpful to re-mention the fact that you may need to prepend the host name.

xml/mdns_device-info_txt.xml
```diff
@@ -0,0 +1,134 @@
...
+      <param pos="0" name="os.family" value="Mac OS X"/>
+      <param pos="0" name="os.product" value="Mac OS X"/>
+      <param pos="0" name="os.version" value="10.0"/>
+   </fingerprint>
+
+
+   <fingerprint pattern="^model=AirPort.*$">
+      <description>AirPort</description>
+      <!-- TODO: example -->
+      <param pos="0" name="os.certainty" value="0.8"/>
```
jhart-r7: Why 0.8?  For that matter, why specify a certainty for any of these, especially if for the {{_workstation_txt}} records you aren't specifying a certainty at all.

xml/mdns_device-info_txt.xml
```diff
@@ -0,0 +1,134 @@
...
+      <param pos="0" name="os.vendor" value="Apple"/>
+      <param pos="0" name="os.family" value="Mac OS X"/>
+      <param pos="0" name="os.product" value="Mac OS X"/>
+      <param pos="0" name="os.version" value="10.0"/>
+   </fingerprint>
+
+
+   <fingerprint pattern="^model=AirPort.*$">
+      <description>AirPort</description>
+      <!-- TODO: example -->
```
jhart-r7:
```
model=AirPort5,114
```

xml/mdns_device-info_txt.xml
gwiseman-r7: The Avahi service fingerprint (from workstation TXT) has a certainty of 1.0. The others I admit are a bit arbitrary; they just need to be higher than what Nexpose uses for TCP/IP stack fingerprinting. Also, they came from your original PoC ;)

xml/mdns_device-info_txt.xml
gwiseman-r7: Thank you

xml/mdns_device-info_txt.xml
gwiseman-r7: Done

jhart-r7:
This looks almost ready to land.  

Regarding the certainties, the fact that this Recog content will eventually be used in Nexpose (and that there is that IP stack fingerprinting gotcha) is somewhat irrelevant to this and most Recog PRs in general.  But, since you mentioned it, in Nexpose it will be up to the protocol helper for mDNS to specify what the default OS and service fingerprint certainties should be.  That is effectively what you are doing here, the only wrinkle being the certainty-less AirPort fingerprint and the 1.0 Avahi fingerprint.  

My suggestion?  Remove the certainties from this file entirely, and in your mDNS protocol helper, define the default certainty for OS and service fingerprints obtained from each of the respective records.

gwiseman-r7:
Thanks Jon. What you say regarding certainties makes sense; there should be more, ahem, certainty around what the values mean. I've removed them from the fingerprints. Please let me know if there's anything else necessary from my end to land this.

PR #52, opened 2015-05-14
Update SNMP regex for Extreme Networks XOS v15

gschneider-r7:
The new example banner contains some hardware information: X460 series 24-port switch.

I have no idea what else this might look like for other hardware, so it's very likely this will still miss some banners. This was just one banner found to be missing a match recently. I may have more examples coming up to improve this further.

Test:
```shell
$ bin/recog_match xml/snmp_sysdescr.xml xos15.txt
MATCH: {"matched"=>"Extreme Networks Switches", 
"os.vendor"=>"Extreme Networks", 
"os.device"=>"Switch", 
"os.product"=>"XOS", 
"os.version"=>"15.6.1.4", 
"data"=>"ExtremeXOS (X460-24x) version 15.6.1.4 v1561b4 by release-manager on Tue Sep 30 14:06:19 EDT 2014"}
```

cc @jhart-r7

### DIFF
--- a/xml/snmp_sysdescr.xml
+++ b/xml/snmp_sysdescr.xml
@@ -2314,7 +2314,7 @@ Copyright (c) 1995-2005 by Cisco Systems
    =======================================================================-->
 
    <!-- Sneaky. The XOS uses both Zero and Oh depending on version -->
-   <fingerprint pattern="Extreme(?:X[0O]S|Ware X[0O]S) version (\S+) \S+ by \S+ on .*$">
+   <fingerprint pattern="Extreme(?:X[0O]S|Ware X[0O]S) (?:\s?\([Xx]\d{0,3}-\d{0,2}[Xx]\)\s)?version (\S+) \S+ by \S+ on .*$">
       <description>Extreme Networks Switches</description>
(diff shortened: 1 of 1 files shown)


xml/snmp_sysdescr.xml
```diff
@@ -2314,7 +2314,7 @@ Copyright (c) 1995-2005 by Cisco Systems
    =======================================================================-->
 
    <!-- Sneaky. The XOS uses both Zero and Oh depending on version -->
-   <fingerprint pattern="Extreme(?:X[0O]S|Ware X[0O]S) version (\S+) \S+ by \S+ on .*$">
+   <fingerprint pattern="Extreme(?:X[0O]S|Ware X[0O]S) (?:\s?\([Xx]\d{0,3}-\d{0,2}[Xx]\)\s)?version (\S+) \S+ by \S+ on .*$">
```
gschneider-r7: This might be too narrow and may result in missed matches, but I don't have any other examples to go on right now.

jhart-r7: Why not just add `(?:\S+ )?`?  That way it'll allow any other non-whitespaced token before the version.

gschneider-r7: Done, thanks for the suggestion. KISS tends to work better anyway right?

PR #58, opened 2015-05-21
Adds/corrects more SNMP sysDescr fingerprints

jhart-r7:
/CC @alynn71 @gwiseman-r7

### DIFF
--- a/xml/snmp_sysdescr.xml
+++ b/xml/snmp_sysdescr.xml
@@ -3363,9 +3363,18 @@ Copyright (c) 1995-2005 by Cisco Systems
    </fingerprint>
 
    <fingerprint pattern="^Integrated Lights-Out (\d) \d+\.\d+ [A-Za-z]{3} \d{1,2} \d{4}$">
-      <description>HP Integrated Lights-Out (iLO)</description>
+      <description>HP Integrated Lights-Out (iLO) with firmware version</description>
       <example os.version="4">Integrated Lights-Out 4 2.02 Aug 18 2014</example>
       <!--2.02 is firmware version-->
+      <param pos="0" name="os.vendor" value="HP"/>
+      <param pos="0" name="os.family" value="iLO"/>
+      <param pos="0" name="os.device" value="iLO"/>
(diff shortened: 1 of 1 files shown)


xml/snmp_sysdescr.xml
```diff
@@ -2631,22 +2632,10 @@ Copyright (c) 1995-2005 by Cisco Systems
       <param pos="4" name="os.version.version"/>
    </fingerprint>
 
-   <fingerprint pattern="^(\S+) (.*?) Digital UNIX V(\S+) \(Rev\. ([^\)]+)\).*TCP/IP$">
+   <fingerprint pattern="^(\S+) (.*?) Digital UNIX V(\S+)\s+\(Rev\. ([^\)]+)\).*TCP/IP$">
       <description>Digital/Compaq/HP Tru64 Unix</description>
-      <example>clima.igpsdfpe COMPAQ AlphaServer DS20E 500 MHz Digital UNIX V4.0F (Rev. 1229); Mon Apr 12 13:30:41 EDT 1999 TCP/IP</example>
```
hmoore-r7: What is the driver for removing these examples?

jhart-r7: My preference is to have as many examples as possible that test what the regular expression is designed to handle but as few as necessary.  All of these are basically the same in that they don't each test something particular in the fingerprint that others don't.  For example, if all that differs between two examples it the version and the hostname, and the regular expression captures both fields the same way in both examples, there is little point in having the second example.   

There is also a performance hit here.  Part of the unit tests test that any given example only matches the `fingerprint` in which it was specified, which means trying to match every example against every `fingerprint` pattern specified before its enclosing `fingerprint`.

I don't believe we are losing any test coverage here but feel free to point out any instances that you see.

hmoore-r7: Got it thanks, will eyeball them once more for unique test cases that might have been lost

hmoore-r7: A few unique examples are lost, but they don't make a difference to the pattern.

PR #60, opened 2015-05-22
Improve some SMTP fingerprints

jhart-r7:
/CC @alynn71 @gwiseman-r7

### DIFF
--- a/xml/smtp_banners.xml
+++ b/xml/smtp_banners.xml
@@ -913,6 +913,19 @@ The system or service fingerprint with the highest certainty overwrites the othe
       <param pos="4" name="system.time"/>
    </fingerprint>
 
+   <fingerprint pattern="^(\S+) V\S+, OpenVMS V(\S+) (\S+) ready at .*$">
+      <description>Some unknown mail server on OpenVMS</description>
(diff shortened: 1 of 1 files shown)


xml/smtp_banners.xml
```diff
@@ -1713,13 +1820,21 @@ The system or service fingerprint with the highest certainty overwrites the othe
...
     </fingerprint>
 
-   <fingerprint pattern="^([^ ]+) E?SMTP(?: Ready\.?)?$">
-     <description>
-       catch all for daemons that have no distinguishing fingerprint whatsoever
-     </description>
-      <example>foo.example.com ESMTP</example>
-      <example>foo.example.com ESMTP Ready</example>
-      <example>foo.example.com SMTP</example>
+    <fingerprint pattern="^(\S+) E?SMTP Perl" flags="REG_ICASE">
```
hmoore-r7: Is this fingerprint useful to have?

jhart-r7: What I think you are elluding to is what I meant to do -- asserting `host.name` is useful but just Perl for the service fingerprint isn't useful (at least not today).  I could go either way.

hmoore-r7: Got it, thanks! Makes sense just to get host.name

PR #62, opened 2015-05-29
Add OS X 10.8 and 10.9 NTP fingerprints

jhart-r7:
These will typically only listen/respond locally, but just in case...

/CC @gschneider-r7, who originally reported the missing 10.9 support

### DIFF
--- a/xml/ntp_banners.xml
+++ b/xml/ntp_banners.xml
@@ -277,7 +277,7 @@ NTP "banners", taken from a readvar response
     <param pos="0" name="os.certainty" value="0.9"/>
   </fingerprint>
   <fingerprint pattern="^.*version=&quot;ntpd ([^ ]+)[^&quot;]+&quot;,.*processor=&quot;([^&quot;]+)&quot;,.*system=&quot;Darwin/?13\.([^&quot;]+)&quot;" flags="REG_DOT_NEWLINE,REG_ICASE">
-    <description>ntpd running on Mac OSX 10.9/Lion</description>
+    <description>ntpd running on Mac OSX 10.9/Mavericks</description>
     <example service.version="4.2.6@1.2089-o" os.arch="x86_64" os.version.version="4.0">
       version="ntpd 4.2.6@1.2089-o Fri May 28 01:20:53 UTC 2010 (1)",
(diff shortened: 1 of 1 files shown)


xml/ntp_banners.xml
```diff
@@ -258,6 +258,42 @@ NTP "banners", taken from a readvar response
...
+    <param pos="0" name="os.family" value="Mac OS X"/>
+    <param pos="0" name="os.product" value="Mac OS X"/>
+    <param pos="0" name="os.device" value="General"/>
+    <param pos="2" name="os.arch"/>
+    <param pos="0" name="os.version" value="10.8"/>
+    <param pos="3" name="os.version.version"/>
+    <param pos="0" name="os.certainty" value="0.9"/>
+  </fingerprint>
+  <fingerprint pattern="^.*version=&quot;ntpd ([^ ]+)[^&quot;]+&quot;,.*processor=&quot;([^&quot;]+)&quot;,.*system=&quot;Darwin/?13\.([^&quot;]+)&quot;" flags="REG_DOT_NEWLINE,REG_ICASE">
+    <description>ntpd running on Mac OSX 10.9/Lion</description>
```
gschneider-r7: Should be [10.9/Mavericks](http://en.wikipedia.org/wiki/OS_X#Versions)

xml/ntp_banners.xml
```diff
@@ -258,6 +258,42 @@ NTP "banners", taken from a readvar response
...
+      processor="x86_64", system="Darwin/13.4.0", leap=3, stratum=16,
+    </example>
+    <param pos="0" name="service.family" value="NTP"/>
+    <param pos="0" name="service.product" value="NTP"/>
+    <param pos="1" name="service.version"/>
+    <param pos="0" name="os.vendor" value="Apple"/>
+    <param pos="0" name="os.family" value="Mac OS X"/>
+    <param pos="0" name="os.product" value="Mac OS X"/>
+    <param pos="0" name="os.device" value="General"/>
+    <param pos="2" name="os.arch"/>
```
gschneider-r7: I think it's a given that anything since 10.7 is going to be Intel 64-bit arch

PR #63, opened 2015-05-29
Force match input to be binary

hmoore-r7:
This change resolves a warning where the regular expressions were specified as non-unicode (//n), but the actual call to ```match()``` would accept utf-8 input. This forces all input match strings to be 8-bit ascii and patches two examples to remove high-ascii text (actual responses should still match).


jhart-r7:
This doesn't seem correct to me; perhaps I'm missing something. Examples before and after a distinctly different, right?  You've replaced 4 2-byte code points (I may be getting the terminology wrong...) with 4 1-byte characters.

Since you are just changing the example, I guess I now question whether the regex will ever match the original example banner in the wild.  hôte is French for host and, at least in Ruby, does match a Regex of `/^h.te$/`.

So, looks good.

hmoore-r7:
The original examples were one-byte utf-8, so the "." after forcing to binary should be valid for both Ruby and Java use cases

jhart-r7:
The original example is `h�te`, which is actually 6 bytes (`0x68 0xef 0xbf 0xbd 0x74 0x65`).  

Host in French is hôte, which is 5 bytes (`0x68 0xc3 0xb4 0x74 0x65`).

The new regex won't match the original example, which is why you changed it:

```
irb(main):007:0> 'hôte'.force_encoding('UTF-8') =~ /^h.te$/
=> 0
irb(main):008:0> 'hôte'.force_encoding('BINARY') =~ /^h.te$/
=> nil
```

So, from what I gather, this means that any users of Recog must also always force the input to binary.  In general I'm fine with that, but the thing I don't like about it is that contributors/we then need to hand-mangle examples that might legitimately be in UTF-8 in the real world so that they'd properly match the  regex.  It also would hamper any sort of automation efforts to bulk import new fingerprints into recog from real world data (Internet scans, logs, etc).

Please correct me if any of this is wildly incorrect.  This seems like a simple change with a non-trivial impact.

hmoore-r7:
Ah, thanks! The challenge is that the current regexes (in ruby) append //n, which forces the input to be binary. We might have a more complicated problem to solve then.

hmoore-r7:
It seems like we either need to decide that all input is UTF8 (and scrub accordingly and fix the //n specifier) or go the other route and account for more than one byte for these.

hmoore-r7:
Scanning the regexes, it seems forcing to binary and changing those examples to be .{1,5} or similar may be better

jhart-r7:
I think something like `.{1,5}` will just cause other problems as it will then match any sequence of 1 to 5 characters, which isn't what we want.  We want it to match one character, specifically the second letter of host in French.

Going back even further on this, I'm now questioning the original examples that @TomSellers provided.  The original example is (again) h�te, (0x68 0xef 0xbf 0xbd 0x74 0x65), and � (0xef 0xbf 0xbd) is the unicode replacement character, used when the unicode character is unknown or can't be displayed.  It seems unlikely to me that the actual character sent by the French version of this message is unknown, so I took a look at the mysql source and my assumption seems correct.  Here is what the source says:

```
Le hôte '%-.64s' n'est pas authorisé à se connecter à ce serveur MySQL"
```

I'm not sure the provided French example is correct.  Same goes for the Spanish version.  

@TomSellers, can you shed some more light on where you got these examples?

jhart-r7:
Before we do anything with this PR, we need to handle #65, as it is the only thing that would catch any critical failures.

jhart-r7:
@hmoore-r7 I pushed some changes to this PR that I think should sort everything out.

PR #64, opened 2015-06-01
Miscellaneous SNMP, FTP, SSH and HTTP fingerprints

jhart-r7:
Found after using Recog data to fingerprint a private lab.

### DIFF
--- a/xml/ftp_banners.xml
+++ b/xml/ftp_banners.xml
@@ -842,6 +842,15 @@ against these patterns to fingerprint FTP servers.
     <param pos="2" name="os.version"/>
   </fingerprint>
 
+  <fingerprint pattern="^(\S+) FTP server \(MikroTik ([\d\.]+)\) ready\.?$">
+    <description>MikroTik</description>
+    <example host.name="example.com" os.version="6.18">example.com FTP server (MikroTik 6.18) ready</example>
+    <param pos="0" name="os.vendor" value="MikroTik"/>
+    <param pos="0" name="os.product" value="RouterOS"/>
+    <param pos="1" name="host.name"/>
+    <param pos="2" name="os.version"/>
+  </fingerprint>
+
   <fingerprint pattern="^(\S+) FTP server ready.?$" flags="REG_ICASE">
     <description>Generic FTP fingerprint with a hostname</description>
     <example host.name="example.com">example.com FTP server ready.</example>
--- a/xml/ftp_banners.xml
+++ b/xml/ftp_banners.xml
@@ -184,12 +184,21 @@ against these patterns to fingerprint FTP servers.
     <param pos="0" name="service.family" value="ProFTPD"/>
     <param pos="0" name="service.product" value="ProFTPD"/>
   </fingerprint>
+  <fingerprint pattern="^=\(&lt;\*&gt;\)=\-\.:\. \(\( Welcome to Pure-FTPd ([\d\.]+) \)\) \.:\.\-=\(&lt;\*&gt;\)=\-$">
+    <!-- yes, the leading and trailing text is not balanced.
+         the leading text is missing the - at the beginning -->
+    <example service.version="1.0.11">=(&lt;*&gt;)=-.:. (( Welcome to Pure-FTPd 1.0.11 )) .:.-=(&lt;*&gt;)=-</example>
+    <description>Pure-FTPd versions &lt;= 1.0.13 (at least as far back as 1.0.11)</description>
+    <param pos="0" name="service.family" value="Pure-FTPd"/>
+    <param pos="0" name="service.product" value="Pure-FTPd"/>
+    <param pos="1" name="service.version"/>
+  </fingerprint>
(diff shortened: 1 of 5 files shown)


xml/ftp_banners.xml
```diff
@@ -184,8 +184,9 @@ against these patterns to fingerprint FTP servers.
     <param pos="0" name="service.family" value="ProFTPD"/>
     <param pos="0" name="service.product" value="ProFTPD"/>
   </fingerprint>
-  <fingerprint pattern="^-{10} Welcome to Pure-FTPd (.*)-{10}$">
+  <fingerprint pattern="^-{9,10} Welcome to Pure-FTPd (.*)-{9,10}$">
```
dpicotte-r7: is looking for the ----'s even needed?

jhart-r7: I've never seen a default Pure-FTPDd banner that didn't have them, but that doesn't mean there aren't any.  In general, in situations like this, we use our best judgement.  

I took a look, and Pure-FTPd versions at least from 1.0.11 (from ~2002) use the following for a default banner:

```
    addreply_noformat(0, "=(<*>)=-.:. (( " MSG_WELCOME_TO 
                      " PureFTPd " VERSION " )) .:.-=(<*>)=-");
```

Versions newer than that, however, use 9 (not 10) `-`s.  My guess is that the original regex here was wrong but I don't know where that example came from.  For now I'd say that the 9 is more correct based on source code but 9-10 may be more accurate for all versions of Pure-FTPd.  

For now, I'm going to expand this to account for both variants.

xml/ftp_banners.xml
```diff
@@ -184,11 +184,21 @@ against these patterns to fingerprint FTP servers.
     <param pos="0" name="service.family" value="ProFTPD"/>
     <param pos="0" name="service.product" value="ProFTPD"/>
   </fingerprint>
-  <fingerprint pattern="^-{10} Welcome to Pure-FTPd (.*)-{10}$">
+  <fingerprint pattern="^=\(&lt;\*&gt;\)=\-\.:\. \(\( Welcome to Pure-FTPd ([\d\.]+) \)\) \.:\.\-=\(&lt;\*&gt;\)=\-$">
```
pmiseiko-r7: Don't need to escape -'s because they're not in blocks.
Don't need to escape the . in the block for the version because it's not considered a wild card inside a block.

xml/ftp_banners.xml
```diff
@@ -841,6 +851,15 @@ against these patterns to fingerprint FTP servers.
...
+  <fingerprint pattern="^(\S+) FTP server \(MikroTik ([\d\.]+)\) ready\.?$">
+    <description>MikroTik</description>
+    <example host.name="example.com" os.version="6.18">example.com FTP server (MikroTik 6.18) ready</example>
+    <param pos="0" name="os.vendor" value="MikroTik"/>
+    <param pos="0" name="os.product" value="RouterOS"/>
+    <param pos="1" name="host.name"/>
+    <param pos="2" name="os.version"/>
+  </fingerprint>
+
   <fingerprint pattern="^(\S+) FTP server ready.?$" flags="REG_ICASE">
```
pmiseiko-r7: Didn't escape the . here, but did above.

xml/http_servers.xml
```diff
@@ -3376,12 +3376,13 @@
       <param pos="1" name="service.version"/>
    </fingerprint>
 
-   <fingerprint pattern="^Boa/(\d+\.[a-zA-Z0-9\.\-]+)">
+   <fingerprint pattern="^Boa/([\d\.]+(?:[a-zA-Z0-9\.\-]*))">
```
pmiseiko-r7: Don't need to escape the - because it's at the end (or if you put it at the start).
Don't need to escape the . because it's not a wild card inside a block.

hmoore-r7: @pmiseiko-r7 The dot needs to be escaped for this to parse right? (7.something25.taco-party). It would prevent matching a version that didn't start with a number and a dot.

xml/ftp_banners.xml
hmoore-r7: The dot seems necessary here as well to match the literal ```.:.```

jhart-r7: Paul's comment about `.` was for `([\d\.]+)`, which is correct.  Fixed on both accounts (nice catch).

xml/ftp_banners.xml
jhart-r7: Outside the scope of this PR, but fixed anyway.

xml/ftp_banners.xml
jhart-r7: Fixed.

xml/http_servers.xml
jhart-r7: The reason is the same as it was above -- the `.` in square brackets does not need to be escaped.  I'll fix that because I fixed it in response @pmiseiko-r7's other comment too.  

To the point about not needing to escape the `-` because it is at the end, true, but I'm going to avoid that problem and just change this regex  to simply look for something that begins with a digit and a period and then contains any non-whitespace.

PR #67, opened 2015-06-12
Correct Ubuntu OpenSSH fingerprints

jhart-r7:
I noticed that newer Ubuntu 12.04 systems were not getting as-useful OS fingerprints from SSH.  The reason was that these newer 12.04 systems have a slightly different format which did not match the existing 12.04 fingerprint, but the overly greedy fingerprint later on that does match but fails to assert OS fingerprint.  So, I removed the greedy fingerprint and updated the 12.04 one.


jhart-r7:
Actually, going to close this for now.  there are more problems.

PR #68, opened 2015-06-12
Improve Debian and Ubuntu OpenSSH fingerprints

jhart-r7:
There were several catch-all OpenSSH fingerprints that were resulting in us only getting very generic versions from some OpenSSH installations.  I've removed those and updated all of the Ubuntu and Debian fingerprints I could find.

I also ran this PR through a set of Sonar data for 22/TCP from earlier this year.  As I expected, we actually lose quite a bit of coverage.  Without this PR, we miss 76021 out of 78603 unique banners.  With this PR, we miss 77494 out of 78603.  So, neither are especially good, but now we are worse in terms of overall coverage but it will now give us a chance to track down all of the banners that had previously been matching these catch-alls and see if we can do better.

/CC @gwiseman-r7

### DIFF
--- a/xml/ssh_banners.xml
+++ b/xml/ssh_banners.xml
@@ -477,7 +477,7 @@ fingerprint SSH servers.
 
    <!-- SSH-1.99-OpenSSH_4.3p2-4.cern-hpn-CERN-4.3p2-4.cern -->
 
-   <fingerprint pattern="^OpenSSH_?([^\s]*)\s*(.*)$">
+   <!--<fingerprint pattern="^OpenSSH_?([^\s]*)\s*(.*)$">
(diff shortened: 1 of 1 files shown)


gwiseman-r7:
I'd be inclined to leave the catch-alls in for now, to minimize the chance of introducing false negatives due to missed fingerprints. Reduced certainty seems like a good idea.

Perhaps they could be temporarily dropped at some point for a focused data collection exercise to see what we're missing.

:anchor: ship it

cmccrisken-r7:
Agree on reducing the certainty to promote quality over quantity, whilst casting the net a bit wider. :shipit: ship it.

PR #71, opened 2015-06-23
Respect .ruby-version and .ruby-gemset

todb-r7:
Since I work in a few different repos, keeping my gemsets straight is kind of a thing.

If there's a reason to not support .ruby-version and .ruby-gemset, then by all means, close this out, but it seems weird to ignore it specifically, and expect people to manually manage their own gemsets.


hmoore-r7:
This could have unwanted downstream affects due to what includes recog.

shuckins-r7:
@todb-r7 ruby-version isn't usually tracked in VCS for gems unless the gem actually requires that version. And ruby-gemset should not be checked in for gems at all from what I've seen: http://stackoverflow.com/a/16738020

You would be trading some personal convenience for making it appear that this gem is stuck at a particular version of Ruby. I think we should definitely expect people to manage their own gemsets when dealing with a standalone gem.

todb-r7:
Ah gotcha. Will close then, thanks for the education.

PR #72, opened 2015-06-23
Create a CHANGELOG, other thoughts on release

todb-r7:
It's pretty simple, just git shortlog and trim the malformed commit messages. After incrementing the version, just run: `git shortlog v2.0.5...HEAD`, add a header and some enclosing ticks to avoid markdown formatting, and be done. Next time, append to the top, I guess?

Open for discussion and automation. If this is useless, ignore.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,52 @@
+Changes from 2.0.5 to 2.0.6
+===========================
+
+```
+HD Moore (2):
+      Force input to be binary, resolves warnings about //n
+      Convert unicode to plain text in examples
+
+Joakim Kennedy (1):
+      DCA-1216: Windows 10 Fingerprint
+
+Jon Hart (33):
+      Revert "Convert unicode to plain text in examples"
+      Revert "Force input to be binary, resolves warnings about //n"
+      Use no Regexp flags by default
+      Rename regexp_factory rspec so that it actually runs
+      Correct regexen and examples for French and Portuguese (not Spanish)
+      Correct spec to test for proper flags
+      Add test to ensure that multiple flags are combined
+      Remove now unnecessary (right?) jruby rspec pending
+      Correct required ruby version
+      Minor code cleanup
+      Update Ubuntu 12.04 OpenSSH fingerprint
+      Remove over-gready fingerprint for OpenSSH on Ubuntu
+      Add Debian Squeeze OpenSSH banner; remove greedy
+      Add a fingerprint for OpenSSH with just a version, no comment
+      Temporarily disable OpenSSH catch-all
+      Relax the OpenSSH with just a version regex
+      Add older Ubuntu 7.10 OpenSSH fingerprint
+      Update Ubuntu 8.04 OpenSSH regex
+      Consolidate and improve Ubuntu 10.04 SSH fingerprints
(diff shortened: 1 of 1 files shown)


jhart-r7:
I have no real issues with the Changelog; it is simple enough for now and we can change in the future if we want.  It would be cool if this could be hooked into `rake release`.

todb-r7:
Well, you can always just drop a compare in the release notes, too, which is what I did here:

https://github.com/rapid7/recog/releases/tag/v2.0.6

That way you're not cluttering up with generated text. I'm sure approximately zero people read changelogs when they install from rubygems.

todb-r7:
BTW it'd be nice to have the release version tag signed.

jhart-r7:
If we are going to have changelog information in multiple places (in `CHANGELOG.md` and in the Github release), IMO they should be the same.  In fact, because we also want to automate the Github release process (IOW, have it happen automagically when `rake release` is run`), it makes even more sense IMO to have one.  

What about this -- `CHANGELOG.md` will be a living document that includes all changelog information for all releases.  We can either start it now with 2.0.6, but we could also generate it historically too.  Over time, we'll add future releases.  The per-release changelog information will be 100% identical to what is ultimately put in the corresponding Github release.

jhart-r7:
@todb-r7, I agree re: release tag signing.  I sign all of my commits by default but only because of my `.gitconfig` aliases.  I don't understand the `rake release` magic enough to see if we can enforce it there.

todb-r7:
I'm just of the opinion that nobody ever reads CHANGELOG, except developers, who are going to look at commit history anyway. I don't see a lot of upside in generating one.

I suppose in the distant future, when we've moved off of git and github and onto whatever the next thing is, it'd be nice to have a flat text tile that has all the changes?

todb-r7:
poking around, I'm with you @jhart-r7 , can't figure out if pushing signed tags is at all an option with Bundler. Weird.

jhart-r7:
@todb-r7, I have a PR against yours in the works that should help automate this.

jhart-r7:
Maybe a .gitconfig alias for `tag` would be sufficient here -- simply have it pass in `-S` all the time.  

To the point about whether or not we need a changelog, as of #73 we don't need it for github releases because those are now gone (as in we'll stop producing them).  I actually like the idea of having a changelog over time, but having it also be part of the commit history will be a little weird so perhaps using https://github.com/rapid7/recog/compare/<v1>...<v2> is sufficient for any use case that would require changelog-like data.  That is about all I can say as a con with regards to keeping a changelog.

If someone has a particularly compelling reason why generating a changelog would be useful/helpful, then we can use my suggested script in https://github.com/todb-r7/recog/issues/1

todb-r7:
## On Changelog:

I, for one, love the script at todb-r7#1. Super fun.

But... I question the usefulness.

## On signing tags

Well, according to the discussion over at https://github.com/bundler/bundler-features/issues/90 , Bundler has no built in mechanism to get people to do The Right Thing and sign tags, and they're not very keen to start. So yeah, a `t` alias would be the best bet (I have `c` for commit -S and `m` for merge -S already).

## On this PR

Looks dead to me. Reopen if there's more to this, but I think we've talked it out.

PR #73, opened 2015-06-23
Remove documentation on github releases

jhart-r7:
To the best of my knowledge, nobody and nothing is actually using the Github releases.  Nexpose stuff was, at one point, but not any more.  Lets just remove it since it is currently a manual and often forgotten step.

/CC @todb-r7 @hmoore-r7 for any possible metasploit-y questions

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,9 +3,8 @@
 The users and maintainers of Recog would greatly appreciate any contributions
 you can make to the project.  These contributions typically come in the form of
 filed bugs/issues or pull requests (PRs).  These contributions routinely result
-in new versions of the [recog gem](https://rubygems.org/gems/recog) and the
-[recog release](https://github.com/rapid7/recog/releases) to be released.  The
-process for everything is described below.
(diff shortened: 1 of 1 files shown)


hmoore-r7:
Totally works for me, thanks!

mrongali-r7:
I the

Sent from my iPhone

NOn Jun 23, 2015, at 12:59 PM, HD Moore <notifications@github.com<mailto:notifications@github.com>> wrote:


Totally works for me, thanks!

-

todb-r7:
@mrongali-r7 , did not copy.

gschneider-r7:
Is this mainly driven by it being a manual process and frequent releases? We've been using the github releases for nexpose-client as both a release notes vehicle and alternative binary distribution source, which seems to work well for us.

jhart-r7:
Not necessarily a problem because of frequency, but it is enough of a manual step that produces something that nobody uses that I just wanted to get rid of it.  It simplifies the process.  About all it really removes for us functionality was the changelog stuff that was happening before.  The binary distributions should still be automatically available based on the tags.

PR #76, opened 2015-07-07
Use Debian-format machine readable license

todb-r7:
I like this format a lot because it means that I don't have to read the
entire LICENSE every time just to remember if it's BSD 3-clause,
2-clause, MIT, or whatever (they're all really similar). Also, someday,
we might want to package it in a machine readable format, so might as
well get on the stick with that now.

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -1,0 +1,23 @@
+Copyright (c) 2014-2015, Rapid7
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without modification,
+are permitted provided that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright notice, this
+  list of conditions and the following disclaimer.
+
+* Redistributions in binary form must reproduce the above copyright notice, this
+  list of conditions and the following disclaimer in the documentation and/or
+  other materials provided with the distribution.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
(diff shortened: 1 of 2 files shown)


jhart-r7:
FWIW, I took a quick look at this.  My only concerns, which are probably minor, are:

* Some documentation I read seemed to imply that the full text of the license must still be included in the same file that references the license short-name.  IOW, moving the license text to `COPYING` may be incorrect.  I couldn't find a definitive answer.
* At least when used in Debian (and presumably variants), it looks like they expect the file to literally be named `copyright`.  Does this matter?  Put another way, if we were to ship this, should we expect tools that know how to parse these formats to be smart enough to able to know that the license text is in `COPYING` and not `LICENSE`?

After reading https://help.github.com/articles/open-source-licensing/ and re-thinking my questions, how about you simply convert the existing `LICENSE` to use the new format but also include the full-license text at the bottom (which I believe is valid)?

todb-r7:
I'm happy to do it any way that makes sense, but note that the format on this PR matches 

https://github.com/rapid7/metasploit-framework/blob/master/COPYING

and

https://github.com/rapid7/metasploit-framework/blob/master/LICENSE

These are used as part of the normal Debian packaging for the Kali distro, and I haven't heard any complaints there.

(cc @bturner-r7 @cdoughty-r7 )

bturner-r7:
All fair points.  Yes, debian expects the file to be `debian/copyright`.  We do this when we package Framework for Debian/Kali.  The reason I ended up putting it all in LICENSE was so that when folks edited the license file, I didn't have to go back and edit the debian version.

I've never seen the debian machine-readable format used anywhere else, so I'm not sure its worth it when there isn't an associated debian package for recog.

If we do use the format, I agree with @jhart-r7 that we should include the full text in LICENSE like we do in Framework's: https://github.com/rapid7/metasploit-framework/blob/master/LICENSE#L280-L300

PR #77, opened 2015-07-07
Add missing OS for iLO using SNMP and OpenVMS using SSH

jhart-r7:
It was reported that some fingerprints were missing the appropriate `os.product` and I corrected the two instances.  While here I also made the `os.device` for HP iLO more consistent with what is done elsewhere and similarly with Comware.

/CC @alynn71


gwiseman-r7:
:anchor: ship it

dpicotte-r7:
Ship it

PR #79, opened 2015-07-23
Improve Aspen and Oracle Containers HTTP fingerprints

jhart-r7:
Newer versions of Oracle containers use a slightly different format that at least in the one example I found does not include a version.  Aspen is a Python based HTTP server.


dpicotte-r7:
ship it

PR #80, opened 2015-08-05
Fix several Sendmail SMTP regular expressions

jhart-r7:
Initially the problem was that a SMTP banner like this wouldn't result in an AIX fingerprint:

```
example.com ESMTP Sendmail AIX5.1/8.11.6p2/8.11.0; Fri, 28 Aug 1970 19:42:05 -0800
```

It was still matching and asserting a Sendmail fingerprint but that was wrong.  The root cause was several defects which you can see in my individual commits.


dpicotte-r7:
Ah vendors with their non-ISO date formats... ship it!

PR #82, opened 2015-08-06
Add remaining MySQL error messages

### LINKED ISSUE #66: Add all remaining, missing MySQL error messages
We currently have some of them but are missing many of the language variants from http://osxr.org/mysql/source/sql/share/errmsg-utf8.txt.  For the error messages we currently check, make sure all of the supported languages are in Recog.

Specifically, we need to add the missing languages from the following sections:

* ER_HOST_IS_BLOCKED
* ER_HOST_NOT_PRIVILEGED
* ER_CON_COUNT_ERROR 
* ER_CANT_CREATE_THREAD

/cc @TomSellers, FYI.

jhart-r7:
AFAIK this data is currently unused, but it fixes #66 regardless.

### DIFF
--- a/misc/convert_mysql_err
+++ b/misc/convert_mysql_err
@@ -2,24 +2,19 @@
 
 # Takes the MySQL error messages from sql/share/errmsg-utf8.txt, locates the
 # provided error message type (for example, ER_HOST_NOT_PRIVILEGED), then
-# creates XML snippets for each language to be used in Recog
+# creates XML snippets for each locale to be used in Recog.  Note that this
+# cannot be used as-is to generate mysql_errors.xml, or oftentimes even parts
+# -- it merely spits out XML snippets that you can start with; many will still
+# need to be modified by hand.
 
 require 'builder'
 require 'open-uri'
 require 'securerandom'
 
-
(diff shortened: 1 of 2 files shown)


misc/convert_mysql_err
```diff
@@ -0,0 +1,53 @@
...
+
+
+def generate_recog(error_name, lang, error_message)
+  xml = Builder::XmlMarkup.new(target: STDOUT, indent: 2)
+  # if this error message has a format string in it, warn because
+  # this script won't attempt to handle the format string in the regex
+  # or the example
+  format_count = error_message.scan(/%/).size
+  unless format_count == 0
+    warn("#{format_count} possible format strings found -- you'll need to deal with this")
```
gwiseman-r7: a bit of context (the error_name/lang) might be helpful to add to this warning

gwiseman-r7:
:anchor: ship it

misc/convert_mysql_err
jhart-r7: Good point.  In general, every error message of a particular type will have the same format string part as across all languages, so what I'll do here is just warn once.

Issue #75, opened 2015-06-30
Document and improve experience related to fingerprint regular expression flags


jhart-r7:
As experienced in #74, there is room for improvement as it relates to how fingerprint regular expression flags are handled.  

Today, the `flags` attribute is parsed in such a way that `fingerprint` authors can use multiple different formats for regular expression flags, and that during *recog testing time*, the variety of ways in which the flags are specified are normalized to something that Recog, which is written in Ruby, can understand.  The issue with this, as we saw in #74, is that if we support multiple methods of specifying the same flag (in this case, `REG_MULTILINE` is the Java/GNU/perl friendly way and `MULTILINE` was the Ruby friendly way (per http://ruby-doc.org/core-2.1.1/Regexp.html#method-i-options)), any product that consumes Recog content must also have the same support for multiple methods of specifying options, or Recog itself needs to provide that mechanism.

Further proof of this is still lurking in `IGNORECASE`, which would still be allowed in current Recog but would break any Java/GNU/Perl implementation because `REG_ICASE` is the preferred method.

PR #83, opened 2015-08-10
POC for #75

jhart-r7:
Right now this is just a POC for the second proposed solution in #75.  Basically I am proposing that we do away with the `flags` attribute entirely and utilize special grouping or non-grouping constructs as necessary to provide equivalent functionality.  So, for example instead of `pattern="foo" flags="REG_ICASE"`, we will use `pattern="(?i:foo)"`.

There are basically 3 parts to this PR:

  1. Write a utility to convert only affected `fingerprints` to the new format
  2. Convert the existing XML fingerprint files
  3. Remove all code related to parsing of `flags`

To merge this, at a minimum we'd need to bump to 3.x and any projects that utilize recog will need to ensure that their usage of the Recog content properly handles these new regular expressions.


jhart-r7:
Closing for now.

PR #84, opened 2015-08-13
Add schema for and validate recog content; fix content

### LINKED ISSUE #81: Add schema for recog content
Today we do some validation like things all over the place, recog_verify, recog_match, some unit tests, some cucumber tests.  We should remove anything that is better served in a schema and add it that way instead.

jhart-r7:
Fixes #81.  There are three main things happening here:

* A new schema was added that I believe enforces correctly formatted recog content
* This schema is used to test all content shipped with recog or tested with `recog_verify` and `recog_match`
* All existing content was updated to conform to the schema.  Most changes were just ordering, but there were some legitimate problems lurking too.  See https://github.com/rapid7/recog/pull/84/files?w=1 for an easier view, since there was a ton of whitespace changes.

There is still a bit of a disconnect between what is tested in the unit tests and what is tested by `recog_verify`, but I'm not going to solve that now.  This just ensures that the content adheres to our standard.

## Validation
 - [x] Unit tests pass
 - [x] Recog content and supporting ruby code tested out in some dependent projects (metasploit-framework and nexpose)


pmiseiko-r7:
This looks good.

PR #87, opened 2015-08-19
Make Cyrus MTA IMAP and POP consistent

jhart-r7:


gwiseman-r7:
Including the Invoca RPM stuff in the version seems legit. :anchor: ship it.

PR #88, opened 2015-08-26
Update dependent gems; fix/ignore warnings

jhart-r7:
My initial goal was to just update the versions of the gems we use, which personally makes my life easier  since I switch between numerous projects with similar dependencies.  Of course it wasn't so simple after all.  In a nutshell:

* All gems which we had pinned at specific versions/ranges have been updated and retested
* All gem dependency management has moved to the gemspec, rather than being in both
* The cucumber-based testing of the two executables we ship has been replaced with rspec-based testing instead.  Having multiple test frameworks makes contributions harder.

## Validation
 - [x] All tests pass without errors or warnings
 - [x] Gem builds and installs without errors or warnings
 - [ ] Test in metasploit-framework, perhaps pro


jhart-r7:
Going to dump this.  To make this PR useful we need to bump some gems, but that will break msf and/or pro or at least greatly slow this PR down since the gems would need to bump there too and that would take time.

PR #89, opened 2015-08-31
SMTP: sendmail.config => .config

zt2:
Hi all, I found there has some unnecessary field in `smtp_banners.xml`.

For example, each `sendmail.config.xxx` in `smtp_banners.xml` has same `service.product`: `Sendmail`.

Can we omit the prefix `sendmail.` ?

### DIFF
--- a/xml/smtp_banners.xml
+++ b/xml/smtp_banners.xml
@@ -1256,8 +1256,8 @@ The system or service fingerprint with the highest certainty overwrites the othe
       <param pos="0" name="system.time.format" value="EEE, dd MMM yyyy HH:mm:ss zzz"/>
       <param pos="1" name="host.name"/>
       <param pos="2" name="service.version"/>
-      <param pos="3" name="sendmail.config.version"/>
-      <param pos="4" name="sendmail.vendor.version"/>
+      <param pos="3" name="config.version"/>
+      <param pos="4" name="vendor.version"/>
       <param pos="5" name="system.time"/>
    </fingerprint>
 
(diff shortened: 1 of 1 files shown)


jhart-r7:
I don't see a problem with this and agree with the removal.  The only other change I'd suggest while here is making this consistent.  There is at least one other fingerprint that uses a `<foo>.config`.

xml/smtp_banners.xml
```diff
@@ -1256,8 +1256,8 @@ The system or service fingerprint with the highest certainty overwrites the othe
       <param pos="0" name="system.time.format" value="EEE, dd MMM yyyy HH:mm:ss zzz"/>
       <param pos="1" name="host.name"/>
       <param pos="2" name="service.version"/>
-      <param pos="3" name="sendmail.config.version"/>
-      <param pos="4" name="sendmail.vendor.version"/>
+      <param pos="3" name="config.version"/>
```
jhart-r7: We should probably move all of these to `service`, so for this it would be `service.config.version`.  Leaving them without the `service` prefix makes it a bit ambiguous as to what this particular extraction is about.

pmiseiko-r7:
@zt2, thank you for your contribution.  Could you clarify the reason you changed sendmail.config.version to config.version?  Per @jhart-r7's comment, service.config.version seems like a better option.  If there's no objection, we'll rename them service.* and accept the additional examples you provided.

zt2:
Hi @pmiseiko-r7 , This because I consider the prefix for some fields like `sendmail.config.version` is a little unnecessary. For example, if we fingerprint a smtp product as a `sendmail`, some fields in this fingerprint such as 'sendmail.config.version' is also marked as `sendmail` analysis results, omit the prefix `sendmail` is okey because we know this field is `sendmail` analysis result yet.

pmiseiko-r7:
@zt2, agree with you on removing sendmail since that's too specific, and it might not be sendmail.  However, what do you think about @jhart-r7's suggestion to prefix "config.version" with service so instead of config.version it is "service.config.version"?  That solves your problem of "sendmail.config.version" being too specific, and maintains a relationship that the "config.version" is part of the "service".

zt2:
@pmiseiko-r7, yes I agree with that, a `service` prefix might be more clearly

jhart-r7:
Closing for now.  This PR has gotten stale and can no longer be merged.

PR #91, opened 2015-09-01
SNMP fingerprints for Mac OS X

jkennedy-r7:
DCA-10
* Added SNMP fingerprints for 10.3 -> 10.10

### DIFF
--- a/xml/snmp_sysdescr.xml
+++ b/xml/snmp_sysdescr.xml
@@ -5158,6 +5158,121 @@ Copyright (c) 1995-2005 by Cisco Systems
    </fingerprint>
 
    <!--======================================================================
+                              Mac OS X
+   =======================================================================-->
+   <fingerprint pattern="^Darwin (\S*) \S* Darwin Kernel Version 7.(\d+).\d+: .*RELEASE_\S* ?\s* (\S* ?\S*)$">
+      <description>Mac OS X 10.3 (Panther)</description>
+      <example host.name="mac-10-3-s-u" os.arch="Power Macintosh" os.version.version="0">Darwin mac-10-3-s-u 7.0.0 Darwin Kernel Version 7.0.0: Wed Sep 24 15:48:39 PDT 2003; root:xnu/xnu-517.obj~1/RELEASE_PPC Power Macintosh</example>
+      <param pos="1" name="host.name"/>
+      <param pos="0" name="os.vendor" value="Apple"/>
+      <param pos="0" name="os.family" value="Mac OS X"/>
+      <param pos="0" name="os.product" value="Mac OS X"/>
+      <param pos="0" name="os.device" value="General"/>
+      <param pos="3" name="os.arch"/>
+      <param pos="0" name="os.version" value="10.3"/>
+      <param pos="2" name="os.version.version"/>
+      <param pos="0" name="os.certainty" value="0.9"/>
+   </fingerprint>
+
+   <fingerprint pattern="^Darwin (\S*) \S* Darwin Kernel Version 8.(\d+).\d+: .*RELEASE_\w* (\S* ?\S*)$">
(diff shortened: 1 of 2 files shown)


jhart-r7:
Odd... at least for NTP, some of this was added in https://github.com/rapid7/recog/pull/62.

jkennedy-r7:
Missed the newer fingerprints, the "BSD/OS with a version and arch" fingerprint between 10.6 and 10.7 tricked me.

xml/snmp_sysdescr.xml
```diff
@@ -5158,6 +5158,121 @@ Copyright (c) 1995-2005 by Cisco Systems
    </fingerprint>
 
    <!--======================================================================
+                              Mac OS X
+   =======================================================================-->
+   <fingerprint pattern="^Darwin (\S*) \S* Darwin Kernel Version 7.(\d+).\d+: .*RELEASE_\S* ?\s* (\S* ?\S*)$">
+      <description>Mac OS X 10.3 (Panther)</description>
+      <example host.name="mac-10-3-s-u" os.arch="Power Macintosh" os.version.version="0">Darwin mac-10-3-s-u 7.0.0 Darwin Kernel Version 7.0.0: Wed Sep 24 15:48:39 PDT 2003; root:xnu/xnu-517.obj~1/RELEASE_PPC Power Macintosh</example>
```
jhart-r7: really minor, but I try to sanitize hostnames when possible, typically using `example.com`

xml/snmp_sysdescr.xml
```diff
@@ -5158,6 +5158,121 @@ Copyright (c) 1995-2005 by Cisco Systems
...
+      <param pos="0" name="os.family" value="Mac OS X"/>
+      <param pos="0" name="os.product" value="Mac OS X"/>
+      <param pos="0" name="os.device" value="General"/>
+      <param pos="3" name="os.arch"/>
+      <param pos="0" name="os.version" value="10.3"/>
+      <param pos="2" name="os.version.version"/>
+      <param pos="0" name="os.certainty" value="0.9"/>
+   </fingerprint>
+
+   <fingerprint pattern="^Darwin (\S*) \S* Darwin Kernel Version 8.(\d+).\d+: .*RELEASE_\w* (\S* ?\S*)$">
```
jhart-r7: All the `pattern`s from here down are the same (minus the major version), and the one above differs slightly.  I don't see why.  Can you elaborate?  Also, is `\w` the right choice here (that'll only match a-z, A-Z, 0-9 and _), especially given that you use `\S` above as part of the matching of `RELEASE_`?

jkennedy-r7:
"All the patterns from here down are the same (minus the major version), and the one above differs slightly. I don't see why. Can you elaborate? Also, is \w the right choice here (that'll only match a-z, A-Z, 0-9 and ), especially given that you use \S above as part of the matching of `RELEASE`?"

@jhart-r7, when I scanned one of our 10.3 targets, no banner match was found. The banner had one extra blank space between the release and the arch, so I amended the regex with an optional space between the release and the arch.

 I changed the other patterns to also have this optional space. \w should be enough for the release so I changed it.

cmccrisken-r7:
Assuming the recog_verify tests passed and it was tested against the lab targets (which I know you did), then I'm all good for it to :ship:

mdobrska-r7:
I still don't think it's a purely unauthenticated way of fingerprinting (I think you must be using default community strings), but I guess no harm having these.

cmccrisken-r7:
@mdobrska-r7 That's a good point, actually. SNMP is a weird halfway house between authenticated and unauthenticated. But, then again, if we get data using root/default via SSH, then this is considered authenticated

pmiseiko-r7:
Valid point.  Unauthenticated is not the correct terminology but otherwise a good improvement.

:sheep: it

PR #92, opened 2015-09-08
New fingerprints for OS X

jkennedy-r7:
* Added banner fingerprint for 10.3 using QuickTime Stream Server.
* Added banner fingerprint for 10.5 and 10.6 using Samba

### DIFF
--- a/xml/http_servers.xml
+++ b/xml/http_servers.xml
@@ -3512,7 +3512,7 @@
     </fingerprint>
 
            <!-- Apple QuickTime streaming server -->
-    <fingerprint pattern="^QTSS\/([\d*.]*) \(Build\/[\d*.]*; Platform\/MacOSX; Release\/Panther">
+    <fingerprint pattern="^QTSS\/([\d\.]+) \(Build\/[\d\.]+; Platform\/MacOSX; Release\/Panther">
       <description>QTSS on OS X 10.3</description>
       <example service.version="5.0">QTSS/5.0 (Build/452; Platform/MacOSX; Release/Panther; )</example>
       <example service.version="5.0.3.3">QTSS/5.0.3.3 (Build/452.22.4; Platform/MacOSX; Release/Panther; Update/Security; )</example>
--- a/xml/http_servers.xml
+++ b/xml/http_servers.xml
@@ -3512,14 +3512,27 @@
     </fingerprint>
 
            <!-- Apple QuickTime streaming server -->
-    <fingerprint pattern="Platform\/MacOSX; Release\/Panther;">
-      <description>OS X 10.3</description>
-      <example>QTSS/5.0 (Build/452; Platform/MacOSX; Release/Panther; )</example>
-      <example>QTSS/5.0.3.3 (Build/452.22.4; Platform/MacOSX; Release/Panther; Update/Security; )</example>
+    <fingerprint pattern="^QTSS\/([\d*.]*) \(Build\/[\d*.]*; Platform\/MacOSX; Release\/Panther">
+      <description>QTSS on OS X 10.3</description>
+      <example service.version="5.0">QTSS/5.0 (Build/452; Platform/MacOSX; Release/Panther; )</example>
+      <example service.version="5.0.3.3">QTSS/5.0.3.3 (Build/452.22.4; Platform/MacOSX; Release/Panther; Update/Security; )</example>
       <param pos="0" name="os.vendor" value="Apple"/>
       <param pos="0" name="os.family" value="Mac OS X"/>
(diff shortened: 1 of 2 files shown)


xml/http_servers.xml
```diff
@@ -3510,4 +3510,16 @@
       <param pos="0" name="service.product" value="gSOAP"/>
       <param pos="1" name="service.version"/>
     </fingerprint>
+
+           <!-- Apple QuickTime streaming server -->
+    <fingerprint pattern="Platform\/MacOSX; Release\/Panther;">
```
jhart-r7: Why not anchor this and also fingerprint the service.

gwiseman-r7:
:anchor: ship it

xml/http_servers.xml
```diff
@@ -3510,4 +3510,29 @@
...
+      <param pos="0" name="os.vendor" value="Apple"/>
+      <param pos="0" name="os.family" value="Mac OS X"/>
+      <param pos="0" name="os.product" value="Mac OS X"/>
+      <param pos="0" name="os.device" value="General"/>
+      <param pos="0" name="os.version" value="10.3"/>
+      <param pos="0" name="service.product" value="QTSS"/>
+      <param pos="1" name="service.version"/>
+    </fingerprint>
+
+    <fingerprint pattern="^QTSS\/([\d*.]*) \(Build\/[\d*.]*; Platform\/MacOSX; Release\/Mac OS X">
```
jhart-r7: Here and the other places where you are matching version-like strings, I think `[\d*.]*` could be improved -- the `*` inside will match the literal *, the `.` should be explicitly escaped and the  `*` at the end should probably be a `+`.  I'd suggest `[\d\.]+`.

xml/smb_native_os.xml
```diff
@@ -384,6 +384,33 @@
       <param pos="2" name="os.build"/>
    </fingerprint>
 
+      <!-- Mac OS X -->
+   <fingerprint pattern="^Samba 3.0.28a-apple">
+      <description>OS X 10.6</description>
```
jhart-r7: Perhaps mention Samba in here

xml/smb_native_os.xml
```diff
@@ -384,6 +384,33 @@
...
+      <param pos="0" name="os.family" value="Mac OS X"/>
+      <param pos="0" name="os.product" value="Mac OS X"/>
+      <param pos="0" name="os.device" value="General"/>
+      <param pos="0" name="os.version" value="10.6"/>
+      <param pos="0" name="service.vendor" value="Samba"/>
+      <param pos="0" name="service.product" value="Samba"/>
+      <param pos="0" name="service.version" value="3.0.28a-apple"/>
+   </fingerprint>
+
+   <fingerprint pattern="^Samba 3.0.25b-apple">
```
jhart-r7: Escape those periods and add a trailing anchor.  Also, even though you are looking for an explicit version, IMO it is simpler, cleaner and easier to add more in the future if you capture the version up here and use it in the `param` down below with a `pos` of 1.

xml/http_servers.xml
jkennedy-r7: Fixed

xml/smb_native_os.xml
jkennedy-r7: Added

xml/smb_native_os.xml
jkennedy-r7: Addressed

xml/smb_native_os.xml
```diff
@@ -384,6 +384,33 @@
       <param pos="2" name="os.build"/>
    </fingerprint>
 
+      <!-- Mac OS X -->
+   <fingerprint pattern="^Samba (3\.0\.28a-apple)$">
```
jhart-r7: Better, but I'm not 100% sure we want to be including the `-apple` part in the version.  There are pros and cons to doing it and not doing it.  I don't have a strong opinion either way but it is worth thinking about for the future.

jkennedy-r7: I can get both the point of including it or not. I guess we might want the version recored in the same way authenticated version check is done. If you run `smbd --version`,  the version string returned contains the `-apple`.

jhart-r7: The `smbd --version` point is proof enough.  Lets keep it as is.

To the point about authenticated version checks, keep in mind that recog is a generic framework that is not specific to Nexpose, metasploit or even necessarily any Rapid7 product (although today they are the primary consumers).  We need to ensure that any changes we make here are not overloaded with things that are product specific.

mdobrska-r7: :shipit:

Issue #90, opened 2015-09-01
Incorrect formatted xml fingerprint not caught by recog_verify


jkennedy-r7:
A malformatted fingerprints with a pos greater than 0 and a "value" attribute passes the recog_verify.
The fingerprint below passes the verify without an error or warning but is not imported by the protocol-fingerprinter and nexpose. 

```xml
   <fingerprint pattern="^Darwin (\S*) \S* Darwin Kernel Version 7.(\d+).\d+: .*RELEASE_\S* ?\s* (\S* ?\S*)$">
      <description>Mac OS X 10.3 (Panther)</description>
      <example host.name="mac-10-3-s-u" os.version.version="0">Darwin mac-10-3-s-u 7.0.0 Darwin Kernel Version 7.0.0: Wed Sep 24 15:48:39 PDT 2003; root:xnu/xnu-517.obj~1/RELEASE_PPC Power Macintosh</example>
      <param pos="1" name="host.name"/>
      <param pos="0" name="os.vendor" value="Apple"/>
      <param pos="0" name="os.family" value="Mac OS X"/>
      <param pos="0" name="os.product" value="Mac OS X"/>
      <param pos="0" name="os.device" value="General"/>
      <param pos="3" name="os.arch" value="Power Macintosh"/>
      <param pos="0" name="os.version" value="10.3"/>
      <param pos="2" name="os.version.version"/>
      <param pos="0" name="os.certainty" value="0.9"/>
   </fingerprint>
```

hmoore-r7:
The value tag not being in quotes is problematic as well.

jkennedy-r7:
Sorry HD, it was just me missing the quotes when I recreated the fingerprint.

hmoore-r7:
No worries, I think unquoted values may be worth validating as well.

jhart-r7:
@jkennedy-r7 and I talked about this outside of github.  #84 will help with some of this, but not (as currently written) with the issue of `param` with a non-zero `pos` and a non-empty `value`

PR #93, opened 2015-09-10
Update PHP regex to work with ZendServer string (X-POWERED-BY header)

gschneider-r7:
Based on the header `X-POWERED-BY: PHP/5.3.14 ZendServer/5.0`

I don't know what else besides ZendServer could appear in these headers. I also didn't investigate the difficulty (or possibility) of also capturing the ZendServer name and version at the same time as PHP in this fingerprint. I did not see anything that could conflict on this, but I didn't look very hard either. It might be worth looking through scans.io or Sonar data to see if this could be further improved for other situations.

Tests:
- [x] Ran `bin/recog_verify xml/http_servers.xml` and passed
- [x] Ran `bin/recog_match xml/http_servers.xml` and used STDIN to test some examples:
```
$ bin/recog_match -f d -c xml/http_servers.xml
PHP/5.3.14 ZendServer/5.0
MATCH: {"matched"=>"PHP", "service.component.product"=>"PHP", "service.component.version"=>"5.3.14", "data"=>"PHP/5.3.14 ZendServer/5.0"}
PHP/4.4.2-1build1
MATCH: {"matched"=>"PHP", "service.component.product"=>"PHP", "service.component.version"=>"4.4.2-1build1", "data"=>"PHP/4.4.2-1build1"}
```

### DIFF
--- a/xml/http_servers.xml
+++ b/xml/http_servers.xml
@@ -1766,11 +1766,18 @@
       <param pos="1" name="service.version"/>
    </fingerprint>
 
-   <fingerprint pattern="^PHP/(\S+)(?:\s.+)?$">
+   <fingerprint pattern="^PHP/(\S+)$">
       <example service.component.version="4.4.2-1build1">PHP/4.4.2-1build1</example>
       <example service.component.version="5.1.2">PHP/5.1.2</example>
-      <example service.component.version="5.3.14">PHP/5.3.14 ZendServer/5.0</example>
       <description>PHP</description>
+      <param pos="0" name="service.component.product" value="PHP"/>
+      <param pos="1" name="service.component.version"/>
+   </fingerprint>
+
+   <!-- TODO: Capture ZendServer version in fingerprint -->
+   <fingerprint pattern="^PHP/(\S+)\s+ZendServer/\S+$">
+      <example service.component.version="5.3.14">PHP/5.3.14 ZendServer/5.0</example>
+      <description>PHP with ZendServer</description>
       <param pos="0" name="service.component.product" value="PHP"/>
       <param pos="1" name="service.component.version"/>
    </fingerprint>
--- a/xml/http_servers.xml
+++ b/xml/http_servers.xml
@@ -1766,9 +1766,10 @@
       <param pos="1" name="service.version"/>
    </fingerprint>
 
-   <fingerprint pattern="^PHP/(\S+)$">
-      <example>PHP/4.4.2-1build1</example>
-      <example>PHP/5.1.2</example>
+   <fingerprint pattern="^PHP/(\S+)(\s.+)?$">
+      <example service.component.version="4.4.2-1build1">PHP/4.4.2-1build1</example>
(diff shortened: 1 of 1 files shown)


gschneider-r7:
I guess I should run `rake tests` in addition to verify and match, huh? Although it seems deceptive that verify and match are cool with what I did, but the rspec tests didn't like it.

jhart-r7:
I don't believe you ran into #90, but rather the long-standing disconnect between what can/is tested in rspec and what what can/is tested elsewhere.  I don't have a solution for that now other than updating the documentation to ensure/remind contributors to run the tests before submitting, which seems to a recurring problem.

xml/http_servers.xml
```diff
@@ -1766,9 +1766,10 @@
       <param pos="1" name="service.version"/>
    </fingerprint>
 
-   <fingerprint pattern="^PHP/(\S+)$">
-      <example>PHP/4.4.2-1build1</example>
-      <example>PHP/5.1.2</example>
+   <fingerprint pattern="^PHP/(\S+)(?:\s.+)?$">
+      <example service.component.version="4.4.2-1build1">PHP/4.4.2-1build1</example>
+      <example service.component.version="5.1.2">PHP/5.1.2</example>
+      <example service.component.version="5.3.14">PHP/5.3.14 ZendServer/5.0</example>
```
jhart-r7: IMO, if the only current example of a new "banner" that doesn't match what is already here only includes the ZendServer information, IMO it is worth adding a new fingerprint that explicitly matches PHP+ZendServer and also asserts ZendServer and the version.

gschneider-r7: Have an example of asserting a component of a component? Or is there a more sane-sounding way to do that?

The best thing I can think of otherwise is having 2 fingerprint matchers, but that requires the consumer to run the header string through the matchers twice...

jhart-r7: Yeah, I don't believe there is a good way to be doing this today, in Recog or the projects that utilize Recog.

For now, I'd suggest still breaking these into two fingerprints, but in the one that will match ZendZerver, don't assert ZendServer and simply leave a TODO comment.

This will help ensure that if other things come along which are not ZendServer they will not match and we'll stand a chance of identifying these mismatches and handling them as needed rather than letting them get absorbed in a catch-all.

gschneider-r7: For the new fingerprint, should it explicitly match on ZendServer or use this same catch-all regex?

jhart-r7: IMO, just match ZendServer.  Catch-alls are nice in that they will match in unknown situations, but are bad in that the handling of these unknown situations may need improvement and the catch all could hide it.

Take OpenSSH recently, as an example.  There were Ubuntu-version specific fingerprints for OpenSSH, and then a generic OpenSSH catch-all.  Unfortunately, there were other Ubuntu OpenSSH versions that we did not have an explicit match for which were being matched by the catch all, resulting in us missing the opportunity for getting an OS fingerprint.

PR #94, opened 2015-09-10
Enable use of Travis-CI's container infrastructure

gschneider-r7:
This may improve performance for tests/builds on Travis-CI. Faster build starts and better caching of dependencies are the high points of this change. I don't believe Recog requires anything that is restricted in the container environment so this should Just Work (:tm:).

See http://docs.travis-ci.com/user/migrating-from-legacy/ for more info.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: ruby
+sudo: false
 cache: bundler
 rvm:
   - 2.1.5


gschneider-r7:
Small sample size so far, but for Ruby 2.1.5 and 2.2.0 the tests completed in ~2.5 minutes on each Ruby for this PR's build. Compared to #93 which took ~4.5 minutes per Ruby version.

Looking at build history seems average for non-failing builds is ~8 minutes so this could be cutting build/test times in half going forward.

PR #95, opened 2015-09-25
Update recog_verify to catch invalid params, be more useful

jhart-r7:
This fixes @jkennedy-r7's #90.


gwiseman-r7:
:anchor: ship it

PR #97, opened 2015-10-09
Split out SMB Native LM fingerprints

### LINKED ISSUE #96: Add SMB native LM fingerprints
As indicated by the name, smb_native_os.xml is for fingerprinting devices based on the "native OS" field of the SMB response(s).  In trying to provide some testing for https://github.com/rapid7/metasploit-framework/pull/6005, I discovered that the fingerprints in https://github.com/jkennedy-r7/recog/blob/1c7456ae4dec863ddc92566bd6dc816c3af17426/xml/smb_native_os.xml#L387-L426 don't work -- these "banners" seem to be taken from the SMB "native LM" field and will not work.  It may be the case that other existing fingerprints in that file are also similarly problematic, but I don't believe so.

Assuming this is correct, we should move them to a new file and add the corresponding improvement to metasploit (https://github.com/rapid7/metasploit-framework/issues/6021)

jhart-r7:
Fixes #96

Eventually all of the fingerprints in smb_native_os will need to have their counterparts added in smb_native_lm.

Specs pass, tested by locally releasing a newer version of recog with this new fingerprint in it, then using that new version of recog in a local fix for https://github.com/rapid7/metasploit-framework/issues/6021 and saw it also using the SMB Native LM fields for fingerprinting.

### DIFF
--- a/spec/data/smb_native_os.txt
+++ b/spec/data/smb_native_os.txt
@@ -23,9 +23,3 @@ Windows 7 Starter 7601 Service Pack 1
 Windows 7 Home Premium 7600
 Windows 7 Enterprise 7601 Service Pack 1
 Windows 7 Enterprise 7600
-Samba 3.6.9-151.el6_4.1
-Samba 3.6.6
-Samba 3.6.3
-Samba 3.0.32-0.2-2210-SUSE-SL10.3
-Samba 3.0.28a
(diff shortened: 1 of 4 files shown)


jyeung-r7:
:heart:

jkennedy-r7:
Looks good. Ship it

PR #99, opened 2015-11-18
Implement multi-match for recog

lholloway-r7:
This PR is to implement simple multi-match for recog.  Instead of stopping at the first match it will continue through the fingerprints attempting to match against each one.

It is controlled through the use of the `--multi-match` flag on the command line or through passing a boolean variable in the options object if using recog programmatically.

### DIFF
--- a/lib/recog/matcher.rb
+++ b/lib/recog/matcher.rb
@@ -29,17 +29,25 @@ class Matcher
         line = line.to_s.unpack("C*").pack("C*").strip.gsub(/\\[rn]/, '')
         found_extractions = false
 
+        all_extractions = []
         fingerprints.each do |fp|
           extractions = fp.match(line)
           if extractions
             found_extractions = true
             extractions['data'] = line
-            reporter.match "MATCH: #{extractions.inspect}"
-            break unless multi_match
+            if multi_match
+              all_extractions << extractions
+            else
+              reporter.match "MATCH: #{extractions.inspect}"
+              break
+            end
           end
         end
 
-        unless found_extractions
+        if found_extractions
+          match_prefix = all_extractions.size > 1 ? 'MATCHES' : 'MATCH'
+          reporter.match "#{match_prefix}: #{all_extractions.map(&:inspect).join(',')}" if multi_match
+        else
           reporter.failure "FAIL: #{line}"
         end
 
--- a/features/match.feature
+++ b/features/match.feature
@@ -19,10 +19,8 @@ Feature: Match
     When I run `recog_match multiple_banners_fingerprints.xml sample_banner.txt --multi-match`
     Then it should pass with:
       """
-      MATCH: {"matched"=>"Generic FTP, Checks for the existence of the word FTP in the line", "data"=>"---------- Welcome to Pure-FTPd [privsep] [TLS] ----------"}
-      MATCH: {"matched"=>"Pure-FTPd Config data can be zero or more of: [privsep] [TLS]", "pureftpd.config"=>"[privsep] [TLS] ", "service.family"=>"Pure-FTPd", "service.product"=>"Pure-FTPd", "data"=>"---------- Welcome to Pure-FTPd [privsep] [TLS] ----------"}
-      MATCH: {"matched"=>"Generic FTP, Checks for the existence of the word FTP in the line", "data"=>"polaris FTP server (SunOS 5.8) ready."}
-      MATCH: {"matched"=>"SunOS/Solaris", "os.vendor"=>"Sun", "os.family"=>"Solaris", "os.product"=>"Solaris", "os.device"=>"General", "host.name"=>"polaris", "os.version"=>"5.8", "data"=>"polaris FTP server (SunOS 5.8) ready."}
+      MATCHES: {"matched"=>"Generic FTP, Checks for the existence of the word FTP in the line", "data"=>"---------- Welcome to Pure-FTPd [privsep] [TLS] ----------"},{"matched"=>"Pure-FTPd Config data can be zero or more of: [privsep] [TLS]", "pureftpd.config"=>"[privsep] [TLS] ", "service.family"=>"Pure-FTPd", "service.product"=>"Pure-FTPd", "data"=>"---------- Welcome to Pure-FTPd [privsep] [TLS] ----------"}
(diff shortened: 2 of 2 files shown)


bin/recog_match
```diff
@@ -33,6 +33,10 @@ option_parser = OptionParser.new do |opts|
     options.color = true
   end
 
+  opts.on("--multi-match", "Enable multiple matches") do
```
jhart-r7: I'd suggest:

```
opts.on("--[no-]multi-match", "Enable or disable multiple matches (defaults to disabled)") do |o|
  options.multi_match = o
end
```

This tells the user what the default is and gives them the ability to be explicit in its operation.

lib/recog/matcher.rb
```diff
@@ -23,20 +27,25 @@ def match_banners(banners_file)
 
         line = line.to_s.unpack("C*").pack("C*").strip.gsub(/\\[rn]/, '')
         extractions = nil
+        foundExtractions = false
```
jhart-r7: Minor style nit, but snake case (`found_extractions`) is preferred.

lib/recog/matcher.rb
```diff
@@ -23,20 +27,25 @@ def match_banners(banners_file)
...
         fingerprints.each do |fp|
-          break if (extractions = fp.match(line))
+          extractions = fp.match(line)
+          if extractions
+            foundExtractions = true
+            extractions['data'] = line
+            reporter.match "MATCH: #{extractions.inspect}"
+            break if(!multi_match)
```
jhart-r7: `break unless multi_match` is more Ruby.

lib/recog/matcher.rb
```diff
@@ -23,20 +27,25 @@ def match_banners(banners_file)
...
+          end
         end
 
-        if extractions
-          extractions['data'] = line
-          reporter.match "MATCH: #{extractions.inspect}"
-        else
+        if (!foundExtractions)
```
jhart-r7: `unless found_extractions`

lib/recog/matcher.rb
```diff
@@ -23,20 +27,25 @@ def match_banners(banners_file)
 
         line = line.to_s.unpack("C*").pack("C*").strip.gsub(/\\[rn]/, '')
         extractions = nil
```
jhart-r7: This can now be removed since `extractions` is used inside the loop only.

lib/recog/matcher.rb
lholloway-r7: :+1:, I'll fix it.

lib/recog/matcher.rb
```diff
@@ -1,10 +1,14 @@
 module Recog
 class Matcher
-  attr_reader :fingerprints, :reporter
+  attr_reader :fingerprints, :reporter, :multi_match
 
-  def initialize(fingerprints, reporter)
+  # @param fingerprints Array of [Recog::Fingerprint]
+  # @param reporter [Recog::MatchReporter]
+  # @param multi_match [Boolean]
```
jhart-r7: Thanks for adding the docs.  Could you also describe the parameters while here?

bin/recog_match
lholloway-r7: Agree, will fix.

jhart-r7:
This looks reasonable.  The only thing I think we need to address before landing is the tests -- right now you are only testing the `recog_match` executable that we ship but not the API.    We should ensure that the API defaults to non-multimatch but supports multimatch, and that both have tests in `spec` to ensure this.

lholloway-r7:
Good catch, I will add a method to Nizer.rb to allow multi-match and provide spec tests for it.

lholloway-r7:
I was thinking nil on return for `multi_match` to maintain the same sort of contract that was done with `match` but you're right, I'd rather it just be returned as an empty list.

lib/recog/nizer.rb
```diff
@@ -41,6 +41,24 @@ def self.match(match_key, match_string)
     nil
   end
 
+  def self.multi_match(match_key, match_string)
```
jhart-r7: I would suggest just adding this on to `self.match` much in the same way you changed initialize in `lib/recog/matcher.rb`, except this time make it take an optional, off-by default third argument to turn multi-match on/off.

Metasploit uses Recog::Nizer.match to do most of its work and it would be nice to just have one method, not two.

spec/lib/recog/nizer_spec.rb
```diff
@@ -28,6 +28,33 @@
     end
   end
 
+  describe ".multi_match" do
```
jhart-r7: Regardless of whether or not you just collapse this into `match`, you should test the negative case where there are no matches.

lib/recog/nizer.rb
lholloway-r7: I implemented the second method as multi-match because using `match` with single match returns a single object that is the matched fingerprint.  If we change it to multi match it becomes an array of matched fingerprints.  If we were to change it to always return an Array of fingerprints, I agree we could make it work... but then we break Metasploit.

spec/lib/recog/nizer_spec.rb
lholloway-r7: I'll setup a negative test case.

lib/recog/nizer.rb
jhart-r7: That is OK -- Metasploit pins its version of recog at a specific version, so even if we were to release this in a minor version, they wouldn't get it until they explicitly bumped.  But the right way to do this would be to do the API-breaking change in a major version bump.  

For now, though, yes, lets keep it as is.  This gets us what we need short term, doesn't break metasploit, allows us to deal with the problems that multi match will likely uncover and conveniently avoids the problem of how metasploit would handle multiple matches (can be done, but requires work on that end).

lib/recog/matcher.rb
```diff
@@ -22,21 +27,26 @@ def match_banners(banners_file)
...
-        extractions = nil
+        found_extractions = false
+
         fingerprints.each do |fp|
-          break if (extractions = fp.match(line))
+          extractions = fp.match(line)
+          if extractions
+            found_extractions = true
+            extractions['data'] = line
+            reporter.match "MATCH: #{extractions.inspect}"
```
jhart-r7: In testing this locally, I realized that the multi match functionality works but perhaps not in a way that allows us to deal with addressing the multiple matches issue in the fingerprints because `recog_match` does not make it obvious in the output whether or not there were multiple matches.  As written, it will output one line per matching fingerprint found against the current input string, but because `recog_match` can handle reading from STDIN or reading multiple banners from a file, the output very quickly becomes cluttered and you won't be able to tell what belongs to what or when multiple matches happen.  I'll shoot you a PR to correct this.

PR #104, opened 2016-03-02
Adding support for modern Nexpose fingerprint matching

mhuffman-r7:
Adding an additional fingerprint pattern for the Nexpose Security Console that uses a new pattern to be changed in an upcoming release of Nexpose. This pattern is version-less.

### DIFF
--- a/xml/http_servers.xml
+++ b/xml/http_servers.xml
@@ -2813,7 +2813,7 @@
     <description>Rapid7 NSC</description>
     <example>NSC/0.6.4 (JVM)</example>
     <param pos="0" name="service.vendor" value="Rapid7"/>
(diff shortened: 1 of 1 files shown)


xml/http_servers.xml
```diff
@@ -2815,6 +2815,12 @@
     <param pos="0" name="service.vendor" value="Rapid7"/>
     <param pos="0" name="service.product" value="NeXpose"/>
```
jhart-r7: While here, we should probably correct this.

mhuffman-r7: I can do that. No problem.

jhart-r7:
Thank you for the contribution, @mhuffman-r7.

PR #105, opened 2016-03-19
Changes to MySQL banners for Nexpose support

gschneider-r7:
Changed uses of `\h` to ~~`[a-fA-F\d]`~~ (update: now `[a-f\d]` with case-insensitive flag) since `\h` seems to be a ruby-specific thing and Nexpose could not match with it.

~~Added `os.device` to fingerprints that have OS info.~~

~~Added `os.certainty` to fingerprints that have OS versions. Somewhat arbitrarily chose 0.80 as likely to be useful and not likely to interfere with potentially superior fingerprints from SSH or other services.~~

One thing I wonder about is switching the MariaDB version match group to the 2nd set of numbers, but that probably means lots of edits/new fingerprints to keep compatibility with banners that don't have the 2nd set. **Update 4/13: Done** - I swapped the optional groups around for MariaDB so that 10.x.x versions will capture instead of 5.5.5, but 5.x.x versions can still capture when there is no 10.x.x version.

cc @jhart-r7 @pdeardorff-r7 @TomSellers (basically everyone who touched this file and might have input)

### DIFF
--- a/xml/mysql_banners.xml
+++ b/xml/mysql_banners.xml
@@ -62,7 +62,6 @@
     <param pos="0" name="service.edition" value="Community Edition"/>
     <param pos="0" name="os.vendor" value="Microsoft"/>
     <param pos="0" name="os.family" value="Windows"/>
-    <param pos="0" name="os.device" value="General"/>
   </fingerprint>
   <fingerprint pattern="^(\d{1,2}\.\d{1,3}\.[a-f\d]{1,3})(?:-alpha)?(?:-beta)?(?:-rc)?(?:-gamma)?(?:-max)?-nt(?:-max)?(?:-log)?$" flags="REG_ICASE">
     <description>Oracle MySQL with Named Pipes (Windows)</description>
@@ -77,7 +76,6 @@
     <param pos="0" name="service.product" value="MySQL"/>
     <param pos="0" name="os.vendor" value="Microsoft"/>
     <param pos="0" name="os.family" value="Windows"/>
-    <param pos="0" name="os.device" value="General"/>
   </fingerprint>
   <fingerprint pattern="^(\d{1,2}\.\d{1,3}\.[a-f\d]{1,3})-cll$" flags="REG_ICASE">
     <description>Oracle MySQL hosted on CPanel</description>
@@ -87,7 +85,6 @@
     <param pos="0" name="service.family" value="MySQL"/>
     <param pos="0" name="service.product" value="MySQL"/>
     <param pos="0" name="os.family" value="Linux"/>
-    <param pos="0" name="os.device" value="General"/>
   </fingerprint>
   <fingerprint pattern="^(\d{1,2}\.\d{1,3}\.[a-f\d]{1,3})-community-cll$" flags="REG_ICASE">
     <description>Oracle MySQL Community hosted on CPanel</description>
@@ -99,7 +96,6 @@
     <param pos="0" name="service.edition" value="Community Edition"/>
     <param pos="0" name="os.family" value="Linux"/>
     <param pos="0" name="os.product" value="Linux"/>
-    <param pos="0" name="os.device" value="General"/>
   </fingerprint>
   <fingerprint pattern="^(\d{1,2}\.\d{1,3}\.[a-f\d]{1,3})-cll-lve$" flags="REG_ICASE">
     <description>Oracle MySQL on CloudLinux in a Lightweight Virtual Environment (LVE)</description>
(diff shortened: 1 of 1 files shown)


pmiseiko-r7:
You didn't need to add OS certainty.  You specified the default OS certainty in the protocol evaluator.  Note that the default you specified was one of 0.7 (error) or 0.75 (banner).  Probably better to remove explicit os.certainty definitions and let the default implicitly define the OS certainty.

You also didn't need os.device, but no harm in adding it.

gschneider-r7:
To me some of the fingerprints here have a better certainty for OS than others, and the protocol evaluator would be less trusting for any that aren't explicitly defined. Overkill I guess.

os.device was just because I didn't like seeing the empty column in the fingerprints table ;)

gschneider-r7:
I changed the patterns for MariaDB to prefer the 10.x.x versions over 5.5.5, while maintaining support for actual 5.x.x versions.

tomhart-r7:
LGTM

xml/mysql_banners.xml
```diff
@@ -132,6 +138,8 @@
     <param pos="0" name="os.family" value="Linux"/>
     <param pos="0" name="os.product" value="Linux"/>
     <param pos="2" name="os.version"/>
+    <param pos="0" name="os.certainty" value="0.80"/>
```
jhart-r7: Why are you adding a certainty here but not everywhere else?

xml/mysql_banners.xml
```diff
@@ -175,6 +186,8 @@
     <param pos="0" name="os.family" value="Linux"/>
     <param pos="0" name="os.product" value="Linux"/>
     <param pos="0" name="os.version" value="12.04"/>
+    <param pos="0" name="os.certainty" value="0.80"/>
+    <param pos="0" name="os.device" value="General"/>
```
jhart-r7: IMO, we should only set `os.device` to something if it is unique in some way -- `General` is too generic.

xml/mysql_banners.xml
```diff
@@ -1107,8 +1208,9 @@
     <param pos="0" name="service.product" value="Percona Server"/>
     <param pos="0" name="os.family" value="Linux"/>
     <param pos="0" name="os.product" value="Linux"/>
+    <param pos="0" name="os.device" value="General"/>
   </fingerprint>
-  <fingerprint pattern="^(\d{1,2}\.\d{1,3}\.\h{1,3}) Complete MySQL by Server Logistics(?:-log)?$">
+  <fingerprint pattern="^(\d{1,2}\.\d{1,3}\.[a-fA-F\d]{1,3}) Complete MySQL by Server Logistics(?:-log)?$">
```
jhart-r7: Rather than using `[a-fA-F]`, it might be better to make the whole regex case insensitive.  That simplifies it slightly and heads off the almost inevitable case in the future where a few characters differ in case and this misfingerprints as a result.

xml/mysql_banners.xml
gschneider-r7: I felt that some fingerprints can be more certain of OS than others - should I just remove all certainty and let the protocol helper handle it?

xml/mysql_banners.xml
gschneider-r7: Sure, I can remove these. I was trying to keep it in line with (most) desktop/server OS fingerprints in Nexpose.

xml/mysql_banners.xml
gschneider-r7: Will do.

xml/mysql_banners.xml
jhart-r7: More certain or more complete?  I don't want to derail this review, so for now I'd say don't make any changes to the existing certainty values for mysql in recog.  Eventually I'd like to see the certainty values rethought and I imagine what will come of that is that the certainty values will be removed from recog.  Obviously that is a big change with a big impact on things that integrate recog so it will take some time and discussions.

gschneider-r7: I _think_ the only certainty values for MySQL fingerprints were added by me, specifically for OS.

xml/mysql_banners.xml
gschneider-r7: Yeah that was probably me on this same PR using the "I do it so the UI looks better" excuse. :smiley:

gschneider-r7:
I've updated the case-sensitive stuff and removed the additions I made for os.device and os.certainty. Let me know if you'd like me to squash commits before landing.

jhart-r7:
Looks good.  Whether or not the mariadb version is equivalent to the mysql version is TBD, I guess, but for now I think it makes sense to keep them separate.  If, in the future, someone proves otherwise or shows evidence that having both would be better, we can change. Landing.

jhart-r7:
Oh, I should also note that in general we should be very hesitant to change things in recog so that they work in Nexpose or they work better/properly in Nexpose.  Recog is a standalone project used by (at least) Metasploit and Nexpose and as such it must remain as independent as possible.  In this case, moving to a more compatible regular expression format is perfectly OK and has been done before.

PR #107, opened 2016-03-30
SMB banners for VisionFS and OS/400

jkennedy-r7:
Banners for VisionFS and OS/400

According to our code comments VisionFS returns [AR][OS][VFVR] which is arch, os and VisionFS version. The remainder of the visionFS banner contains the "uname -a" output, such as: `[abbrev] [uname -a output...`:
```
i3un0210 UnixWare unixware7 5 7 i386
```

After discussion within the team, we decided to error on the safe side and only match on the [abbrev].

RSpec output:
```rspec
Finished in 53.32 seconds (files took 1.9 seconds to load)
17868 examples, 0 failures

Randomized with seed 64103

/opt/ruby2.2/lib/ruby/gems/2.2.0/gems/simplecov-0.10.0/lib/simplecov/defaults.rb:50: warning: global variable `$ERROR_INFO' not initialized
Coverage report generated for RSpec to /home/jkennedy/rapid7/recog/coverage. 690 / 722 LOC (95.57%) covered.
```

### DIFF
--- a/xml/smb_native_os.xml
+++ b/xml/smb_native_os.xml
@@ -364,9 +364,9 @@
 
   <!-- VisionFS -->
 
-  <fingerprint pattern="^(?:ax|i3|m8|mp|pa|pp|rs|sp)ai([\d]+4)">
+  <fingerprint pattern="^(?:ax|i3|m8|mp|pa|pp|rs|sp)ai(\d{4})">
     <description>AIX</description>
-    <example service.version="1234">axai1234</example>
+    <example service.version="9876">axai9876</example>
     <param pos="0" name="os.vendor" value="IBM"/>
     <param pos="0" name="os.family" value="AIX"/>
     <param pos="0" name="os.product" value="AIX"/>
@@ -374,9 +374,9 @@
     <param pos="1" name="service.version"/>
   </fingerprint>
 
-  <fingerprint pattern="^(?:ax|i3|m8|mp|pa|pp|rs|sp)dg([\d]+4)">
+  <fingerprint pattern="^(?:ax|i3|m8|mp|pa|pp|rs|sp)dg(\d{4})">
     <description>DG/UX</description>
-    <example service.version="1234">i3dg1234</example>
+    <example service.version="9876">i3dg9876</example>
     <param pos="0" name="os.vendor" value="Data General"/>
     <param pos="0" name="os.family" value="DG/UX"/>
     <param pos="0" name="os.product" value="DG/UX"/>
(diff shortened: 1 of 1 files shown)


rkirk-r7:
ship it

bobbymckeown-r7:
:shipit:

xml/smb_native_os.xml
```diff
@@ -351,4 +351,196 @@
     <param pos="0" name="service.vendor" value="Wind River"/>
     <param pos="0" name="service.product" value="VxWorks CIFS"/>
   </fingerprint>
+
+  <fingerprint pattern="^OS/400 \D(\d+)\D(\d+)\D(\d+)">
```
jhart-r7: For this and most others, it might be worth adding the trailing `$` to the regex.

xml/smb_native_os.xml
```diff
@@ -351,4 +351,196 @@
...
+    <param pos="0" name="os.vendor" value="IBM"/>
+    <param pos="0" name="os.product" value="OS/400"/>
+    <param pos="1" name="os.version"/>
+    <param pos="2" name="os.version.version"/>
+    <param pos="3" name="os.version.version.version"/>
+  </fingerprint>
+
+  <!-- VisionFS -->
+
+  <fingerprint pattern="^(?:ax|i3|m8|mp|pa|pp|rs|sp)ai([\d]+4)">
```
jhart-r7: Is the `4` at the end expected?  Or are you trying to match anything with 4 digits at the end?

jhart-r7:
Looks good to me minus my one concern above.

Outside the scope of recog, if the remainder of the banner for visionfs is `uname -a`, you should capture, store and fingerprint of of that as well.

xml/smb_native_os.xml
jkennedy-r7: Should match 4 digits. Fixed it.

jkennedy-r7:
I didn't add the uname part because I don't have access all these banners. I don't want to miss a matching because the "hacked up" banner is incorrect.

jhart-r7:
Looks good to me.  

Regarding the `uname` stuff, to be clear, I'm saying that _something else_, **not recog**, should probably parse this `uname` output when it detects a VisionFS-based SMB instance.  To allow this to happen, you could capture it now, store it as, say, `smb.visionfs.uname_stuff` and then recog's data/interface's will return the `uname` output to you if you add it as a capture in your regex.

jhart-r7:
LGTM, landing and releasing now because @gschneider-r7 asked for #105.

PR #108, opened 2016-03-30
SMB banners for Windows

jkennedy-r7:
LM Banners for Windows. This regex is required to maintain current behaviour of Nexpose. We might create more specific banners later on.

### DIFF
--- a/xml/smb_native_lm.xml
+++ b/xml/smb_native_lm.xml
@@ -54,6 +54,7 @@
     <example service.product="Windows Server 2012 R2 Standard 6.3">Windows Server 2012 R2 Standard 6.3</example>
     <param pos="0" name="service.vendor" value="Microsoft"/>
     <param pos="1" name="service.product"/>
+    <param pos="0" name="service.certainty" value="1.00"/>
   </fingerprint>
 
 </fingerprints>
--- a/xml/smb_native_lm.xml
+++ b/xml/smb_native_lm.xml
@@ -41,6 +41,7 @@
     <param pos="0" name="service.vendor" value="Samba"/>
     <param pos="0" name="service.product" value="Samba"/>
     <param pos="1" name="service.version"/>
+    <param pos="0" name="service.certainty" value="1.00"/>
   </fingerprint>
 
   <!-- Windows -->
--- a/xml/smb_native_lm.xml
+++ b/xml/smb_native_lm.xml
@@ -42,4 +42,18 @@
     <param pos="0" name="service.product" value="Samba"/>
     <param pos="1" name="service.version"/>
   </fingerprint>
+
+  <!-- Windows -->
+  <fingerprint pattern="^(Windows [\W*\w*]*)$">
+    <description>Windows</description>
+    <example service.product="Windows Server 2008 R2 Enterprise 6.1">Windows Server 2008 R2 Enterprise 6.1</example>
+    <example service.product="Windows 7 Enterprise 6.1">Windows 7 Enterprise 6.1</example>
+    <example service.product="Windows Server 2003 R2 5.2">Windows Server 2003 R2 5.2</example>
+    <example service.product="Windows Server 2012 Standard 6.2">Windows Server 2012 Standard 6.2</example>
+    <example service.product="Windows 8 Enterprise 6.2">Windows 8 Enterprise 6.2</example>
+    <example service.product="Windows Server 2012 R2 Standard 6.3">Windows Server 2012 R2 Standard 6.3</example>
+    <param pos="0" name="service.vendor" value="Microsoft"/>
+    <param pos="1" name="service.product"/>
+  </fingerprint>
+
 </fingerprints>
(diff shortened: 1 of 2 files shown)


cmccrisken-r7:
:ship: it

rkirk-r7:
ship it

rkirk-r7:
Removed my ship it as the PR has changed :smile: 

My comment from the merged PR is still valid, the device classification should also be added to the windows fingerprints, as nexpose's existing CIFS fingerprinting would assert this field. Happy once this is addressed

xml/smb_native_lm.xml
```diff
@@ -41,5 +41,21 @@
     <param pos="0" name="service.vendor" value="Samba"/>
     <param pos="0" name="service.product" value="Samba"/>
     <param pos="1" name="service.version"/>
+    <param pos="0" name="service.certainty" value="1.00"/>
   </fingerprint>
+
+  <!-- Windows -->
+  <fingerprint pattern="^(Windows [\W*\w*]*)$">
+    <description>Windows</description>
+    <example service.product="Windows Server 2008 R2 Enterprise 6.1">Windows Server 2008 R2 Enterprise 6.1</example>
```
jhart-r7: Shouldn't these be `os.`?

xml/smb_native_os.xml
```diff
@@ -8,87 +8,108 @@
     <example os.product="Windows NT 4.0">Windows NT 4.0</example>
     <param pos="0" name="os.certainty" value="1.0"/>
     <param pos="0" name="os.vendor" value="Microsoft"/>
+    <param pos="0" name="os.family" value="Windows"/>
     <param pos="1" name="os.product"/>
+    <param pos="0" name="os.device" value="General"/>
```
jhart-r7: IMO, we should not be sprinkling `General` everywhere, especially if it is just to make the Nexpose UI less ugly in this regard.  IMO, if the banner allows us to determine something less generic, add it as `os.device`, otherwise leave it off.

xml/smb_native_lm.xml
```diff
@@ -41,5 +41,21 @@
     <param pos="0" name="service.vendor" value="Samba"/>
     <param pos="0" name="service.product" value="Samba"/>
     <param pos="1" name="service.version"/>
+    <param pos="0" name="service.certainty" value="1.00"/>
   </fingerprint>
+
+  <!-- Windows -->
+  <fingerprint pattern="^(Windows [\W*\w*]*)$">
```
jhart-r7: I agree that this is a simpler approach rather than having one regular expression per flavor of Windows, but the problem with this (especially if you convert these to using `os.` rather than `service.`) is that you are taking the Windows banner as-is and asserting that as the product with 100% certainty.  There are numerous problems with this approach, the biggest / most common of which is the quality of the `.product` since you are believing whatever is there.

xml/smb_native_lm.xml
jkennedy-r7: This is technically correct. It's matching the LAN manager which for Windows is Windows :smile:

xml/smb_native_lm.xml
jkennedy-r7: Which is, unless it's Samba, how Nexpose does it....

jhart-r7: True, but recog is not Nexpose.

xml/smb_native_lm.xml
```diff
@@ -41,5 +41,21 @@
...
+    <description>Windows</description>
+    <example service.product="Windows Server 2008 R2 Enterprise 6.1">Windows Server 2008 R2 Enterprise 6.1</example>
+    <example service.product="Windows 7 Enterprise 6.1">Windows 7 Enterprise 6.1</example>
+    <example service.product="Windows Server 2003 R2 5.2">Windows Server 2003 R2 5.2</example>
+    <example service.product="Windows Server 2012 Standard 6.2">Windows Server 2012 Standard 6.2</example>
+    <example service.product="Windows 8 Enterprise 6.2">Windows 8 Enterprise 6.2</example>
+    <example service.product="Windows Server 2012 R2 Standard 6.3">Windows Server 2012 R2 Standard 6.3</example>
+    <param pos="0" name="service.vendor" value="Microsoft"/>
+    <param pos="1" name="service.product"/>
+    <param pos="0" name="service.certainty" value="1.00"/>
```
jhart-r7: Non-recog related aside, but AFAIK service certainties do nothing in Nexpose.

xml/smb_native_lm.xml
jhart-r7: Well sure, but now you are missing an OS fingerprint.  

In general, IMO, for every "banner" that recog is capable of matching against, we should assert as much fingerprint information as possible.  The LM and Native OS strings here should allow to assert both an operating system and a service fingerprint.

xml/smb_native_lm.xml
jhart-r7: We should be very careful here when "recog-ifying" any remaining parts of Nexpose.  What ever ends up in Recog should be as generic and non-product (Nexpose) specific as possible.

jhart-r7:
@jkennedy-r7, closing this for now.  Feel free to re-open if this work picks up again.

PR #109, opened 2016-04-30
Generic matchers for Operating System names

gschneider-r7:
These are generic matchers for OS names that are meant to be used in any context. For example: from user input, reading config files, Windows registry, etc.

**Points of interest for review:**

Darwin/Mac OS X: 
- ~~Is darwin an appropriate name for the file?~~
- Is it useful to have a version match from the code name?

Linux:
- Lots of copy+paste, so some of the regexes may have redundant things

Windows:
- Monster regexes to match all the things! Should these be split up more?
- Does using the service pack as the version really make sense? (seems to be done this way elsewhere already)
- Do the editions make sense when they look like `Home, Premium N Edition`?

Additional OS types:

Where to draw the line at OS family? For example, should I keep BSD separate from Linux, or put all unix-likes into one file?

**To do:**
- [x] BSD Family
- [x] AIX Family
- [x] HP-UX Family
- [x] Cisco
- [x] Juniper
- [ ] Android (not doing for now)
- [x] Apple iOS


xml/operating_system_darwin.xml
```diff
@@ -0,0 +1,110 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Patterns for common names of Mac OS X operating systems.
+-->
+<fingerprints matches="operating_system.name">
+  <fingerprint pattern="^Mac OS X\s?(\d+?(?:\.\d+?)+?)?$" flags="REG_ICASE">
+    <description>Mac OS X with version number</description>
+    <example os.version="10.10.5">Mac OS X 10.10.5</example>
```
mhuffman-r7: Last I looked I think the examples are living test cases that get used to test each pattern. Is there a way to provide negative cases?

xml/operating_system_darwin.xml
```diff
@@ -0,0 +1,110 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Patterns for common names of Mac OS X operating systems.
+-->
+<fingerprints matches="operating_system.name">
+  <fingerprint pattern="^Mac OS X\s?(\d+?(?:\.\d+?)+?)?$" flags="REG_ICASE">
```
mhuffman-r7: So I think this will not match patterns like "Mac OS X 10", and "Mac OS 10.5". Should we make this a little less greedy. For example, how about making the "X" optional, and then using a * rather than + for the nested versions. I guess what I guess what I am favoring is a "weak" fingerprint vs no fingerprint.

gschneider-r7: You're right, I didn't consider only providing the major version. Simple enough to use a `*` instead of `+` on the repetition of the nested versions.

xml/operating_system_darwin.xml
gschneider-r7: I don't think there is currently a way to do negative tests within the fingerprints files - @jhart-r7 can you confirm that?

I suppose rspec tests could be added to test negative cases.

jhart-r7: There is no support for negative tests.  I would not suggest putting this in rspec.  If you have a fingerprint you want a pattern #N to not match, add a pattern before #N that explicitly matches it;  order is enforced and tested.

xml/operating_system_darwin.xml
```diff
@@ -0,0 +1,134 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Patterns for common names of Mac OS X operating systems.
+-->
+<fingerprints matches="operating_system.name">
+  <!-- Match Mac OS Classic first due to weak matching on Mac OS X -->
+  <fingerprint pattern="^(?:Apple )?Mac OS ([7-9](?:\.\d+?)*?)$" flags="REG_ICASE">
```
jhart-r7: btw, given that this is a new fingerprint file, it would be great to see it not make #75 (see: #83 too) worse.  IOW, remove `flags` and just make the regular expression case insensitive (non-capture group the entire thing, toss in the `i` option, like `^(?i:(?:Apple )?Mac OS ([7-9](?:\.\d+?)*?))$`

gschneider-r7: Sure I can do that. I take it that syntax style is supported by all known current consumers of Recog?

jhart-r7: Anything that uses Recog's interfaces (metasploit), yes, that will work.  For projects that utilize Recog's data (nexpose),  I'm not sure at this point.  Probably best to run this by @pmiseiko-r7 internally.

mhuffman-r7:
Is there a benefit to separating the fingerprints by quasi-vendor? I am just a bit worried this makes it slightly more difficult to consume all the OS-level matchers together.

gschneider-r7:
I think originally there was concern about the size/manageability of a single file. What I have so far seems fine for a single file actually. It probably won't get too much bigger with other OS families being added either.

jhart-r7:
For any integrating products that want to load all recog fingerprints of a particular type, you should find and filter them by using the top-level `matches` attribute.  Personally I like splitting these up even though there aren't that many now.

mhuffman-r7:
Logically this makes sense @jhart-r7. The only thing I am fighting is the current Java code to handle this is really coupled to a single file. We can obviously iterate over many of them and merge together, but its a PITA. Perhaps that just needs rewriting.

xml/operating_system_windows.xml
```diff
@@ -0,0 +1,98 @@
...
+  </fingerprint>
+  <fingerprint pattern="^(?i:(?:Microsoft )?(Windows\s?(?:95|98|98SE|98 SE|98 Second Edition|ME|Millenium Edition)))$">
+    <description>Windows 9x</description>
+    <example os.product="Windows 98 SE">Windows 98 SE</example>
+    <param pos="0" name="os.vendor" value="Microsoft"/>
+    <param pos="0" name="os.family" value="Windows"/>
+    <param pos="1" name="os.product"/>
+  </fingerprint>
+  <fingerprint pattern="^(?i:(?:Microsoft )?Windows(?:\s?\w+?)*?\s?(\d+?(?:\.\d+?)*?)?)$">
+    <description>Windows catch-all</description>
```
jhart-r7: Catch-alls are catch-22s.  On the one hand,  yeah, you'll at least get a Windows fingerprint, but what if "Windows 2020" comes out?  Then with this file you won't know that you are missing an explicit match for that OS.

pmiseiko-r7: If you insist on using a catch all give it a lower certainty which will ensure the matcher never returns the catch all when another match was made.

xml/operating_system_linux.xml
```diff
@@ -0,0 +1,252 @@
...
+  <!-- Linux catch-all goes at the bottom-->
+  <fingerprint pattern="^(?i:.*Linux?\s?(\d+?(?:\.\d+?)*?)?)$">
+    <description>Linux catch-all</description>
+    <example os.version="2.42.6">Linux 2.42.6</example>
+    <param pos="0" name="os.vendor" value="Linux"/>
+    <param pos="0" name="os.family" value="Linux"/>
+    <param pos="0" name="os.product" value="Linux"/>
+    <param pos="1" name="os.version"/>
+  </fingerprint>
+  <!-- other Unix-likes -->
```
pmiseiko-r7: I'm insulted that you put this under Linux. :)

xml/operating_system_linux.xml
```diff
@@ -0,0 +1,252 @@
...
+    <example os.version="12.04.4">Xubuntu 12.04.4 LTS</example>
+    <example os.version="14.04">Xubuntu Linux 14.04</example>
+    <example os.version="16.04" os.edition="LTS">Xubuntu 16.04 LTS</example>
+    <param pos="0" name="os.vendor" value="Xubuntu"/>
+    <param pos="0" name="os.family" value="Linux"/>
+    <param pos="0" name="os.product" value="Linux"/>
+    <param pos="1" name="os.version"/>
+    <param pos="2" name="os.edition"/>
+  </fingerprint>
+  <!-- Linux catch-all goes at the bottom-->
```
pmiseiko-r7: Same comment as the Windows catch all.

xml/operating_system_linux.xml
```diff
@@ -0,0 +1,252 @@
...
+  </fingerprint>
+  <fingerprint pattern="^(?i:Slackware(?: Linux)\s?(\d+?(?:\.\d+?)*?)?)$">
+    <description>Slackware Linux</description>
+    <example os.version="14.1">Slackware Linux 14.1</example>
+    <param pos="0" name="os.vendor" value="Slackware"/>
+    <param pos="0" name="os.family" value="Linux"/>
+    <param pos="0" name="os.product" value="Linux"/>
+    <param pos="1" name="os.version"/>
+  </fingerprint>
+  <fingerprint pattern="^(?i:(?:Oracle|Sun)?\s?(?:Solaris|SunOS)\s?(\d+?(?:\.\d+?)*?)?)$">
```
pmiseiko-r7: Not Linux. :)

xml/operating_system_linux.xml
```diff
@@ -0,0 +1,252 @@
...
+    <description>Oracle Enterprise Linux</description>
+    <example os.version="5.11">Oracle Enterprise Linux 5.11</example>
+    <example os.version="6.7">Oracle Linux 6.7</example>
+    <param pos="0" name="os.vendor" value="Oracle"/>
+    <param pos="0" name="os.family" value="Enterprise Linux"/>
+    <param pos="0" name="os.product" value="Linux"/>
+    <param pos="1" name="os.version"/>
+  </fingerprint>
+  <fingerprint pattern="^(?i:(?:Oracle|Sun)?\s?OpenSolaris\s?(\d+?(?:\.\d+?)*?)?)$">
+    <description>OpenSolaris</description>
```
pmiseiko-r7: Not Linux :)

xml/operating_system_linux.xml
gschneider-r7: Yeah it's starting to get silly actually. Should be an `operating_system_unix.xml`? If so, then do the Linux OSes fall into that as derivatives? 😄 

I had a similar thought on BSD and Darwin...

xml/operating_system_windows.xml
gschneider-r7: Good point. For my use case I was not expecting to be too strict on how valid the Windows version is that would be provided. I also don't (really) care about certainty in this situation, but I can add it here for any other consumers of these matchers that would care. 👍

xml/operating_system_linux.xml
pmiseiko-r7: It's a argument for having a single file but yes I would either break it into more files or put it all in one.

pmiseiko-r7:
I agree with keeping it in a single file.  Why complicate the process.  Recog was designed for a single file to meet a particular purpose.  By breaking it into multiple files you make it harder to track defects as well i.e. when something produces multiple undesired matches across operating systems.  For example, the examples fields should not work for peers (though I don't think we have a test for that).

jhart-r7:
In general I like the idea of putting these in separate files.  That is at least part of what the top-level `matches` attribute was added for.  

The problem then becomes how both recog's client/API and users of recog's data handle the loading of multiple files; obviously issues like ordering, first/last/all match, etc areas of possible concern.  The issue is that recog's API/client can behave one way with regards to load ordering, matching preference, etc, but we can't control how those who just use recog's data act.  

I don't have a good solution for this right now, so I'd say just put it all in one file for now.

xml/operating_system_linux.xml
```diff
@@ -0,0 +1,252 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Patterns for common names of Linux operating systems.
```
rkirk-r7: should this be renamed to be operating_system_unix ? as it covers a lot more than just linux

mhuffman-r7:
I actually implemented the Java recog client to be able to handle either looking up the matchers by file name and/or by the "matches" element. I basically just "union" the content together, so by definition we should not have overlapping regexs, but its a first-one wins. I am fine with either approach.

xml/operating_system_darwin.xml
```diff
@@ -0,0 +1,134 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Patterns for common names of Mac OS X operating systems.
+-->
+<fingerprints matches="operating_system.name">
+  <!-- Match Mac OS Classic first due to weak matching on Mac OS X -->
+  <fingerprint pattern="^(?i:(?:Apple )?Mac OS ([7-9](?:\.\d+?)*?))$">
```
mhuffman-r7: recommend making "Mac" optional in all these regexs. I think `Apple OS 10.7` and `Mac OS 10.7`should be a valid match. Basically, I think it should at minimum have either Apple or Mac, but not need both.

gwiseman-r7: This is classic (pre-OS X) Mac OS. I've never heard anything referred to as Apple OS (there was an Apple DOS and of course iOS, but that's it). If anything, we should match on "System 7" -- it wasn't actually called Mac OS until version 7.6. Oh the good old days.

mhuffman-r7: Yeah, I picked that first regex, but I really meant for OS X in general. so e.g. `Mac OS 10.11` and `Apple OS 10.11`

gwiseman-r7: Apple OS 10.x still looks pretty weird to me. But maybe there are such banners out there...

gwiseman-r7: But actually yeah if we're making the X optional after Mac OS, I guess I do see where you're coming from.

mhuffman-r7: The primary reason I came up with this is that I was testing our API usage of recog, and I expected this to be parsed. Perhaps its not something that would generally be "collected" of an asset, but certainly I can see it being input by a human (if don't manually). Perhaps I am crazy ;)

gwiseman-r7: Right I forgot about the use case here.. if you thought of this, there surely are crazier people out there who would input this ;)

gschneider-r7:
I added a 0.5 certainty for the catch-all matchers and merged everything into a single file (though as @mhuffman-r7 noted the Java code has been updated to handle multiple files with a shared "matches" value), ordered as follows - mostly based on how common I expect to see them used for what we're working on:

* Windows (catch-all at the end of the "block")
* Linux (alphabetically sorted, with catch-all at the end of the "block")
* Mac OS X/iOS
* Other Unixes
* Network gear (this might actually be more common than other Unixes, but oh well)

jhart-r7:
This looks good to me.  Thanks for adding this and addressing some technical debt while here.  👍

pmiseiko-r7:
Wait.  The matchers now work across files and stop on the first match?  That -only- impacts this particular Recog content and none of the existing fingerprinting content, correct?

gschneider-r7:
@pmiseiko-r7 the changes were to add support for it, sorry for not being clear. It does not change the current behavior.

pmiseiko-r7:
Ok.  So long as legacy behavior is preserved since such a change would likely cause really weird issues.

:ship: it.

PR #112, opened 2016-10-26
Add missing OpenSSH and HipServ SSH fingerprints

pdeardorff-r7:
This is just a start, will push more missing SSH banners as I go.  Current questions: 

- Rasbian is for the Raspberry Pi but as-is this isn't mentioned in the fingerprint.  Where should it go?  And should we add a new "IoT" category or family? 
- HipServ is a home NAS to store media is there a better family for that?  Otherwise I've called it a Router...

### DIFF
--- a/xml/ssh_banners.xml
+++ b/xml/ssh_banners.xml
@@ -479,9 +479,9 @@ fingerprint SSH servers.
     <param pos="0" name="service.family" value="OpenSSH"/>
     <param pos="0" name="service.product" value="OpenSSH"/>
     <param pos="0" name="os.vendor" value="HipServ"/>
-    <param pos="0" name="os.device" value="Router"/>
-    <param pos="0" name="os.family" value="RouterOS"/>
-    <param pos="0" name="os.product" value="RouterOS"/>
+    <param pos="0" name="os.device" value="NAS"/>
+    <param pos="0" name="os.family" value="Linux"/>
+    <param pos="0" name="os.product" value="HipServ"/>
   </fingerprint>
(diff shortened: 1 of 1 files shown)


xml/ssh_banners.xml
```diff
@@ -439,6 +471,18 @@ fingerprint SSH servers.
...
+    <description>OpenSSH on HipServ</description>
+    <example service.version="4.3">OpenSSH_4.3-HipServ</example>
+    <param pos="1" name="service.version"/>
+    <param pos="0" name="service.vendor" value="OpenBSD"/>
+    <param pos="0" name="service.family" value="OpenSSH"/>
+    <param pos="0" name="service.product" value="OpenSSH"/>
+    <param pos="0" name="os.vendor" value="HipServ"/>
+    <param pos="0" name="os.device" value="Router"/>
+    <param pos="0" name="os.family" value="RouterOS"/>
+    <param pos="0" name="os.product" value="RouterOS"/>
```
jhart-r7: The product, family and maybe device for `os` seem like copy-pasta from the previous fingerprint.

pdeardorff-r7: Yep, it is.  In the description I put that I wasn't sure about HipServe because it's a NAS but I couldn't find a suitable device for that so I called it a Router for now.  What would you recommend for that?

gschneider-r7: `os.family` and `os.product` should not be RouterOS, though, right? Those belong to MikroTik not HipServ?

pdeardorff-r7: Makes sense, thanks.  Correct definitely shouldn't be `RouterOS` I initially though that was a generic term but now see it actually belongs to MikroTik

PR #113, opened 2016-11-08
Add missing ssh banners

pdeardorff-r7:
Second pass at adding more SSH banners

### DIFF
--- a/xml/ssh_banners.xml
+++ b/xml/ssh_banners.xml
@@ -392,7 +392,7 @@ fingerprint SSH servers.
     <param pos="0" name="os.product" value="Linux"/>
     <param pos="0" name="os.version" value="8.0"/>
   </fingerprint>
-  <fingerprint pattern="^OpenSSH_([^\s]+)\s+(Debian.+[-6|squeeze].*)$">
+  <fingerprint pattern="^OpenSSH_([^\s]+)\s+(Debian-6(?:squeeze)?.*)$">
     <description>OpenSSH running on Debian 6.0 (squeeze)</description>
     <example service.version="5.5p1" openssh.comment="Debian-6+squeeze4">OpenSSH_5.5p1 Debian-6+squeeze4</example>
     <example service.version="5.5p1" openssh.comment="Debian-6">OpenSSH_5.5p1 Debian-6</example>
--- a/xml/ssh_banners.xml
+++ b/xml/ssh_banners.xml
@@ -313,7 +313,6 @@ fingerprint SSH servers.
     <param pos="0" name="service.vendor" value="OpenBSD"/>
     <param pos="0" name="service.family" value="OpenSSH"/>
     <param pos="0" name="service.product" value="OpenSSH"/>
-    <param pos="0" name="os.device" value="General"/>
   </fingerprint>
   <fingerprint pattern="^OpenSSH_(6\.6(?:\.\d)?p1) (Ubuntu-2ubuntu\d+(?:\.\d+)?)$">
     <description>OpenSSH running on Ubuntu 14.04</description>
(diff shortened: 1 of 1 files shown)


xml/ssh_banners.xml
```diff
@@ -304,6 +304,17 @@ fingerprint SSH servers.
     <param pos="0" name="os.product" value="Linux"/>
     <param pos="0" name="os.version" value="13.04"/>
   </fingerprint>
+  <fingerprint pattern="^OpenSSH_(\d+\.\d+(?:\.\d+)?(?:p\d+)?)[_|-](hpn13v11)$">
+    <description>OpenSSH</description>
```
tsellers-r7: I suggest changing the label 'OpenSSH with HPN patches' or similar.

xml/ssh_banners.xml
```diff
@@ -381,9 +392,10 @@ fingerprint SSH servers.
     <param pos="0" name="os.product" value="Linux"/>
     <param pos="0" name="os.version" value="8.0"/>
   </fingerprint>
-  <fingerprint pattern="^OpenSSH_([^\s]+)\s+(Debian.+squeeze.*)$">
+  <fingerprint pattern="^OpenSSH_([^\s]+)\s+(Debian.+[-6|squeeze].*)$">
```
tsellers-r7: Do you have an example of this banner w/o `Debian-6` in it?  It would seem that the `-6` shouldn't be optional.  Also, if you test this on Rubular.com you will find that the regex matches `OpenSSH_4.1p1 Debian-7ubuntu4`.

I recommend something similar to 
`^OpenSSH_([^\s]+)\s+(Debian-6(?:squeeze)?.*)$`

pdeardorff-r7: A yes the `.+` on the left side is too greedy. I like your recommendation, I think that's the best way to make the `squeeze` optional.

xml/ssh_banners.xml
```diff
@@ -304,6 +304,17 @@ fingerprint SSH servers.
...
+  <fingerprint pattern="^OpenSSH_(\d+\.\d+(?:\.\d+)?(?:p\d+)?)[_|-](hpn13v11)$">
+    <description>OpenSSH with HPN patches</description>
+    <example service.version="6.1" openssh.comment="hpn13v11">OpenSSH_6.1_hpn13v11</example>
+    <example service.version="5.8p1" openssh.comment="hpn13v11">OpenSSH_5.8p1-hpn13v11</example>
+    <param pos="1" name="service.version"/>
+    <param pos="2" name="openssh.comment"/>
+    <param pos="0" name="service.vendor" value="OpenBSD"/>
+    <param pos="0" name="service.family" value="OpenSSH"/>
+    <param pos="0" name="service.product" value="OpenSSH"/>
+    <param pos="0" name="os.device" value="General"/>
```
jhart-r7: I'd suggest removing this.  I don't think you can discern anything about the OS or the "device" from this fingerprint.

xml/ssh_banners.xml
```diff
@@ -304,6 +304,16 @@ fingerprint SSH servers.
     <param pos="0" name="os.product" value="Linux"/>
     <param pos="0" name="os.version" value="13.04"/>
   </fingerprint>
+  <fingerprint pattern="^OpenSSH_(\d+\.\d+(?:\.\d+)?(?:p\d+)?)[_|-](hpn13v11)$">
```
jhart-r7: During my final review of this, I noticed that the `hpn` part here is hardcoded.  It looks like this can absolutely be something more than just hpn13v11, for example I see what looks to be {{hpn14v5}} and {{hpn14v9}} at https://launchpad.net/~yoda-jazz-kc/+archive/ubuntu/hpn-ssh.  I'd suggest just changing the relevant part of this regex to `(hpn\d+v\d+)`

PR #114, opened 2016-11-10
[WIP] Add support for matching against all XML databases

tsellers-r7:
The goal of this PR is to enable the use of Recog to search **all** fingerprint databases for a match for a given string.  To do this effectively Recog must support filtering of the results in order to tailor them to the caller's needs.   It must allow prioritization of the databases to ensure that those containing the highest quality matches are processed first.  These changes **must** not impact the parameters of existing functions or materially change the results that they currently provide.

The changes here were written with an eye towards existing use cases.  They **should** be able to drop in w/o breakage for any environment where the XML XSD has not been modified or isn't being overwritten by a build process.

###  Code Changes

* Add `Nizer#match_all_db` to support matching a string against multiple databases.
  * Allow filtering of the above searches by providing a Hash of filter options
    * `:match_key` - Same as the `match_key` used in the `Nizer#match` search. This
                      value is set at the top level `fingerprints` element in an XML fingerprints DB file. If 
                      this value isn't set then it will be the name of the XML file without the extension.
                      This value in the XML file and it's related auto-population were in the existing code.

    * `:datbase_type` - Allows filtering by fingerprint type such as `service`, `util.os`, etc with the goal of
                      limiting undesirable matches.  An example of this would be matching against an 
                      `util.os` fingerprint database just because it has `Windows 2008` in the name. This
                      is a **NEW** value set at the top level `fingerprints` element in an XML fingerprints
                      DB file.

    * `:protocol` - Allows filtering by protocol such as `smtp`, `ftp`, etc.  This is a **NEW**
                      value set at the top level `fingerprints` element in an XML fingerprints DB
                      file **but** can be over written by specifying a `service.protocol` element in an 
                      individual fingerprint match definition.

    * `:multi_match` - Allows `Nizer#match_all_db` to return multiple matches.  This allows 
                      `Nizer#match_all_db` to replace the functionality of `Nizer#multi_match`.

    Example use of `Nizer#match_all_db` with filters:

```Ruby
    result_filter = { database_type: 'service', protocol: ‘smtp’  }
    fp_match = Recog::Nizer.match_all_db(banner_to_match, result_filter)
```

* Add `Nizer#load_db` which allows using `Nizer` to load a specific XML fingerprint file or directory of files.  Previously it would, via DBManager, only load the databases in the `xml` directory under the `Recog` directory in use.  It will automatically sort the databases when they are loaded by `Nizer#load_db`.

* Add `Nizer#unload_db` which destroys the `Nizer::DBManager` class level object in the variable.

* Add `Nizer#display_db_order` which outputs the currently loaded fingerprints in the order in which they will be queried. It also displays their priority and database type.  This is useful for debugging issues related to matching across multiple databases. An example of the output has been provided at the end of this description.

* Modify `DBManager#load_databases` so as to permit loading an individual file.  Previously it would attempt to load all XML files in the specified directory. Now it will detect if the specified path is a file or directory and handle as appropriate.

* `Fingerprint#initialize` now supports passing in **optional** parameters `match_key` and `protocol` so the fingerprint definitions will contain this information.  By doing it this way it provides this information to any applications that are using the fingerprinting but not doing so via `Recog::Nizer`.  `recog_match` is an example of this.

**NOTE**:

* `Nizer#match` and `Nizer#multi_match` have been rewritten as wrappers around 
  `Nizer#match_all_db`. The former calling syntax is still in place and the parameters are
  mapped to filters.  The wrappers still pass all of the original RSpec tests as well as some
  new tests.

* `rspec` tests have been written to handle new code and some tests for existing code were expanded.

### XML Changes

* Added support for specifying a `preference` at the top level `fingerprints` of a database. This 
   is used by `Nizer#match_all_db`  to determine the order in which databases will be used to
   process a banner string. Preference is a float value between 0 and 1.0 with the highest number
   being given priority (processed first). If an XML database doesn't have a priority value set then
   it will be set to the value of `DEFAULT_FP_PREFERENCE` as  defined in `Recog::DB`. This value
  is currently 0.10.  I've set the values on certain current databases based on the likelihood that 
   the db will provide a quality match. If databases were thought to have similar quality then in 
  some cases priority was given to a database that covered a more common service. The highest
  `preference` value currently set is 0.90 and the lowest is 0.10.  The intent is to allow users to
   always have room to create/modify databases that will have priority over existing databases by
   setting #'s above 0.90 and to set catch-all/low preference databases by setting #'s
   below 0.10.  The XML schema in `xml/fingerprints.xsd` limits `preference` to values between
   0.0 and 1.0 inclusive.  RSpec tests limit the `preference` values for XML databases that
   are included in this project to values between 0.10 and 0.90 inclusive.

  **Obviously these values will be the subject of discussion and will be tuned over time.**

* Add support for specifying a `protocol` at the top level `fingerprints` of a database and at the individual fingerprint level. This value indicates the application protocol of the fingerprints found within the XML fingerprint database. Examples of this would be 'ftp', 'smtp', 'ssh', etc. Within Recog this value will be superseded by a `service.protocol` attribute on a specific  fingerprint match. See `Recog::DB#parse_fingerprints`.  This value has two purposes. It can be used for filtering  ( See Recog::Nizer#match_all_db and #multi_match_all_db ) and is also returned as part of any successful match.
  * Top level XML example:
    ```xml
    <fingerprints matches="ldap.search_result" protocol="ldap">
    ```
  * Within an individual fingerprint
    ```xml
    <param pos="0" name="service.protocol" value="ldap"/>
    ```
* Added support for specifying a `database_type` at the top level `fingerprints` of a database. This is used to indicates the type of fingerprints matches expected to be found within the database. These values are used by `Recog::Nizer#match_all_db` to filter matches to just the type of database that is relevant to the match string.

  This value is **NOT** returned as part of successful matches.

   Current values are:

    * **service**: These fingerprints are intended to match banners or other responses from services. Fingerprint matches in 'service' database do not necessarily have to return 'service.' attributes in the match data.

    * **util.os**: These fingerprints are intended to be used to identify or extract OS related information from strings that are not responses to service probes. This may be used in a utility capacity and may provide for data enrichment via an independent all after a service banner match has already be made.

* Updated `xml/fingerprints.xsd` to support the new `fingerprints` attributes for protocol, type, and priority.

### Results output changes

All output, even that which doesn't call new methods, will contain values for `fingerprint_db` and `service.protocol` when a match is made and if the data is populated in the source XML.  The logic in `Recog::DB` will ensure that `fingerprint_db` is always populated.

```bash
echo 'OpenSSH_6.6p1 Ubuntu-2ubuntu1' | bin/recog_match xml/ssh_banners.xml -

MATCH: {
  "matched"=>"OpenSSH running on Ubuntu 14.04", 
  "service.version"=>"6.6p1", 
  "openssh.comment"=>"Ubuntu-2ubuntu1", 
  "service.vendor"=>"OpenBSD", 
  "service.family"=>"OpenSSH", 
  "service.product"=>"OpenSSH", 
  "os.vendor"=>"Ubuntu", 
  "os.device"=>"General", "os.family"=>"Linux", 
  "os.product"=>"Linux", "os.version"=>"14.04", 
  "service.protocol"=>"ssh", 
  "fingerprint_db"=>"ssh.banner", 
  "data"=>"OpenSSH_6.6p1 Ubuntu-2ubuntu1"}
```

Current fingerprint database priority, highest priority first:
```
Preference  Database                Type 
     0.900  http_header.server      service 
     0.900  ftp.banner              service 
     0.900  imap4.banner            service 
     0.900  pop3.banner             service 
     0.900  ssh.banner              service 
     0.800  h323_callresp           service 
     0.800  mysql.error             service 
     0.800  ntp.readvar             service 
     0.800  operating_system.name   util.os 
     0.800  ldap.search_result      service 
     0.750  mysql.banners           service 
     0.200  http_header.wwwauth     service 
     0.200  smtp.banner             service 
     0.190  smtp_ehlo               service 
     0.180  smtp_help               service 
     0.170  smtp_noop               service 
     0.160  smtp_expn               service 
     0.150  smtp_vrfy               service 
     0.140  smtp_debug              service 
     0.130  smtp_turn               service 
     0.120  smtp_rset               service 
     0.110  smtp_quit               service 
     0.100  ssdp_header.server      service 
     0.100  architecture            util.os 
     0.100  hp_pjl_id               service 
     0.100  http_header.cookie      service 
     0.100  mdns.device-info.txt    util.os 
     0.100  mdns.workstation.txt    service 
     0.100  nntp.banner             service 
     0.100  rsh_resp                service 
     0.100  sip_header.server       service 
     0.100  sip_header.user_agent   service 
     0.100  smb.native_lm           service 
     0.100  smb.native_os           util.os 
     0.100  smtp_mailfrom           service 
     0.100  smtp_rcptto             service 
     0.100  snmp.sys_description    service 
     0.100  snmp.sys_object_id      service 
     0.100  apache_os               util.os 
```

### DIFF
--- a/features/data/matching_banners_fingerprints.xml
+++ b/features/data/matching_banners_fingerprints.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<fingerprints protocol="ftp" type="service">
+<fingerprints protocol="ftp" database_type="service">
   <fingerprint pattern="^-{10} Welcome to Pure-FTPd (.*)-{10}$">
     <example>---------- Welcome to Pure-FTPd ----------</example>
     <description>Pure-FTPd
--- a/features/data/matching_banners_fingerprints.xml
+++ b/features/data/matching_banners_fingerprints.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<fingerprints>
+<fingerprints protocol="ftp" type="service">
   <fingerprint pattern="^-{10} Welcome to Pure-FTPd (.*)-{10}$">
     <example>---------- Welcome to Pure-FTPd ----------</example>
     <description>Pure-FTPd
--- a/features/data/matching_banners_fingerprints.xml
+++ b/features/data/matching_banners_fingerprints.xml
@@ -8,10 +8,12 @@
     <param pos="1" name="pureftpd.config"/>
     <param pos="0" name="service.family" value="Pure-FTPd"/>
     <param pos="0" name="service.product" value="Pure-FTPd"/>
+    <param pos="0" name="service.protocol" value="ftp"/>
   </fingerprint>
   <fingerprint pattern="^(\S+) FTP Server \(SunOS (\S+)\) ready\.?$" flags="REG_ICASE">
     <description>SunOS/Solaris</description>
     <example>example.com FTP server (SunOS 5.7) ready.</example>
+    <param pos="0" name="service.protocol" value="ftp"/>
     <param pos="0" name="os.vendor" value="Sun"/>
     <param pos="0" name="os.family" value="Solaris"/>
     <param pos="0" name="os.product" value="Solaris"/>
--- a/features/data/matching_banners_fingerprints.xml
+++ b/features/data/matching_banners_fingerprints.xml
@@ -13,7 +13,6 @@
   <fingerprint pattern="^(\S+) FTP Server \(SunOS (\S+)\) ready\.?$" flags="REG_ICASE">
     <description>SunOS/Solaris</description>
     <example>example.com FTP server (SunOS 5.7) ready.</example>
-    <param pos="0" name="service.protocol" value="ftp"/>
     <param pos="0" name="os.vendor" value="Sun"/>
     <param pos="0" name="os.family" value="Solaris"/>
     <param pos="0" name="os.product" value="Solaris"/>
--- a/lib/recog/db.rb
+++ b/lib/recog/db.rb
@@ -13,19 +13,19 @@ class DB
   #   against strings that make sense for the {#match_key}
   attr_reader :fingerprints
 
-  # @return [String] Taken from the `fingerprints/matches` element, or
+  # @return [String] Taken from the `fingerprints/matches` attribute, or
   #   defaults to the basename of {#path} without the `.xml` extension.
   attr_reader :match_key
 
-  # @return [String] Taken from the `fingerprints/protocol` element, or
+  # @return [String] Taken from the `fingerprints/protocol` attribute, or
   #   defaults to an empty string
   attr_reader :protocol
 
-  # @return [String] Taken from the `fingerprints/database_type` element
+  # @return [String] Taken from the `fingerprints/database_type` attribute
   #   defaults to an empty string
   attr_reader :database_type
 
-  # @return [Float] Taken from the `fingerprints/preference` element,
+  # @return [Float] Taken from the `fingerprints/preference` attribute,
   #   defaults to 0.10.  Used when ordering databases, highest numbers
   #   are given priority and are processed first.
   attr_reader :preference
--- a/lib/recog/db.rb
+++ b/lib/recog/db.rb
@@ -25,20 +25,22 @@ class DB
   #   defaults to an empty string
   attr_reader :database_type
 
-  # @return [Integer] Taken from the `fingerprints/priority` element,
-  #   defaults to 100.  Used when ordering databases, lowest numbers
-  #   first.
-  attr_reader :priority
+  # @return [Float] Taken from the `fingerprints/preference` element,
+  #   defaults to 0.10.  Used when ordering databases, highest numbers
+  #   are given priority and are processed first.
+  attr_reader :preference
 
-  # Default Fingerprint database priority when it isn't specified in file
-  DEFAULT_FP_DB_PRIORITY = 100
+  # Default Fingerprint database preference when it isn't specified in file
+  # Do not use a value below 0.10 so as to allow users to specify lower
+  # values in their own custom XML that will always run last.
+  DEFAULT_FP_PREFERENCE = 0.10
 
   # @param path [String]
   def initialize(path)
     @match_key = nil
     @protocol = ''
     @database_type = ''
-    @priority = DEFAULT_FP_DB_PRIORITY
+    @preference = DEFAULT_FP_PREFERENCE.to_f
     @path = path
     @fingerprints = []
 
@@ -60,7 +62,7 @@ class DB
       @match_key = fbase['matches'].to_s if fbase['matches']
       @protocol = fbase['protocol'].to_s if fbase['protocol']
       @database_type = fbase['database_type'].to_s if fbase['database_type']
-      @priority = fbase['priority'].to_i if fbase['priority']
+      @preference = fbase['preference'].to_f if fbase['preference']
 
     end
 
--- a/lib/recog/db.rb
+++ b/lib/recog/db.rb
@@ -21,9 +21,9 @@ class DB
   #   defaults to an empty string
   attr_reader :protocol
 
-  # @return [String] Taken from the `fingerprints/type` element
+  # @return [String] Taken from the `fingerprints/database_type` element
   #   defaults to an empty string
-  attr_reader :db_type
+  attr_reader :database_type
 
   # @return [Integer] Taken from the `fingerprints/priority` element,
   #   defaults to 100.  Used when ordering databases, lowest numbers
@@ -37,7 +37,7 @@ class DB
   def initialize(path)
     @match_key = nil
     @protocol = ''
-    @db_type = ''
+    @database_type = ''
     @priority = DEFAULT_FP_DB_PRIORITY
     @path = path
     @fingerprints = []
@@ -49,35 +49,24 @@ class DB
   def parse_fingerprints
     xml = nil
 
-    File.open(self.path, "rb") do |fd|
+    File.open(self.path, 'rb') do |fd|
       xml = Nokogiri::XML(fd.read(fd.stat.size))
     end
 
     raise "#{self.path} is invalid XML: #{xml.errors.join(',')}" unless xml.errors.empty?
 
-    xml.xpath("/fingerprints").each do |fbase|
-      if fbase['matches']
-        @match_key = fbase['matches'].to_s
-      end
+    xml.xpath('/fingerprints').each do |fbase|
 
-      if fbase['protocol']
-        @protocol = fbase['protocol'].to_s
-      end
+      @match_key = fbase['matches'].to_s if fbase['matches']
+      @protocol  = fbase['protocol'].to_s if fbase['protocol']
+      @database_type = fbase['database_type'].to_s if fbase['database_type']
+      @priority  = fbase['priority'].to_i if fbase['priority']
 
-      if fbase['type']
-        @db_type = fbase['type'].to_s
-      end
-
-      if fbase['priority']
-        @priority = fbase['priority'].to_i
-      end
     end
 
-    unless @match_key
-      @match_key = File.basename(self.path).sub(/\.xml$/, '')
-    end
+    @match_key = File.basename(self.path).sub(/\.xml$/, '') unless @match_key
 
-    xml.xpath("/fingerprints/fingerprint").each do |fprint|
+    xml.xpath('/fingerprints/fingerprint').each do |fprint|
       @fingerprints << Fingerprint.new(fprint, @match_key, @protocol)
     end
 
--- a/lib/recog/db.rb
+++ b/lib/recog/db.rb
@@ -19,12 +19,23 @@ class DB
 
   # @return [String] Taken from the `fingerprints/protocol` element, or
   #   defaults to an empty string
-  attr_reader :match_protocol
+  attr_reader :protocol
+
+  # @return [String] Taken from the `fingerprints/type` element
+  #   defaults to an empty string
+  attr_reader :db_type
+
+  # @return [Integer] Taken from the `fingerprints/priority` element,
+  #   defaults to 100.  Used when ordering databases, lowest numbers
+  #   first.
+  attr_reader :priority
 
   # @param path [String]
   def initialize(path)
     @match_key = nil
-    @match_protocol = ''
+    @protocol = ''
+    @db_type = ''
+    @priority = 100
     @path = path
     @fingerprints = []
 
@@ -47,7 +58,15 @@ class DB
       end
 
       if fbase['protocol']
-        @match_protocol = fbase['protocol'].to_s
+        @protocol = fbase['protocol'].to_s
+      end
+
+      if fbase['type']
+        @db_type = fbase['type'].to_s
+      end
+
+      if fbase['priority']
+        @priority = fbase['priority'].to_i
       end
     end
 
@@ -56,7 +75,7 @@ class DB
     end
 
     xml.xpath("/fingerprints/fingerprint").each do |fprint|
-      @fingerprints << Fingerprint.new(fprint, @match_key, @match_protocol)
+      @fingerprints << Fingerprint.new(fprint, @match_key, @protocol)
     end
 
     xml = nil
--- a/lib/recog/db.rb
+++ b/lib/recog/db.rb
@@ -58,9 +58,9 @@ class DB
     xml.xpath('/fingerprints').each do |fbase|
 
       @match_key = fbase['matches'].to_s if fbase['matches']
-      @protocol  = fbase['protocol'].to_s if fbase['protocol']
+      @protocol = fbase['protocol'].to_s if fbase['protocol']
       @database_type = fbase['database_type'].to_s if fbase['database_type']
-      @priority  = fbase['priority'].to_i if fbase['priority']
+      @priority = fbase['priority'].to_i if fbase['priority']
 
     end
 
--- a/lib/recog/db.rb
+++ b/lib/recog/db.rb
@@ -30,12 +30,15 @@ class DB
   #   first.
   attr_reader :priority
 
+  # Default Fingerprint database priority when it isn't specified in file
+  DEFAULT_FP_DB_PRIORITY = 100
+
   # @param path [String]
   def initialize(path)
     @match_key = nil
     @protocol = ''
     @db_type = ''
-    @priority = 100
+    @priority = DEFAULT_FP_DB_PRIORITY
     @path = path
     @fingerprints = []
 
--- a/lib/recog/db.rb
+++ b/lib/recog/db.rb
@@ -17,9 +17,14 @@ class DB
   #   defaults to the basename of {#path} without the `.xml` extension.
   attr_reader :match_key
 
+  # @return [String] Taken from the `fingerprints/protocol` element, or
+  #   defaults to an empty string
+  attr_reader :match_protocol
+
   # @param path [String]
   def initialize(path)
     @match_key = nil
+    @match_protocol = ''
     @path = path
     @fingerprints = []
 
@@ -40,6 +45,10 @@ class DB
       if fbase['matches']
         @match_key = fbase['matches'].to_s
       end
+
+      if fbase['protocol']
+        @match_protocol = fbase['protocol'].to_s
+      end
     end
 
     unless @match_key
@@ -47,7 +56,7 @@ class DB
     end
 
     xml.xpath("/fingerprints/fingerprint").each do |fprint|
-      @fingerprints << Fingerprint.new(fprint)
+      @fingerprints << Fingerprint.new(fprint, @match_key, @match_protocol)
     end
 
     xml = nil
--- a/lib/recog/db_manager.rb
+++ b/lib/recog/db_manager.rb
@@ -13,8 +13,12 @@ class DBManager
   end
 
   def load_databases
-    Dir[self.path + "/*.xml"].each do |dbxml|
-      self.databases << DB.new(dbxml)
+    if File.directory?(self.path)
+      Dir[self.path + "/*.xml"].each do |dbxml|
+        self.databases << DB.new(dbxml)
+      end
+    elsif File.file?(self.path) && File.readable?(self.path)
+       self.databases << DB.new(self.path)
     end
   end
 
--- a/lib/recog/db_manager.rb
+++ b/lib/recog/db_manager.rb
@@ -17,8 +17,8 @@ class DBManager
       Dir[self.path + "/*.xml"].each do |dbxml|
         self.databases << DB.new(dbxml)
       end
-    elsif File.file?(self.path) && File.readable?(self.path)
-       self.databases << DB.new(self.path)
+    else
+      self.databases << DB.new(self.path)
     end
   end
 
--- a/lib/recog/fingerprint.rb
+++ b/lib/recog/fingerprint.rb
@@ -43,7 +43,17 @@ class Fingerprint
   # @return [Hash,nil] Keys will be host, service, and os attributes
   def match(match_string)
     # match_string.force_encoding('BINARY') if match_string
-    match_data = @regex.match(match_string)
+    begin
+      match_data = @regex.match(match_string)
+    rescue Encoding::CompatibilityError
+      ## FIXFIX - Determine if this can be addressed
+      return nil
+    rescue Exception => e
+      STDERR.puts e.inspect
+      STDERR.puts e
+      STDERR.puts "Problematic banner:\n#{match_string}"
+      STDERR.puts match_string.length
+    end
     return if match_data.nil?
 
     result = { 'matched' => @name }
--- a/lib/recog/fingerprint.rb
+++ b/lib/recog/fingerprint.rb
@@ -27,7 +27,10 @@ class Fingerprint
   attr_reader :tests
 
   # @param xml [Nokogiri::XML::Element]
-  def initialize(xml)
+  # @param match_protocol [String] Protocol such as ftp, mssql, http, etc.
+  def initialize(xml, match_key, match_protocol)
+    @match_key = match_key
+    @match_protocol = match_protocol
     @name   = parse_description(xml)
     @regex  = create_regexp(xml)
     @params = {}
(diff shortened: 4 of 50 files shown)


lib/recog/nizer.rb
```diff
@@ -60,6 +132,48 @@ def self.multi_match(match_key, match_string)
...
 
   #
+  # Search all fingerprint dbs and attempt to find **all** matching
+  # {Fingerprint fingerprint}s. Returns an array of all matching
+  # fingerprints or an empty array.
+  #
+  # @param match_string [String] Service banner to match
+  # @param [Hash] filters This hash contains filters used to limit the
+  #   results to just those from specific types of fingerprints.
+  #   The matched values are derived from the 'fingerprints' top level
```
jhart-r7: Minor, but slightly confusing here.  Maybe something like "The value to match against are taken from the 'fingerprints' ..."

lib/recog/nizer.rb
```diff
@@ -60,6 +132,48 @@ def self.multi_match(match_key, match_string)
...
+  #   element in the fingerprint DB XML or, in the case of 'protocol',
+  #   this value can be overridden at the individual fingerprint level.
+  # @option filters [String] :match_key Value from XML 'matches' or file name
+  # @option filters [String] :fp_type  Fingerprint type (service, os, etc.)
+  # @option filters [String] :protocol Protocol (ftp, smtp, etc.)
```
jhart-r7: For both type and protocol, maybe also mention the exact attribute name that this corresponds to, similar to what you did for `match_key` just above.

lib/recog/nizer.rb
```diff
@@ -60,6 +132,48 @@ def self.multi_match(match_key, match_string)
...
+  # @return [Array] Array of Fingerprint#match
```
jhart-r7: Minor, but mention the no-match edge-case which will result in this returning an empty array.  I prefer this method, FWIW, but it is common for some code to, say, return `nil` when there is no match.  Better to be explicit than leave it up in the air.

lib/recog/nizer.rb
```diff
@@ -41,6 +74,45 @@ def self.match(match_key, match_string)
...
+  # @param [Hash] filters This hash contains filters used to limit the
+  #   results to just those from specific types of fingerprints.
+  #   The matched values are derived from the 'fingerprints' top level
+  #   element in the fingerprint DB XML or, in the case of 'protocol',
+  #   this value can be overridden at the individual fingerprint level.
+  # @option filters [String] :match_key Value from XML 'matches' or file name
+  # @option filters [String] :fp_type  Fingerprint type (service, os, etc.)
+  # @option filters [String] :protocol Protocol (ftp, smtp, etc.)
+  # @return (see Fingerprint#match)
+  def self.match_all_db(match_string, filters = {})
```
jhart-r7: This and the next method are nearly identical.  Is the only difference that one returns a single match or `nil` and the other returns an array of matches, possibly empty?  If so, just have this method call the other one and return `nil` if empty, the first otherwise.

lib/recog/nizer.rb
```diff
@@ -20,6 +20,39 @@ class Nizer
...
+    @@db_sorted = true
+  end
+
+  #
+  # Display the fingerprint databases in the order in which they will be used
+  # to match banners.  This is useful for fingerprint tuning and debugging.
+  def self.display_db_order()
+    unless @@db_manager
+      @@db_manager = Recog::DBManager.new
+      @@db_manager.databases.sort! { |a,b| a.priority <=> b.priority }
```
jhart-r7: Do you need a similar check of `@@db_sorted` here too?

xml/fingerprints.xsd
```diff
@@ -7,6 +7,9 @@
       <xsd:element name="fingerprint" type="fingerprint_element" minOccurs="1" maxOccurs="unbounded"/>
     </xsd:sequence>
     <xsd:attribute name="matches" type="xsd:string" use="optional"/>
+    <xsd:attribute name="protocol" type="xsd:string" use="optional"/>
```
jhart-r7: It would be good to add some comments about these fields so that future readers/users of this schema don't need to read the code to understand these values and their uses.  Either use XML comments or the xsd `annotation`, which I think is more correct here.

xml/fingerprints.xsd
```diff
@@ -7,6 +7,9 @@
...
+    <xsd:attribute name="type" type="xsd:string" use="optional"/>
+    <xsd:attribute name="priority" type="xsd:string" use="optional"/>
```
jhart-r7: May need to talk this one through a bit.  The values seem a little arbitrary and maybe only specific to your current use case.

lib/recog/fingerprint.rb
```diff
@@ -43,7 +47,17 @@ def initialize(xml)
   # @return [Hash,nil] Keys will be host, service, and os attributes
   def match(match_string)
     # match_string.force_encoding('BINARY') if match_string
-    match_data = @regex.match(match_string)
+    begin
+      match_data = @regex.match(match_string)
+    rescue Encoding::CompatibilityError => e
+      # 'ASCII-8BIT' encoded responses will be caught and fixed here.
+      match_data = @regex.match(match_string.force_encoding('UTF-8'))
```
jhart-r7: FWIW, I'm hitting exceptions here.  I'm not sure we are guaranteed to get perfectly UFT-8-able data, and this will blow up when it doesn't.

tsellers-r7: That's perfect! (no sarcasm).  Toss me info on your dataset when you can so I can test some more.

lib/recog/nizer.rb
tsellers-r7: My thoughts were that display_db_order should only called infrequently and during debugging so sorting an already sorted list shouldn't cause any meaningful performance impact. There are other cases were attempts to sort the database list would be made every time match_all_db or multi_match_all_db were called. In those cases, despite the shortness of the array I tried to avoid sorting them.

lib/recog/nizer.rb
tsellers-r7: I like it... I've also been considering turning match and multi_match ( the existing methods ) into wrappers around the new code that just turn the`match_string` into the correct filter.

lib/recog/fingerprint.rb
tsellers-r7: Handled, but my debugging code is still in place for the moment.  Pretty much everything I tested with today was ASCII-8BIT but the only time this was a problem was when (I guess?) the regex was trying to automatically change it to UTF-8.   I would just encode everything as UTF-8 but I'm not sure about the performance impacts yet.

features/data/matching_banners_fingerprints.xml
```diff
@@ -8,10 +8,12 @@
     <param pos="1" name="pureftpd.config"/>
     <param pos="0" name="service.family" value="Pure-FTPd"/>
     <param pos="0" name="service.product" value="Pure-FTPd"/>
+    <param pos="0" name="service.protocol" value="ftp"/>
   </fingerprint>
   <fingerprint pattern="^(\S+) FTP Server \(SunOS (\S+)\) ready\.?$" flags="REG_ICASE">
     <description>SunOS/Solaris</description>
     <example>example.com FTP server (SunOS 5.7) ready.</example>
+    <param pos="0" name="service.protocol" value="ftp"/>
```
tsellers-r7: Done

jhart-r7: The two individual fingerprints here still use `service.protocol`, overriding the same value now present in the `protocol` element in the top level of the document.  To put this another way, I was suggesting mixing the various ways you can specify the protocol to ensure that all ways result in the output fingerprints having the correct protocol.

lib/recog/fingerprint.rb
```diff
@@ -43,7 +47,33 @@ def initialize(xml)
...
     # match_string.force_encoding('BINARY') if match_string
-    match_data = @regex.match(match_string)
+    begin
+      match_data = @regex.match(match_string)
+    rescue Encoding::CompatibilityError => e
+      begin
+        # Replace invalid UTF-8 characters with spaces, just as DAP does.
+        encoded_str = match_string.encode("UTF-8", :invalid => :replace, :undef => :replace, :replace => '')
+        match_data = @regex.match(encoded_str)
+      rescue Exception => e
```
jhart-r7: Careful here.  Both of calls above could throw an exception but you are only writing to`STDERR` about `match_string`, which is only relevant if line 55 throws.

tsellers-r7: Thanks.  I need to clean up failure handling here but I don't want to loose data about failures to parse.. This is on my to do list.

tsellers-r7: Actually, `match_string` is relevant to both since it is the original data that was being processed. Either we generated an exception on line 51 during the original parse, 55 during encoding, or on line 56 while running regex against the newly encoded `match_string`.  While  the error message is a bit verbose, with its output we should be able to reconstruct any issue without having the source data.

I've cleaned it up a touch and added a bit more context.  The output now looks like:

```
Exception while running regex against match_string
#<Encoding::CompatibilityError: incompatible encoding regexp match (UTF-8 regexp with ASCII-8BIT string)>
Length:   14
Encoding: ASCII-8BIT
Problematic data:
/R?{??\ͥ
Raw bytes:
"\x18/\x03R\eH\x8D{\xD6\x1C\xC8\\\xCD\xA5"
```

lib/recog/db.rb
```diff
@@ -17,9 +17,30 @@ class DB
   #   defaults to the basename of {#path} without the `.xml` extension.
   attr_reader :match_key
 
+  # @return [String] Taken from the `fingerprints/protocol` element, or
```
jhart-r7: attribute, not element.

lib/recog/db.rb
```diff
@@ -17,9 +17,30 @@ class DB
...
+  #   defaults to an empty string
+  attr_reader :protocol
+
+  # @return [String] Taken from the `fingerprints/database_type` element
```
jhart-r7: isn't this the type attribute, not element?

lib/recog/db.rb
```diff
@@ -13,13 +13,34 @@ class DB
...
-  # @return [String] Taken from the `fingerprints/matches` element, or
+  # @return [String] Taken from the `fingerprints/matches` attribute, or
   #   defaults to the basename of {#path} without the `.xml` extension.
   attr_reader :match_key
 
+  # @return [String] Taken from the `fingerprints/protocol` attribute, or
+  #   defaults to an empty string
+  attr_reader :protocol
+
+  # @return [String] Taken from the `fingerprints/database_type` attribute
```
jhart-r7: The attribute is called `type`, right?

lib/recog/nizer.rb
```diff
@@ -20,43 +20,122 @@ class Nizer
...
+  #   level element in the fingerprint DB XML or, in the case of 'protocol',
+  #   this value can be overridden at the individual fingerprint level by
+  #   setting a value for 'service.protocol'
+  #
+  #   With the exception of 'match_key', the filters below match the
+  #   'fingerprints' attributes with the same name.
+  # @option filters [String] :match_key Value from XML 'matches' or file name
+  # @option filters [String] :database_type fprint db type: service, util.os, etc.
+  # @option filters [String] :protocol Protocol (ftp, smtp, etc.)
+  # @option filters [String] :multi_match Return all matches instead of first
```
jhart-r7: Looks like this is specified as a boolean.

lib/recog/nizer.rb
```diff
@@ -20,43 +20,122 @@ class Nizer
...
+  # @return [Array] Array of Fingerprint#match or empty array
+  def self.match_all_db(match_string, filters = {})
+    match_string = match_string.to_s.unpack('C*').pack('C*')
+    matches = Array.new # array to hold all fingerprint matches
 
+    self.load_db unless @@db_manager
+
+    return matches unless !@@db_manager.databases || !@@db_manager.databases.empty?
```
jhart-r7: I think the first part of this `unless` represents a bug, but if nothing else this is confusing to read.  I think it would be cleaner to ensure that db_manager initializes `databases` to an empty array.  Then you can remove this line entirely. `databases` will never be `nil`, and it is fine to iterate over it even if empty since you still just return the empty `matches`.

lib/recog/db.rb
tsellers-r7: No, it was renamed as `database_type` across the board to reduce confusion.

lib/recog/nizer.rb
```diff
@@ -20,43 +20,122 @@ class Nizer
...
+      next if filters[:match_key] && !filters[:match_key].include?(db.match_key)
+      next if filters[:database_type] && !filters[:database_type].include?(db.database_type)
       db.fingerprints.each do |fp|
         m = fp.match(match_string)
-        matches.push(m) if m
+        if m
+          # Filter on protocol after match since each individual fp
+          # can contain its own 'protocol' value that overrides the
+          # one set at the DB level.
+          matches.push(m) unless filters[:protocol] && !filters[:protocol].include?(m['service.protocol'])
```
jhart-r7: Are you intentionally do partial string matching on `:protocol` here, or is the `:protocol` filter meant to be a list?  From what I can see, this is just a string that represents the desired protocol (if any), not a list.  Assuming that is true, I'd suggest either making this an exact match or updating to require a regular expression instead.  Otherwise, users will unknowingly try a filter of `http` and get `https` results (for example).

spec/lib/fingerprint_self_test_spec.rb
```diff
@@ -21,6 +21,13 @@
         expect(db.match_key).not_to be_empty
       end
 
+      it "has valid 'preference' value" do
+          # Reserve values below 0.10 and above 0.90 for users
+          # See xml/fingerprints.xsd
+          expect(db.preference.class).to be ::Float
+          expect(db.preference).to be_between(0.10, 0.90)
```
jhart-r7: Hmm.  I know the values outside of this range are intended for end-user customization, but won't this mean that if they put their XML in the same place as the ones that this tests, that it will fail?  Seems like it would.  Unsure of the best way to fix this.

spec/lib/recog/nizer_spec.rb
```diff
@@ -165,4 +289,23 @@
...
+        subject.unload_db()
+      end
+    end
+
+    context "with invalid file path" do
+      it "raises an error" do
+        expect { subject.load_db('no_such_987_file_path') }.to raise_error(Errno::ENOENT)
+        subject.unload_db()
+      end
+    end
```
jhart-r7: probably also worth testing the case when no path is specified.

lib/recog/nizer.rb
```diff
@@ -20,43 +20,122 @@ class Nizer
...
-      next unless db.match_key == match_key
-      db.fingerprints.each do |fprint|
-        m = fprint.match(match_string)
-        return m if m
-      end
-    end
+    filter = { match_key: match_key, multi_match: false }
+    matches = self.match_all_db(match_string, filter)
+
+    return matches[0] if matches && matches[0]
```
jhart-r7: matches will be empty, not `nil`, according to code and docs.  You replace this and the next line with simply `return matches[0]`.

lib/recog/nizer.rb
```diff
@@ -20,43 +20,122 @@ class Nizer
...
 
+  #
+  # @param match_key [String] Fingerprint DB name, e.g. 'smb.native_os'
+  # @param match_string [String] String to match
+  # @return [Array] Array of Fingerprint#match or empty array
   def self.multi_match(match_key, match_string)
-    match_string = match_string.to_s.unpack("C*").pack("C*")
-    @@db_manager ||= Recog::DBManager.new
+    filter = { match_key: match_key, multi_match: true }
+    matches = self.match_all_db(match_string, filter)
```
jhart-r7: minor, but you can drop the assignment and just let the call to `self.match_all_db` be returned instead.

spec/lib/recog/nizer_spec.rb
```diff
@@ -36,6 +46,84 @@
...
+      data = line.strip
+      context "with smb_native_os:#{data}" do
+        let(:match_all_result) { subject.match_all_db(data, VALID_FILTER) }
+
+        it "returns an array" do
+          expect(match_all_result.class).to eq(::Array)
+        end
+
+        it "returns a successful match" do
+          expect(match_all_result[0]['matched'].to_s).to match(/^[A-Z]/)
```
jhart-r7: Here and several other places you are calling `to_s` on the `matched` value, but not for other values.  Why is this necessary?  These should already be strings and not, say, XML attributes.

xml/fingerprints.xsd
```diff
@@ -3,10 +3,101 @@
...
+            This value is returned as part of any successful match.
+
+        - protocol: indicates the application protocol of the fingerprints
+            found within the XML fingerprint database. Examples of this would
+            be 'ftp', 'smtp', 'ssh', etc. Within Recog this value will be
+            superseded by a 'service.protocol' attribute on a specific
+            fingerprint match. See Recog::DB#parse_fingerprints
+            This value has two purposes. It can be used for filtering
+            ( See Recog::Nizer#match_all_db ) and is returned as part of any
+            successful match.
```
jhart-r7: Might be good to explain why having this returned may be useful to end-users.

xml/fingerprints.xsd
```diff
@@ -3,10 +3,101 @@
...
+
+        - matches: provides a XML fingerprint database identifier. It is
+            intended to be descriptive in nature and often indicates a protocol
+            and function or sub-protocol type. Examples of this would be
+            'mysql.banners' vs 'mysql.errors'. If this attribute is not present
+            in the XML then its value is the base filename without the file
+            extension. See Recog::DB#parse_fingerprints
+            This value is returned as part of any successful match.
+
+        - protocol: indicates the application protocol of the fingerprints
```
jhart-r7: 'application protocol' reads strange not just because I'm not sure that is a thing, but also because in this same description it is referred to as `service.protocol`.  Maybe reword this first sentence to "indicates the name of the service or protocol <something>".

lib/recog/db.rb
tsellers-r7: The only place I spotted `type` is in the cucumber testing XML which isn't covered by spec. I've fixed it there now.

xml/http_wwwauth.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- HTTP WWW-Authenticate headers are matched against these patterns to fingerprint HTTP servers. -->
-<fingerprints matches="http_header.wwwauth">
+<fingerprints matches="http_header.wwwauth" protocol="http" database_type="service" preference="0.20">
```
jhart-r7: I'm asking this both because I don't agree with the preference value and I think it might be useful to understand how you chose this value, since the introduction of this value is a key part of this PR.  The regular expressions used in this file seem to be fairly restrictive -- they have beginning and end anchors and look for very specific strings that are unlikely to match when applied to arbitrary other things.

spec/lib/recog/nizer_spec.rb
tsellers-r7: Copy-pasta from original `match` and `multi_match` tests.  Removed from new code.

lib/recog/nizer.rb
tsellers-r7: FIx.

spec/lib/recog/nizer_spec.rb
```diff
@@ -36,6 +46,84 @@
...
+
+        let(:nomatch_proto_result) { subject.match_all_db(data, NOMATCH_PROTO) }
+        it "returns an array when protocol search doesn't match" do
+          expect(nomatch_proto_result.class).to eq(::Array)
+        end
+        it "returns an empty array when protocol search doesn't match" do
+          expect(nomatch_proto_result).to be_empty
+        end
+
+
```
tsellers-r7: Fixed

lib/recog/nizer.rb
```diff
@@ -20,43 +20,122 @@ class Nizer
...
+    @@db_sorted = false
+  end
+
+  #
+  # Display the fingerprint databases in the order in which they will be used
+  # to match banners.  This is useful for fingerprint tuning and debugging.
+  def self.display_db_order
+    self.load_db unless @@db_manager
+
+    puts format('%s  %-22s  %s ', 'Preference', 'Database', 'Type')
```
jhart-r7: Might be nice to add protocol here too.

lib/recog/nizer.rb
tsellers-r7: Good point.  I've addressed this as well as added an additional test related to this.

spec/lib/recog/nizer_spec.rb
tsellers-r7: I've added an explicit test for calling load_db with an empty path.  Calling with no path 
tested during the testing of `match`, `multi_match`, and `match_all_db` since no path is the default behavior when they load a DB.

tsellers-r7: Went ahead an added an explicit test for calling w/o no parameters because.. stuff happens..

spec/lib/fingerprint_self_test_spec.rb
jhart-r7: Yeah, fair point.  Another side effect of this is that if we *don't* allow things in the `xml` directory with values outside of this range, they won't be able to be tested in the same way that the shipped versions can.  For example, the specs ensure that the examples match and that the ordering is correct.  The utilities in `bin` don't do that.  My be no good solution to this right now. Your call.

lib/recog/nizer.rb
tsellers-r7: Fixed, thanks.

xml/fingerprints.xsd
tsellers-r7: Fixed.

lib/recog/db_manager.rb
```diff
@@ -13,8 +13,12 @@ def initialize(path = DefaultDatabasePath)
   end
 
   def load_databases
-    Dir[self.path + "/*.xml"].each do |dbxml|
-      self.databases << DB.new(dbxml)
+    if File.directory?(self.path)
+      Dir[self.path + "/*.xml"].each do |dbxml|
```
jhart-r7: Hmm.  Thinking about this while looking at the output of your utility that is exercising this code and I think you might need to ensure that the files are always returned in the same order.  From what I've read, `Dir.glob` will not guarantee order.  So, in the event of two XML files with the same preference, there is no guarantee as to the order they will be loaded.  This will allow odd ordering problems that might exist between different operating systems to cause the shipped fingerprints to be evaluated differently.

lib/recog/nizer.rb
tsellers-r7: No, I think this was a hold over from a previous filtering technique that just happened to continue working.  Thanks for catching that,  I have changed the code.

lib/recog/db_manager.rb
tsellers-r7: I've tweaked the current preference values a touch to spread out files where there might be overlap.

PR #128, opened 2017-05-02
Adding banner detection for Intel AMT

cmccrisken-r7:
## Description
Adding banner detection from Intel AMT's HTTP web management console running on `tcp/16992` and `tcp/16993` (HTTPS)

## How Has This Been Tested?
- [x] Run rspec
- [x] Run `bin/recog_verify xml/http_servers.xml`

### DIFF
--- a/xml/http_servers.xml
+++ b/xml/http_servers.xml
@@ -3283,4 +3283,12 @@
     <param pos="0" name="service.product" value="Symantec Endpoint Protection Manager"/>
     <param pos="0" name="service.family" value="Symantec Endpoint Protection Manager"/>
   </fingerprint>
+  <fingerprint pattern="^Intel\(R\) Active Management Technology\s(\d+\.\d+\.\d+\.\d+|\d+\.\d+\.\d+|\d+\.\d+)">
+    <description>Intel(R) Active Management Technology</description>
+    <example>Intel(R) Active Management Technology 7.1.86</example>
+    <param pos="0" name="service.vendor" value="Intel"/>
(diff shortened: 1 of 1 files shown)


xml/http_servers.xml
```diff
@@ -3283,4 +3283,12 @@
     <param pos="0" name="service.product" value="Symantec Endpoint Protection Manager"/>
     <param pos="0" name="service.family" value="Symantec Endpoint Protection Manager"/>
   </fingerprint>
+  <fingerprint pattern="^Intel\(R\) Active Management Technology\s(\d+\.\d+\.\d+\.\d+|\d+\.\d+\.\d+|\d+\.\d+)">
```
jhart-r7: There are older versions that are a subset of AMT called Standard Manageability that has nearly the same fingerprint and is affected by much of the same security issues as of late.   I'll send you a PR

xml/http_servers.xml
```diff
@@ -3283,4 +3283,12 @@
...
+    <description>Intel(R) Active Management Technology</description>
+    <example>Intel(R) Active Management Technology 7.1.86</example>
```
jhart-r7: Any extractions should have tests, IMO.  Will add this as part of my PR against yours.

cmccrisken-r7:
Thanks for the additions @jhart-r7 - I'll land your PR to mine. CVE-2017-5689 affects 6.0-11.6. We should probably enhance the fingerprint to deal with the very old versions but that may have to wait until after this effort, given the turnaround time they want.

PR #129, opened 2017-05-15
Add fingerprints for X11 banners

ekelly-rapid7:
The X11 protocol includes a vendor field when access is successfully
granted. New fingerprints based off this vendor field have been added.

### DIFF
--- a/xml/x11_banners.xml
+++ b/xml/x11_banners.xml
@@ -4,13 +4,13 @@ X11 vendor fields (part of the X11 success response) are matched
 against these patterns to fingerprint servers running the X11 protocol.
 -->
 <fingerprints matches="x11.vendor" protocol="x11">
-  <fingerprint pattern="AT&amp;T Laboratories Cambridge">
+  <fingerprint pattern="^AT&amp;T Laboratories Cambridge$">
     <description>AT&amp;T Laboratories Cambridge</description>
     <example>AT&amp;T Laboratories Cambridge</example>
     <param pos="0" name="service.vendor" value="AT&amp;T Laboratories Cambridge"/>
     <param pos="0" name="service.product" value="Xvnc"/>
   </fingerprint>
-  <fingerprint pattern="CentOS">
+  <fingerprint pattern="^CentOS$">
     <description>CentOS</description>
     <example>CentOS</example>
     <param pos="0" name="os.vendor" value="CentOS"/>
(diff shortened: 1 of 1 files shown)


xml/x11_banners.xml
```diff
@@ -0,0 +1,212 @@
...
+  <fingerprint pattern="Labtam Inc">
+    <description>Labtam Inc</description>
+    <example>Labtam Inc</example>
+    <param pos="0" name="os.vendor" value="Microsoft"/>
+    <param pos="0" name="service.vendor" value="Labtam Inc."/>
+    <param pos="0" name="service.product" value="XSecurePro"/>
+    <param pos="0" name="os.product" value="Windows"/>
+    <param pos="0" name="os.family" value="Windows"/>
+  </fingerprint>
+  <fingerprint pattern="Moba\/X">
```
rkirk-r7: Does the forward slash need to be escaped? Probably does just checking

xml/x11_banners.xml
```diff
@@ -0,0 +1,212 @@
...
+    <param pos="0" name="service.vendor" value="AT&amp;T Laboratories Cambridge"/>
+    <param pos="0" name="service.product" value="Xvnc"/>
+  </fingerprint>
+  <fingerprint pattern="CentOS">
+    <description>CentOS</description>
+    <example>CentOS</example>
+    <param pos="0" name="os.vendor" value="CentOS"/>
+    <param pos="0" name="service.vendor" value="X.Org"/>
+    <param pos="0" name="service.product" value="X.Org X11"/>
+    <param pos="0" name="os.product" value="Linux"/>
```
ncameron-r7: Should we say CentOS here?

xml/x11_banners.xml
```diff
@@ -0,0 +1,212 @@
...
+  <fingerprint pattern="Hummingbird Communications Ltd\.">
+    <description>Hummingbird Communications Ltd.</description>
+    <example>Hummingbird Communications Ltd.</example>
+    <param pos="0" name="os.vendor" value="Microsoft"/>
+    <param pos="0" name="service.vendor" value="Hummingbird Ltd."/>
+    <param pos="0" name="service.product" value="Hummingbird Exceed X server"/>
+    <param pos="0" name="os.product" value="Windows"/>
+    <param pos="0" name="os.family" value="Windows"/>
+  </fingerprint>
+  <fingerprint pattern="Hummingbird Ltd\.">
```
jhart-r7: Combine with the previous and use ^ and $.

xml/x11_banners.xml
```diff
@@ -0,0 +1,212 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+X11 vendor fields (part of the X11 success response) are matched
```
jhart-r7: Could you provide a little more detail here?  What exactly is the response in response to?  Is this the same vendor that is printed in https://github.com/rapid7/metasploit-framework/blob/master/modules/auxiliary/scanner/x11/open_x11.rb?  If so, eventually someone should improve that to have recog coverage too.

ekelly-rapid7: Yeah that's the same field. I'll point to the protocol doc https://www.x.org/archive/X11R7.5/doc/x11proto/proto.pdf and add some additional explanation.

PR #133, opened 2017-06-07
Stmp sonar

nchengusaf:
Initial commit for first batch of fingerprints

Edit by @tsellers-r7: For anyone reviewing this PR - this PR is part of a learning exercise introducing nchengusaf to Recog, Ruby, and regex.  He's being thrown into the deep end..

### DIFF
--- a/xml/smtp_banners.xml
+++ b/xml/smtp_banners.xml
@@ -1,3 +1,4 @@
+asdf
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 SMTP greeting lines (part of the banner after the response code) are matched
--- a/xml/smtp_banners.xml
+++ b/xml/smtp_banners.xml
@@ -1743,7 +1743,7 @@ The system or service fingerprint with the highest certainty overwrites the othe
   </fingerprint>
   <fingerprint pattern="^(\S{0,50}).E?SMTP.([a-zA-Z]{0,50}) (\S{0,50})$">
    <description>
-	internet security directed at content control 
+	Internet security directed at content control 
    </description>
    <example service.website="sjl0vm-hesra17.colo.sonicwall.com" service.product="SonicWALL" service.version="(9.0.1.1225)">sjl0vm-hesra17.colo.sonicwall.com ESMTP SonicWALL (9.0.1.1225)</example>
    <param pos="0" name="service.host" value="Dell Corporation"/>
--- a/xml/smtp_banners.xml
+++ b/xml/smtp_banners.xml
@@ -1,4 +1,4 @@
-asdf
+
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 SMTP greeting lines (part of the banner after the response code) are matched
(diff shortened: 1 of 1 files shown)


xml/smtp_banners.xml
```diff
@@ -1647,4 +1648,93 @@ The system or service fingerprint with the highest certainty overwrites the othe
     <param pos="0" name="service.product" value="Unknown"/>
     <param pos="1" name="host.name"/>
   </fingerprint>
+
```
dabdine-r7: Get rid of the whitespace

xml/smtp_banners.xml
```diff
@@ -1647,4 +1648,93 @@ The system or service fingerprint with the highest certainty overwrites the othe
...
+
+
+
+
+
+
+
+
+        <fingerprint pattern=”^(\w+.\w+.\w)+ Service not available$>
```
dabdine-r7: looks like unicode quotations are used here, and elsewhere.

dabdine-r7: ” instead of "

xml/smtp_banners.xml
```diff
@@ -1647,4 +1648,93 @@ The system or service fingerprint with the highest certainty overwrites the othe
...
+        <description>match to Service not available error
```
dabdine-r7: all the description elements in this file are missing closing tags (`</description>`)

xml/smtp_banners.xml
```diff
@@ -1647,4 +1648,93 @@ The system or service fingerprint with the highest certainty overwrites the othe
...
+        <example service.version=”4.3.2”> 4.3.2 Service not available <example>
+        <param pos="1" name="service.version"/>
+        <param pos=”0” name=”service.vendor” value=”new”/>
+        <param pos=”0” name=”service.family” value=”version”/>
+        <param pos=”0” name=”service.product” value=”version”/>
+        <fingerprint>
+         <fingerprint pattern="^(.+.{4})-\w.+echosting+[^ ]+(cafe+.+.+.[a-z]+[a-z]+[a-z])+ ESMTP$">
+        <description>match to echosting or ESMTP
+        <example service.version="e0000" website.version="cafe24.com"> e0000-e.echosting.cafe24.com ESMTP <example>
```
dabdine-r7: don't include hostnames from the source sonar data (replace hostnames with something like foo.com, and make sure the regular expression still matches).

xml/smtp_banners.xml
```diff
@@ -1647,4 +1648,93 @@ The system or service fingerprint with the highest certainty overwrites the othe
...
+        <param pos="0" name="service.vendor" value="new"/>
+        <fingerprint>
+         <fingerprint pattern="^([a-z]+[a-z]+[a-z]+.)+.cloud.netart +ESMTP +ready$">
+        <description>match to cloud.netart and ESMTP and ready state
+        <example service.version="lbc4"/> lbc4.cloud.netart ESMTP ready <example>
+        <param pos="1" name="service.version"/>
+        <param pos="0" name="service.vendor" value="new"/>
+        <param pos="0" name="service.family" value="version"/>
+        <fingerprint>
+         <fingerprint pattern="^([A-Z]+[A-Z]+[A-Z]+[A-Z])+[^ ]+(\d+[^ ]+\d) (\d\d\d) +Unauthorized$">
```
dabdine-r7: Remove this fingerprint -- it isn't a mail server, but an HTTP server which was included. All sorts of things run on the internet, sometimes we capture HTTP servers running on mail ports.

xml/smtp_banners.xml
```diff
@@ -1647,4 +1648,93 @@ The system or service fingerprint with the highest certainty overwrites the othe
...
+        <description> match to HTTP error
+        <example service.version="HTTP" service.number="1.0" error.number="403"> HTTP/1.0 403 Unauthorized <example>
```
dabdine-r7: Same comment as above (remove this fingerprint)

xml/smtp_banners.xml
dabdine-r7: Also, the fingerprint element here is missing a closing tag

tsellers-r7:
Per comments from Derek, before submitting or updating a PR you should validate that the changes do not cause the tests to fail.

1. `cd <path to clone of recog repo>`
2. Verify that the fingerprints themselves will validate against the examples
    `bin/recog_verify <xml_file_path>`

    For example:
    `bin/recog_verify xml/smtp_banners.xml`

    This may generate many warnings, but no failures.  The warnings are undesirable but don’t break anything. Try not to increase the warning count ;)
3. Run RSpec to run all of the project's unit tests. This will take a while, should complete without errors.

   `rspec`

xml/smtp_banners.xml
```diff
@@ -1647,4 +1648,93 @@ The system or service fingerprint with the highest certainty overwrites the othe
...
+        <param pos="1" name="service.version"/>
+        <param pos="2" name="service.number"/>
+        <param pos="3" name="error.number"/>
+        <fingerprint>
```
tsellers-r7: The above line and similar occurrences at the end of the fingerprint should be a closing `</fingerprint>` tag.

xml/smtp_banners.xml
tsellers-r7: I think we've been trying to standardize on `foo.bar` as the hostname.  The hostname  is data that we capture as part of this process but, with a few exceptions, almost no Recog fingerprints will be tailored to the hosting provider.

xml/smtp_banners.xml
```diff
@@ -1647,4 +1648,93 @@ The system or service fingerprint with the highest certainty overwrites the othe
...
+         <fingerprint pattern="^(\D{2}+\d{3}).secure.ne.jp+ ESMTP">
+        <description> match to secure.ne.jp in Japan
+        <example service.version="aa227"/> aa227.secure.ne.jp ESMTP <example>
+        <param pos="1" name="service.version" />
+        <param pos="0" name="service.vendor" value="japan" />
+        <param pos="0" name="service.family" value="secure"/>
+        <fingerprint>
+         <fingerprint pattern="^(\D+\D+\d+\d).([a-z]+[a-z]+[a-z]+[a-z]+[a-z])+.jp+ ESMTP$">
+        <description> match to domain controller
+        <example service.version="dc51"/> dc51.etius.jp ESMTP <example>
```
tsellers-r7: Be careful about the assertions and assumptions that are expressed as part of a fingerprint.  For example, are you sure that this is a domain controller and that `dc` doesn't stand for Data Center instead?  The `service.vendor` and `service.family` values below are incorrect as well unless there is a vendor named `Japan` with a product family named `etius`.

PR #135, opened 2017-06-14
More robust operating system parsing

vmacdougal-r7:
Adaptive Security uses recog to parse operating systems from outside sources (Azure, Active Directory, etc) when importing assets into Nexpose. Other sources have minor variations in the way they represent OS names (e.g. starting with "SLES" instead of "SUSE SLES", or "Ubuntu-14.04" instead of "Ubuntu 14.04"). Change the OS fingerprinting to also match these minor variations.


gschneider-r7:
Travis-CI is failing on JRuby due to out of memory error. This appears to also be happening on master. Seems to be a Travis-CI environment issue.

vmacdougal-r7:
@gschneider-r7, @egypt, any feedback?

PR #142, opened 2017-06-16
Added Cellopoint Firewall

Xepheron:


xml/smtp_banners.xml
```diff
@@ -1636,7 +1636,16 @@ The system or service fingerprint with the highest certainty overwrites the othe
     <param pos="0" name="service.product" value="Perl"/>
     <param pos="1" name="host.name"/>
   </fingerprint>
-  <fingerprint pattern="^([^ ]+) E?SMTP(?: (?:Service )?Ready\.?)?$" flags="REG_ICASE">
+<fingerprint pattern="Cellopoint E-mail Firewall (v\d+.\d+.\d+) Build (\d+) ready">
+	<description>Matches a Cellopoint Firewall catching the request</description>
+	<example> Cellopoint E-mail Firewall v3.9.11 Build 0730 ready </example>
```
jhart-r7: You should test that the fields are properly extracted here.  Without attributes in `example`, when this is run via `rspec`/etc, as written this will only test that the regular expression matches.

Something like:

```
<example service.version="3.9.11" service.version.version="0730">...</example>

Comment (jhart-r7) [inline xml/smtp_banners.xml]: Minor, but there is a space at the beginning here.  If the space is really there in the response, that is fine, otherwise remove it here and anchor the beginning of the regex above with a `^`

File: xml/smtp_banners.xml [hunk 2]
```diff
@@ -1636,7 +1636,16 @@ The system or service fingerprint with the highest certainty overwrites the othe
...
+	<param pos="0" name="service.vendor" value="Cellopoint"/>
+	<param pos="0" name="service.family" value="Email Firewall"/>
+	<param pos="0" name="service.product" value="Email Firewall"/>
+	<param pos="1" name="service.version"/>
+	<param pos="2" name="service.version.version"/>
+</fingerprint>
+<fingerprint pattern="^([^ ]+) E?SMTP(?: (?:Service )?Ready\.?)?$" flags="REG_ICASE">
```
jhart-r7: Minor, but there is a indentation inconsistency here.  Try to stay consistent.

xml/smtp_banners.xml
```diff
@@ -1636,7 +1636,16 @@ The system or service fingerprint with the highest certainty overwrites the othe
...
   </fingerprint>
-  <fingerprint pattern="^([^ ]+) E?SMTP(?: (?:Service )?Ready\.?)?$" flags="REG_ICASE">
+<fingerprint pattern="Cellopoint E-mail Firewall (v\d+.\d+.\d+) Build (\d+) ready">
+	<description>Matches a Cellopoint Firewall catching the request</description>
+	<example> Cellopoint E-mail Firewall v3.9.11 Build 0730 ready </example>
+	<param pos="0" name="service.vendor" value="Cellopoint"/>
+	<param pos="0" name="service.family" value="Email Firewall"/>
+	<param pos="0" name="service.product" value="Email Firewall"/>
+	<param pos="1" name="service.version"/>
+	<param pos="2" name="service.version.version"/>
```
jhart-r7: I'm not 100% sure this should be here.  Do you have other examples of the build values, in particular as they related to the `service.version`?  For example, do you know if there will ever be other build versions for 3.9.11?

xml/smtp_banners.xml
```diff
@@ -1636,7 +1636,16 @@ The system or service fingerprint with the highest certainty overwrites the othe
     <param pos="0" name="service.product" value="Perl"/>
     <param pos="1" name="host.name"/>
   </fingerprint>
-  <fingerprint pattern="^([^ ]+) E?SMTP(?: (?:Service )?Ready\.?)?$" flags="REG_ICASE">
+<fingerprint pattern="Cellopoint E-mail Firewall (v\d+.\d+.\d+) Build (\d+) ready">
```
jhart-r7: I'd remove the capturing of the `v`, since that really isn't part of the version.

jhart-r7: I'd suggest a leading (`^`) and trailing (`$`) anchor.

PR #144, opened 2017-06-19
Match windows by version

vmacdougal-r7:
Sometimes all that is available is the windows version (e.g. "Windows 6.1" or "Windows 10.0"). This can be partially matched, for instance "Windows 6.1" is either Windows 7 or Windows Server 2008.

### DIFF
--- a/xml/operating_system.xml
+++ b/xml/operating_system.xml
@@ -4,10 +4,11 @@
   -->
 <fingerprints matches="operating_system.name" database_type="util.os" preference="0.80">
   <!-- Windows begin -->
-  <fingerprint pattern="^(?i:(?:Microsoft )?(Windows (?:[a-z]+\s[a-z]+\s|[a-z]+\s)?Server (?:\d{4} R2|\d{4}))(?:,\s|\s)?([a-z]+)?(?: Edition)?(?:\s|\swith(?:out)? Hyper-V\s)?(SP\d|SP \d|Service Pack \d)?)$">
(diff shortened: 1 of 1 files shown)


xml/operating_system.xml
```diff
@@ -89,6 +90,34 @@
     <param pos="0" name="os.family" value="Windows"/>
     <param pos="1" name="os.product"/>
   </fingerprint>
+  <fingerprint pattern="^(?i:(?:Microsoft )?Windows 6.1)$">
```
gschneider-r7: It might be worth including `(?:\sNT)?` between `Windows` and the version number, as that is another way these versions can be displayed.

xml/operating_system.xml
```diff
@@ -89,6 +90,34 @@
     <param pos="0" name="os.family" value="Windows"/>
     <param pos="1" name="os.product"/>
   </fingerprint>
+  <fingerprint pattern="^(?i:(?:Microsoft )?Windows(?:\sNT)? 6.1)$">
+    <description>Windows version 6.1 (Windows 7 or Windows Server 2008)</description>
+    <example>Windows 6.1</example>
+    <param pos="0" name="os.vendor" value="Microsoft"/>
+    <param pos="0" name="os.family" value="Windows"/>
+    <param pos="0" name="os.product" value="Windows 7 or Windows Server 2008"/>
```
gschneider-r7: This should be Server 2008 R2 I believe, as 2008 is actually paired with Vista (6.0)

PR #145, opened 2017-06-20
Fix solaris version matches from ntp

amcclenaghan-r7:
Previously ntp banners for Solaris had their os.product value set to 'Solaris' , however the version was being set to the SunOS version which is different from the Solaris version, see: https://en.wikipedia.org/wiki/Solaris_(operating_system)#Version_history

Due to the first digit of the Solaris version being different from the SunOS version I have created  separate fingerprints for all regex matching SunOS/5.0 -> SunOS/5.6, as their versions have to be hard coded, for example the following regex pattern:
`  <fingerprint pattern="^.*processor=&quot;([^ ]+)&quot;,.*system=&quot;SunOS/5.1&quot;" flags="REG_DOT_NEWLINE,REG_ICASE">
`

Has its os.version set :

`<param pos="0" name="os.version" value="2.1" />
`

From SunOS/5.7 and upwards the Solaris version is just the final digit from the SunOS version so it was possible to capture this within the regex :

`  <fingerprint pattern="^.*processor=&quot;([^ ]+)&quot;,.*system=&quot;SunOS/5.([789]|\d{2,})&quot;" flags="REG_DOT_NEWLINE,REG_ICASE">
`

Has its os.version set:

`   <param pos="2" name="os.version"/>
`

**Testing:**

`bin/recog_verify xml/ssh_banners.xml` completes with 0 failures against new example cases
`Rake tests` passed with 0 failures

### DIFF
--- a/xml/ftp_banners.xml
+++ b/xml/ftp_banners.xml
@@ -129,15 +129,24 @@ example.com FTP server (Version:  Mac OS X Server) ready.</example>
     <param pos="2" name="service.version"/>
     <param pos="1" name="host.name"/>
   </fingerprint>
-  <fingerprint pattern="^(\S+) FTP Server \(SunOS (\S+)\) ready\.?$" flags="REG_ICASE">
+  <fingerprint pattern="^(\S+) FTP Server \(SunOS 5.([789]|\d{2,})\) ready\.?$" flags="REG_ICASE">
     <description>SunOS/Solaris</description>
-    <example>example.com FTP server (SunOS 5.7) ready.</example>
+    <example host.name="example.com" os.version="7">example.com FTP server (SunOS 5.7) ready.</example>
     <param pos="0" name="os.vendor" value="Sun"/>
     <param pos="0" name="os.family" value="Solaris"/>
     <param pos="0" name="os.product" value="Solaris"/>
     <param pos="1" name="host.name"/>
     <param pos="2" name="os.version"/>
   </fingerprint>
+  <fingerprint pattern="^(\S+) FTP Server \(SunOS 5.6\) ready\." flags="REG_ICASE">
+    <description>SunOS 5.6 (Solaris 2.6)</description>
+    <example host.name="example.com">example.com FTP Server (SunOS 5.6) ready.</example>
+    <param pos="0" name="os.vendor" value="Sun"/>
+    <param pos="0" name="os.family" value="Solaris"/>
+    <param pos="0" name="os.product" value="Solaris"/>
+    <param pos="0" name="os.version" value="2.6"/>
+    <param pos="1" name="host.name"/>
+  </fingerprint>  
   <fingerprint pattern="^ProFTPD (\d+\.[^\s]+) Server \(Debian\) \[(.+)\]$">
     <description>ProFTPD on Debian Linux</description>
     <example>ProFTPD 1.3.0rc2 Server (Debian) [host]</example>
--- a/xml/ftp_banners.xml
+++ b/xml/ftp_banners.xml
@@ -132,7 +132,6 @@ example.com FTP server (Version:  Mac OS X Server) ready.</example>
   <fingerprint pattern="^(\S+) FTP Server \(SunOS 5.(1[1-9])\) ready\.?$" flags="REG_ICASE">
     <description>SunOS/Solaris</description>
     <example host.name="example.com" os.version="11">example.com FTP server (SunOS 5.11) ready.</example>
-    <example host.name="example.com" os.version="12">example.com FTP server (SunOS 5.12) ready.</example>
     <param pos="0" name="os.vendor" value="Oracle"/>
     <param pos="0" name="os.family" value="Solaris"/>
     <param pos="0" name="os.product" value="Solaris"/>
--- a/xml/ftp_banners.xml
+++ b/xml/ftp_banners.xml
@@ -129,8 +129,19 @@ example.com FTP server (Version:  Mac OS X Server) ready.</example>
     <param pos="2" name="service.version"/>
     <param pos="1" name="host.name"/>
   </fingerprint>
-  <fingerprint pattern="^(\S+) FTP Server \(SunOS 5.([789]|\d{2,})\) ready\.?$" flags="REG_ICASE">
+  <fingerprint pattern="^(\S+) FTP Server \(SunOS 5.(1[1-9]|[2-9]\d{1,}|\d{3,})\) ready\.?$" flags="REG_ICASE">
     <description>SunOS/Solaris</description>
+    <example host.name="example.com" os.version="11">example.com FTP server (SunOS 5.11) ready.</example>
+    <example host.name="example.com" os.version="21">example.com FTP server (SunOS 5.21) ready.</example>
+    <example host.name="example.com" os.version="123">example.com FTP server (SunOS 5.123) ready.</example>
+    <param pos="0" name="os.vendor" value="Oracle"/>
+    <param pos="0" name="os.family" value="Solaris"/>
+    <param pos="0" name="os.product" value="Solaris"/>
+    <param pos="1" name="host.name"/>
+    <param pos="2" name="os.version"/>
+  </fingerprint>
+  <fingerprint pattern="^(\S+) FTP Server \(SunOS 5.([789]|10)\) ready\.?$" flags="REG_ICASE">
+    <description>SunOS/Solaris 5.7-5.10</description>
     <example host.name="example.com" os.version="7">example.com FTP server (SunOS 5.7) ready.</example>
     <example host.name="example.com" os.version="10">example.com FTP server (SunOS 5.10) ready.</example>
     <param pos="0" name="os.vendor" value="Sun"/>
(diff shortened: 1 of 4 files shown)


xml/ntp_banners.xml
```diff
@@ -419,49 +419,247 @@ NTP "banners", taken from a readvar response
...
+    <param pos="0" name="service.product" value="NTP"/>
+    <param pos="0" name="os.version" value="2.2"/>
+    <param pos="1" name="service.version"/>
+    <param pos="2" name="os.arch"/>
+  </fingerprint>
+  <fingerprint pattern="^.*version=&quot;ntpd ([^ ]+)[^&quot;]+&quot;,.*processor=&quot;([^ ]+)&quot;,.*system=&quot;SunOS/5.3&quot;" flags="REG_DOT_NEWLINE,REG_ICASE">
+    <description> ntpd running on Solaris 2.3 (SunOS/5.3) </description>
     <example>
-      processor="sun4u", system="SunOS5.10"
+      version="ntpd 4.2.0@1.1161-r Wed Apr 20 11:28:05 EDT 2005 (1)",
```
jhart-r7: While here it'd be great if you could add the expected extractions to any `example` elements you are adding.

For example:

```
   <example service.version="4.2.0@1.1161-r" os.arch="sun4u">...
```

jhart-r7:
There are several other locations where this same pattern is used.  For example, https://github.com/rapid7/recog/blob/master/xml/snmp_sysdescr.xml#L7058-L7071, https://github.com/rapid7/recog/blob/master/xml/snmp_sysdescr.xml#L7263-L7282 and probably more.  If this is changed for NTP, I think it should be changed for all the others we know about unless there is something I'm missing.

ncameron-r7:
Yeah if this is similar behaviour across snmp and potentially other protocols then we should probably change it everywhere to reflect the better fp. @amcclenaghan-r7 do you have a target that includes ntp and some other services for easy testing?

xml/ntp_banners.xml
amcclenaghan-r7: Thanks Jon, I will incorporate the expected extractions, I have some additional testing to do as well  with these changes before this is ready for merging. Will have a look into your SNMP comment was not aware of those, it is probably worth fixing that while we're here

ncameron-r7: I agree Adam this will allow you to test that the right fingerprint would get asserted based on your example banner

xml/ntp_banners.xml
```diff
@@ -419,49 +419,247 @@ NTP "banners", taken from a readvar response
     <param pos="0" name="os.family" value="NetWare"/>
     <param pos="0" name="os.product" value="NetWare"/>
   </fingerprint>
-  <fingerprint pattern="^.*version=&quot;ntpd ([^ ]+)[^&quot;]+&quot;,.*processor=&quot;([^ ]+)&quot;,.*system=&quot;SunOS/?([^ ]+)&quot;" flags="REG_DOT_NEWLINE,REG_ICASE">
-    <description>ntpd running on Solaris</description>
+  <fingerprint pattern="^.*version=&quot;ntpd ([^ ]+)[^&quot;]+&quot;,.*processor=&quot;([^ ]+)&quot;,.*system=&quot;SunOS/5.0&quot;" flags="REG_DOT_NEWLINE,REG_ICASE">
+    <description> ntpd running on Solaris 2.0 (SunOS/5.0) </description>
```
ncameron-r7: You have a few wee places like this for description elements

amcclenaghan-r7: Will update the regex back to the original format with /? to take account of the strings where it is 'SunOS5.x' as per that shodan.io we just looked at and add example cases, thanks @ncameron-r7 

Link for anyone else interested: https://www.shodan.io/search?query=SunOS5.6

xml/ftp_banners.xml
```diff
@@ -129,15 +129,24 @@ example.com FTP server (Version:  Mac OS X Server) ready.</example>
     <param pos="2" name="service.version"/>
     <param pos="1" name="host.name"/>
   </fingerprint>
-  <fingerprint pattern="^(\S+) FTP Server \(SunOS (\S+)\) ready\.?$" flags="REG_ICASE">
+  <fingerprint pattern="^(\S+) FTP Server \(SunOS 5.([789]|\d{2,})\) ready\.?$" flags="REG_ICASE">
     <description>SunOS/Solaris</description>
-    <example>example.com FTP server (SunOS 5.7) ready.</example>
+    <example host.name="example.com" os.version="7">example.com FTP server (SunOS 5.7) ready.</example>
```
ekelly-rapid7: Should probably add a check to verify that double digit versions match OK. 
SunOS 5.11 -> Solaris 11

amcclenaghan-r7: Pushed an update with that

xml/ftp_banners.xml
```diff
@@ -129,16 +129,27 @@ example.com FTP server (Version:  Mac OS X Server) ready.</example>
     <param pos="2" name="service.version"/>
     <param pos="1" name="host.name"/>
   </fingerprint>
-  <fingerprint pattern="^(\S+) FTP Server \(SunOS 5.([789]|\d{2,})\) ready\.?$" flags="REG_ICASE">
+  <fingerprint pattern="^(\S+) FTP Server \(SunOS 5.(1[1-9]|[2-9]\d{1,}|\d{3,})\) ready\.?$" flags="REG_ICASE">
```
jhart-r7: Is the increasingly high 2 digit 3+ digit version numbers realistic?  AFAIK, Solaris 123 is not a thing.

xml/ntp_banners.xml
```diff
@@ -419,50 +419,382 @@ NTP "banners", taken from a readvar response
...
+   <param pos="0" name="os.product" value="Solaris"/>
+   <param pos="0" name="os.version" value="2.0" />
+   <param pos="1" name="os.arch"/>
+  </fingerprint>
+  <fingerprint pattern="^.*processor=&quot;([^ ]+)&quot;,.*system=&quot;SunOS/?5.1&quot;" flags="REG_DOT_NEWLINE,REG_ICASE">
+   <description>Solaris 2.1 (SunOS/5.1) with no ntp version</description>
+   <example os.arch="sun4m">
+     processor="sun4m", system="SunOS/5.1"
+   </example>
+   <example os.arch="sun4u">
```
jhart-r7: Really minor, but when adding examples, in particular those in which you are testing the extracted values, I prefer to only test banner/other data that is a distinct use-case covered by the regex in question.  For example, here and throughout this PR, there are cases where you are ensuring that a regex like `([^ ]+)` properly captures non-whitespace properly.  If it captures `sun4u`, it will catch `sun4m`.  

IOW, the purpose of the examples and their tests is not to list all banners we've seen that we expect to match, but should be the unique subset of those that specifically test the particulars of the regex.

The reason this is a little important is that every added example increases the time that the tests take to run, in particular because *every* example is tested against *every* regex (in the same file) that came before it to ensure that the example matches the correct regex, since order is important in these files.  So, for this single fingerprint where you've added 3 somewhat duplicate examples, each of these must be tested against the dozens of `pattern`s that came before it.  Over time this makes the tests take longer and longer without providing additional coverage.

Contrast this with a regex that did something like `(\d+|alpha|beta)`, for capturing a version field that might have digits in it or might simply have "alpha" or "beta".  In cases like this, IMO it would be appropriate to add examples for each of these 3 cases, since those are 3 distinct things that the regex is designed to capture.

xml/operating_system.xml
```diff
@@ -482,15 +482,33 @@
...
     <param pos="0" name="os.vendor" value="Sun"/>
     <param pos="0" name="os.family" value="Solaris"/>
     <param pos="0" name="os.product" value="Solaris"/>
     <param pos="1" name="os.version"/>
   </fingerprint>
+  <fingerprint pattern="^(?i:(?:Oracle|Sun)?\s?SunOS\s?5.([789]|10)?)$">
+    <description>SunOS/Solaris 5.7-5.10</description>
+    <example os.version="7">SunOS 5.7</example>
+    <example os.version="10">SunOS 5.10</example>
+    <param pos="0" name="os.vendor" value="SunOS"/>
```
jhart-r7: Shouldn't this be Solaris

xml/snmp_sysdescr.xml
```diff
@@ -7058,9 +7058,24 @@ Copyright (c) 1995-2005 by Cisco Systems
...
    =======================================================================-->
-  <fingerprint pattern="^SunOS (\S+) 5\.(\S+) \S+ (\S+)$">
-    <description>SunOS/Solaris</description>
-    <example>SunOS sysname 5.10 Generic_118833-33 sun4u</example>
+  <fingerprint pattern="^SunOS (\S+) 5\.(1[1-9]|[2-9]\d{1,}|\d{3,}) \S+ (\S+)$">
+   <description>Oracle/Solaris 5.11 and upwards</description>
+   <example host.name="sysname" os.version="11" os.arch="sun4u">SunOS sysname 5.11 Generic_118833-33 sun4u</example>
+   <example host.name="sysname" os.version="20" os.arch="sun4u">SunOS sysname 5.20 Generic_118833-33 sun4u</example>
+   <example host.name="sysname" os.version="123" os.arch="sun4u">SunOS sysname 5.123 Generic_118833-33 sun4u</example>
+   <param pos="0" name="os.vendor" value="oracle"/>
```
jhart-r7: Oracle

xml/operating_system.xml
amcclenaghan-r7: You're right Jon, missed that, thanks!

xml/ntp_banners.xml
amcclenaghan-r7: I agree, I noticed in the ntp_banners.xml before my changes there were examples such as : 

`    <example>
      processor="sun4m", system="SunOS5.6"
    </example>
`

`    <example>
      processor="sun4u", system="SunOS5.6"
    </example>
`

Within the same fingerprint. My thoughts were the same as yours but I thought we maybe tried to list banners seen rather than unique subsets after seeing these, I will adjust this, cheers for the clarification! :)

PR #149, opened 2017-08-16
Add support for DNS version.bind responses

tsellers-r7:
This PR adds support for fingerprinting text string response from a DNS version.bind request.

For example, the string `dnsmasq-2.76-1-ubnt2` emitted by the command below:

```
$ nslookup -type=txt -class=chaos VERSION.BIND <dns_server> | grep VERSION.BIND | cut -d\" -f2
dnsmasq-2.76-1-ubnt2
```
There are currently 49 fingerprints in the submission.  These fingerprints have been grouped by product and roughly ordered by frequency of occurrence based on Project Sonar studies of the public Internet.

**Note:** Per RFC the DNS queries should be case insensitive.  Microsoft doesn't follow this RFC and so Microsoft Windows DNS servers will respond to `version.bind` but not to `VERSION.BIND` requests.  Also, by default version responses are disabled on modern Microsoft Windows DNS servers.

### DIFF
--- a/xml/dns_versionbind.xml
+++ b/xml/dns_versionbind.xml
@@ -1,0 +1,136 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<fingerprints matches="dns_versionbind" protocol="dns" database_type="service" preference="0.20">
+  <!-- RedHat package naming: 
+      https://fedoraproject.org/wiki/Packaging:DistTag
+      https://fedoraproject.org/wiki/Packaging:Versioning
+  -->
+  <fingerprint pattern="^(9.[^-]+)-(?:P\d-)?RedHat-[\d.]+-([\w.]+el([\d_]+)(?:.[\w.]+)?)$">
+    <description>ISC BIND, RedHat Enterprise Linux</description>
+    <example service.version="9.8.2rc1">9.8.2rc1-RedHat-9.8.2-0.62.rc1.el6_9.2</example>
+    <example service.version.version="38.el7_3.3">9.9.4-RedHat-9.9.4-38.el7_3.3</example>
+    <example os.version="5_11">9.3.6-P1-RedHat-9.3.6-25.P1.el5_11.12</example>
+    <example os.version="6_1">9.7.3-P3-RedHat-9.7.3-2.el6_1.P3.3</example>
+    <example os.version="6">9.8.2rc1-RedHat-9.8.2-0.47.rc1.el6</example>
+    <param pos="0" name="service.vendor" value="ISC"/>
+    <param pos="0" name="service.family" value="BIND"/>
+    <param pos="0" name="service.product" value="BIND"/>
+    <param pos="1" name="service.version"/>
+    <param pos="2" name="service.version.version"/>
+    <param pos="0" name="os.vendor" value="RedHat"/>
+    <param pos="0" name="os.device" value="General"/>
+    <param pos="0" name="os.family" value="Linux"/>
+    <param pos="0" name="os.product" value="Enterprise Linux"/>
+    <param pos="3" name="os.version"/>
+  </fingerprint>
+  <fingerprint pattern="^(9.[^-]+)-(?:P\d-)?RedHat-[\d.]+-([\w.]+fc([\d]+))$">
+    <description>ISC BIND, Fedora</description>
+    <example service.version="9.10.4">9.10.4-P8-RedHat-9.10.4-4.P8.fc25</example>
+    <example service.version.version="10.P3.fc23">9.10.3-P3-RedHat-9.10.3-10.P3.fc23</example>
+    <example os.version="10">9.5.2-RedHat-9.5.2-1.fc10</example>
+    <param pos="0" name="service.vendor" value="ISC"/>
+    <param pos="0" name="service.family" value="BIND"/>
+    <param pos="0" name="service.product" value="BIND"/>
+    <param pos="1" name="service.version"/>
+    <param pos="2" name="service.version.version"/>
+    <param pos="0" name="os.vendor" value="Fedora Project"/>
+    <param pos="0" name="os.device" value="General"/>
+    <param pos="0" name="os.family" value="Linux"/>
+    <param pos="0" name="os.product" value="Fedora"/>
+    <param pos="3" name="os.version"/>
+  </fingerprint>
+  <fingerprint pattern="^dnsmasq-(\d.[\w]+)$">
+    <description>dnsmasq, simple</description>
+    <example service.version="2.40">dnsmasq-2.40</example>
+    <example service.version="2.63rc6">dnsmasq-2.63rc6</example>
+    <example service.version="2.76test8">dnsmasq-2.76test8</example>
+    <param pos="0" name="service.vendor" value="Simon Kelley"/>
+    <param pos="0" name="service.family" value="Dnsmasq"/>
+    <param pos="0" name="service.product" value="Dnsmasq"/>
+    <param pos="1" name="service.version"/>
+  </fingerprint>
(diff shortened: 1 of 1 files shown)


xml/dns_versionbind.xml
```diff
@@ -0,0 +1,685 @@
...
+<fingerprints matches="dns_versionbind" protocol="dns" database_type="service" preference="0.750">
+  <!-- RedHat package naming: 
+      https://fedoraproject.org/wiki/Packaging:DistTag
+      https://fedoraproject.org/wiki/Packaging:Versioning
+  -->
+  <fingerprint pattern="^(9.[^-]+(?:-rpz\d?[+.]rl[\d.]+)?(?:-[SP]\d)?)-RedHat-[\d.]+[-.][\w.]+el([\d_]+)(?:.[\w.]+)?$">
+    <description>ISC BIND: RedHat Enterprise Linux</description>
+    <example service.version="9.8.2rc1">9.8.2rc1-RedHat-9.8.2-0.62.rc1.el6_9.2</example>
+    <example service.version="9.9.4">9.9.4-RedHat-9.9.4-38.el7_3.3</example>
+    <example service.version="9.3.6-P1" os.version="5_11">9.3.6-P1-RedHat-9.3.6-25.P1.el5_11.12</example>
```
jhart-r7: I don't believe there is a good way to handle this in recog, but shouldn't the version be 5.11, not 5_11?

xml/dns_versionbind.xml
```diff
@@ -0,0 +1,685 @@
...
+    <example service.version="9.9.5">9.9.5-3-zentyal1-Ubuntu</example>
+    <param pos="0" name="service.vendor" value="ISC"/>
+    <param pos="0" name="service.family" value="BIND"/>
+    <param pos="0" name="service.product" value="BIND"/>
+    <param pos="1" name="service.version"/>
+    <param pos="0" name="os.vendor" value="Ubuntu"/>
+    <param pos="0" name="os.family" value="Linux"/>
+    <param pos="0" name="os.product" value="Zentyal"/>
+  </fingerprint>
+  <fingerprint pattern="^(9.[^-]+-rpz\d?[+.]rl[\d.]+(?:-[SP]\d)?)-Ubuntu-[\d\.:]+[\w\.]+(?:-[SP]\d)?-\d?ubuntu[\d\.]+\+zentyal\d+$">
```
jhart-r7: can this and the previous one be combined?  they both have identical `param`s so I suspect you could just use a `|` in the regex capture group that extracts the version.

xml/dns_versionbind.xml
```diff
@@ -0,0 +1,685 @@
...
+  </fingerprint>
+  <fingerprint pattern="^Served by POWERDNS (\d\.[\d.]+) \$Id[^$]*\$$">
+    <description>PowerDNS: Served by format with version</description>
+    <example service.version="2.9.22">Served by POWERDNS 2.9.22 $Id: packethandler.cc 1321 2008-12-06 19:44:36Z ahu $</example>
+    <param pos="0" name="service.vendor" value="PowerDNS"/>
+    <param pos="0" name="service.family" value="PowerDNS"/>
+    <param pos="1" name="service.version"/>
+  </fingerprint>
+  <fingerprint pattern="^Served by PowerDNS - https?:\/\/www.powerdns.com\/?$">
+    <description>PowerDNS: Served by format without versionl</description>
```
jhart-r7: versionl?

xml/dns_versionbind.xml
```diff
@@ -0,0 +1,685 @@
...
+  </fingerprint>
+  <fingerprint pattern="^unbound ([\d.]+)$">
+    <description>NLnet Labs Unbound</description>
+    <example service.version="1.4.22">unbound 1.4.22</example>
+    <param pos="0" name="service.vendor" value="NLnet Labs"/>
+    <param pos="0" name="service.family" value="Unbound"/>
+    <param pos="0" name="service.product" value="unbound"/>
+    <param pos="1" name="service.version"/>
+  </fingerprint>
+  <fingerprint flags="REG_ICASE" pattern="^unbound$">
```
jhart-r7: here and wherever else you are using `flags`, the preferred method is to write the regex such that it implements the equivalent of these flags but without the GNU.regex specific names.  In this case, `^(?i:unbound)$` would do it.

xml/dns_versionbind.xml
```diff
@@ -0,0 +1,685 @@
...
+    <param pos="0" name="service.vendor" value="PowerDNS"/>
+    <param pos="0" name="service.family" value="PowerDNS"/>
+    <param pos="0" name="service.product" value="Authoritative Server"/>
+    <param pos="1" name="service.version"/>
+  </fingerprint>
+  <fingerprint pattern="^Served by POWERDNS (\d\.[\d.]+) \$Id[^$]*\$$">
+    <description>PowerDNS: Served by format with version</description>
+    <example service.version="2.9.22">Served by POWERDNS 2.9.22 $Id: packethandler.cc 1321 2008-12-06 19:44:36Z ahu $</example>
+    <param pos="0" name="service.vendor" value="PowerDNS"/>
+    <param pos="0" name="service.family" value="PowerDNS"/>
```
jhart-r7: no product?

xml/dns_versionbind.xml
```diff
@@ -0,0 +1,685 @@
...
+  This fingerprint file matches the text string response from a DNS
+  version.bind request.
+
+  For example, the string 'dnsmasq-2.76-1-ubnt2' emitted by the command below:
+
+  $ nslookup -type=txt -class=chaos VERSION.BIND <dns_server> | grep VERSION.BIND | cut -d\" -f2
+  dnsmasq-2.76-1-ubnt2
+
+-->
+<fingerprints matches="dns_versionbind" protocol="dns" database_type="service" preference="0.750">
```
jhart-r7: There is no standard on this necessarily, but `dns.versionbind` is probably more consistent for `matches`

tsellers-r7: Perhaps, this has been an annoyance for me in the past.  The files are typically named one way and the `matches` another (underscore vs dots).  Since only some of the files have a `matches` section its hit or miss as which name works when you are calling Recog's databases programatically.  My preference would be to pick one naming scheme and stick with it or, failing that, implement a `matches` section for all files using the dot syntax. Unfortunately we'd potentially implement a breaking change for someone's workflow if we did that.  In this case I've made it simple by setting both values to be the same.

xml/dns_versionbind.xml
tsellers-r7: Thanks, fixed

xml/dns_versionbind.xml
tsellers-r7: PowerDNS currently has two products that don't have any overlapping functionality, Authoritative Server and Recursor.  I can't tell which product this is by this fingerprint.  If I had to set a value, given the age, I'd call it Authoritative Server.

jhart-r7: Cool.  Thats fine if you don't know.  Guessing is ok in some situations but I wouldn't in this one.  May be worth just adding as a comment.

xml/dns_versionbind.xml
jhart-r7: Ah, I hadn't made the connection between the two different naming problems and agree that is problematic.

Part of the complication here is that there is the implication that the name of the files matter and perhaps aside from some nuanced ordering issues and the fact that external consumers of recog XML content might depend on the names of the files, the names of the files don't matter but `matches` does.

This is complicated by the fact that before `matches`, the name of th file was used to imply what the file contained and how it was to be used.

For now, do whatever makes the most sense for you.  Longer term we can find a better solution for this.

PR #150, opened 2017-08-29
Specify lower bound on character counts

jhart-r7:
This fixes an edge case when Recog content is used in isolation in systems whose regular expression libraries do not support a lack of a lower bound on a count of a characters, like `\d{,2}`, which should match 1-2 digits.


gschneider-r7:
Actually does it need to be `\d{0,2}` to match correctly with these example?

And should the `.` be escaped to `\.` as a literal?

PR #152, opened 2017-08-31
Improvements to CentOS, Debian, Fedora and Linux; Photon and Alpine support; Linux "catch-all" fallback

mhuffman-r7:


xml/operating_system.xml
```diff
@@ -147,34 +156,37 @@
...
     <param pos="0" name="os.product" value="Linux"/>
     <param pos="1" name="os.version"/>
   </fingerprint>
-  <fingerprint pattern="^(?i:Debian(?: Linux)?\s?(\d+?(?:\.\d+?)*?)?(?:\s[a-z\(\)]+)?)$">
+  <fingerprint pattern="^(?i:Debian(?: (?:GNU\/)?Linux)?\s?((?:\d+?(?:\.\d+?)*?)|(?:\w+?\/sid\s?))?(?:\s[a-z\(\)]+)?)$">
     <description>Debian Linux</description>
     <example os.version="6.0">Debian 6.0</example>
     <example os.version="7">Debian 7 (Wheezy)</example>
     <example os.version="8">Debian Linux 8</example>
+    <example os.version="stretch/sid">Debian GNU/Linux stretch/sid</example>
```
gschneider-r7: This one has been an open question still, but I think it is correct to support this version format for Debian.

xml/operating_system.xml
```diff
@@ -130,6 +130,15 @@
   <!-- Windows end -->
 
   <!-- Liunx begin -->
+  <fingerprint pattern="^(?i:Alpine Linux\s?(?:v)?(\d+?(?:\.\d+?)*?)?)$">
+    <description>Alpine Linux</description>
+    <example os.vendor="Alpine">Alpine Linux v3.4.0</example>
+    <example os.version="3.4.0">Alpine Linux v3.4.0</example>
```
gschneider-r7: We could collapse this to a single example I think

jhart-r7: Agreed.  Also, there is no need to test `os.vendor` in this case because it is hardcoded.  If it was capture from a regular expression, like `os.version`, then you should test it.

xml/operating_system.xml
```diff
@@ -315,6 +327,25 @@
     <param pos="1" name="os.version"/>
     <param pos="2" name="os.edition"/>
   </fingerprint>
+  <fingerprint pattern="^(?i:VMWare Photon(?:\/)?(?:\s?Linux)?\s?(?:v)?(\d+?(?:\.\d+?)*?)?)$">
+    <description>Photon Linux</description>
+    <example os.vendor="VMWare" os.product="Photon Linux">VMWare Photon Linux</example>
```
jhart-r7: It is fine to test this value, but there is no need to test `os.vendor` or `os.product` in this case because you hardcoded them.

xml/operating_system.xml
```diff
@@ -315,6 +326,25 @@
...
+    <example os.version="1.0">VMWare Photon 1.0</example>
+    <param pos="0" name="os.vendor" value="VMWare"/>
+    <param pos="0" name="os.family" value="Linux"/>
+    <param pos="0" name="os.product" value="Photon Linux"/>
+    <param pos="1" name="os.version"/>
+  </fingerprint>
+  <!-- Vendor-based distribution catch-call -->
+  <fingerprint pattern="^(?i:(.*)\sLinux?\s(.*))$">
+    <description>Vendor-based Linux catch-all</description>
+    <example os.vendor="Aurox">Aurox Linux 10.2</example>
```
jhart-r7: Minor, but may as well test the version extraction too

gschneider-r7:
The branch build failure seems unrelated to my changes, and the PR build looks green.

jhart-r7:
@gschneider-r7 I think this branch is quite stale and could use a merge from master to make it current.  It tested with much older rubies, indicating it is stale.

PR #153, opened 2017-09-05
Limit branch builds to master branch only

gschneider-r7:
Avoid duplicating travis builds on tag pushes and PRs from origin branches. This doesn't really do anything for us directly, but it is nice to not spend extra cycles unnecessarily on a free service.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,3 +14,6 @@ env:
   global:
     - CI="travis"
     - JRUBY_OPTS="--server -J-Xms512m -J-Xmx2G"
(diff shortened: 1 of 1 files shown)


jhart-r7:
Does this mean that travis won't run on PRs?

gschneider-r7:
It will run on PRs and commits to master branch only. If you compare this PR's status to my other PR you'll notice 1 check vs 2 checks.

jhart-r7:
Ah right, I see.

PR #160, opened 2017-12-04
smtp_banners: Update using Project Sonar data from 2017.11.30 and 2018.04.05

tsellers-r7:
This PR updates the coverage of `xml/smtp_banners.xml` using data from Project Sonar's SMTP 25/tcp study on 2017.11.30.  Additionally, there was significant reorganization and cleanup of the file.

**Note: Due to effort to cleanup description lines (remove duplicates, remove multilines, provide context, standardize format) almost every value for `<description>` has changed. This *will* impact the value returned as `matched`.**

Original fingerprint count: 129 
Original fingerprint count: 133

#### Significant changes:

- Improved the accuracy and/or flexibility of multiple fingerprints.

- Changed *ALL* instances of `flags="REG_ICASE`  to an inline flag (`(?i:`) in order to make the regex compatible with more languages.

- Implemented fingerprint examples for those fingerprints where examples could be found. This sometimes resulted in removing fingerprints that were actually duplicates or trivially different.

- Reworked `description` fields so as to remove examples and ensure that this field is unique within the file as the value of description serves as an identifier when processing fingerprints. Multiline descriptions were reduced to single line where possible. Almost every description was modified.

- Fixed multiple instances where captures where under/over capturing

- Fixed multiple instances where the portion of the version banner that was captured was different between two products in the same family.

- removed various real and example hostnames from examples and standardized on `foo.bar`

- Corrected `system.time.format` so as to match timestamp provided by service

- Reworked date regex for multiple matches to remove inadvertent requirement for two digit day value when the banner included a single digit day.

**Note: A few tweaks were made after the metrics below were generated and so the match % is higher**

#### Overall fingerprint matches
| Total| 2018-04-05 BEFORE | 2018-04-05 AFTER |
|:---------|------------:|------------:|
| 6,561,395 |  5,364,381 | 5,490,301 |

### New fingerprints and metrics
| Fingerprint | 2017-11-30 Dataset |2018-04-05 Dataset|
|:---------|-------------:|------------:|
| Kerio Connect | 12,862 | 12,713 |
| Ecelerity | 15,719 | 15,084 |
| MailEnable - Simple | 6,344 | 6,780|
| SonicWall Email Security | 5,729 | 5,315 |
| Postfix - Ubuntu, Mail-in-a-Box package | | 5,048 |
| IBM Domino SMTP MTA | 3,632 | 4,472 |
| Twisted SMTP server | 1,842 | 4,165 |
| Communigate Pro | 4,284 |  4,303 |
| PowerMTA | 4,242 | 3,878 |
| Lyris ListManager | 2,981 | 2,742 |
| Ma Jian WinWebMail | 2,982 | 2,726 |
| Sendmail - Debian patch only |  3,955 | 3,664 |
| Sendmail - Debian 7.x (wheezy) | 8,894 | 1,400 |
| Sendmail - Debian 8.x (jessie) | 1,239  |  1,147 |
| Sendmail - Debian 5.x (lenny) | 486 |  400 |
| Sendmail - Debian 3.1 (sarge) | 335 | 334 |
| Sendmail - Debian 4.x (etch) | 122 | 64 |
| Tobit Software David | 1,988 | 1,907 |
| Cellopoint E-mail Firewall | 9,096 |  1,842 |

#### Existing fingerprint value shifts of interest

| Fingerprint | 2017-11-30 - BEFORE |2017-11-30 - AFTER | |2018-04-05 BEFORE | 2018-04-05 AFTER |
|:---------|-------------:|------------: |-|------------:|------------:|
| Null (unmatched) | 1,260,499 | 1,124,855 | | 1,197,014 | 1,071,094 |
| Exim with version string and optional timestamp | 832,821  | 833,333   | | 835,880 |  837,506 |
| Sendmail - Ubuntu | 3,495 | 5,811  | | 1,950 | 3,993  |
| Lotus Domino SMTP MTA | 2,949  | 3,632  | | 2,652 |  3,266 |
| MailEnable - Complex | 80,103  |  97,080 | | 79,843  | 97,076 |
| Postfix - generic banner with amusing comments in parentheses | 17,438  | 13,805 | | 17,407  | 13,845  |
| JAMES SMTP Server | | | | 0| 1,256 |
| A.K.I. PMail | | | | 0| 587 |

`JAMES SMTP Server` picked up 1,256 matches (from 0 matches) in the 2018.04.05 data set simply due to fixing the regex which required 2 digit day values (`05`) which the product didn't emit. Similar for `A.K.I. PMail` which picked up 587 (from 0) after date change and fingerprint tweak.

### DIFF
--- a/xml/smtp_banners.xml
+++ b/xml/smtp_banners.xml
@@ -1058,9 +1058,7 @@ The system or service fingerprint with the highest certainty overwrites the othe
     <param pos="1" name="host.name"/>
   </fingerprint>
   <fingerprint pattern="^([^ ]+) ESMTP.* Postfix *$">
-    <description>
-         Generic Postfix banner.
-      </description>
+    <description>Generic Postfix banner.</description>
     <example>foo.bar.com ESMTP Postfix</example>
     <param pos="0" name="service.family" value="Postfix"/>
     <param pos="0" name="service.product" value="Postfix"/>
@@ -1115,10 +1113,14 @@ The system or service fingerprint with the highest certainty overwrites the othe
     <param pos="2" name="service.version"/>
     <param pos="1" name="host.name"/>
   </fingerprint>
+  <fingerprint pattern="^Sendmail ESMTP ready$">
+    <description>Sendmail - short banner w/o hostname, version, platform, or date.</description>
+    <example>Sendmail ESMTP ready</example>
+    <param pos="0" name="service.family" value="Sendmail"/>
+    <param pos="0" name="service.product" value="Sendmail"/>
+  </fingerprint>
   <fingerprint pattern="^([^ ]+) +ESMTP +Sendmail +([^ ]+) \(PHNE_([^ ]+)\) */ *(.+); *(.+) \(.+\)$">
-    <description>
-         sendmail on HPUX with a PHNE (HP Networking patch) installed
-      </description>
+    <description>Sendmail - HP-UX with a PHNE (HP Networking patch) installed</description>
     <example>foo.bar.com ESMTP Sendmail 8.8.6 (PHNE_14041)/8.7.1; Tue, 6 Feb 2001 10:04:32 -0300 (SAT)</example>
     <param pos="0" name="service.family" value="Sendmail"/>
     <param pos="0" name="service.product" value="Sendmail"/>
@@ -1134,9 +1136,7 @@ The system or service fingerprint with the highest certainty overwrites the othe
     <param pos="5" name="system.time"/>
   </fingerprint>
   <fingerprint pattern="^(\S+) ESMTP Sendmail \S+ version ([\d\.]+) - Revision \S+ HP-UX([\d\.]+).*(\S{3}, \d{2} \S{3} \d{4} \d{2}:\d{2}:\d{2} \S{3})$">
-    <description>
-         sendmail on HPUX
-       </description>
+    <description>Sendmail - HP-UX</description>
     <example host.name="example.com" os.version="11.31" service.version="8.13.3">example.com ESMTP Sendmail @(#)Sendmail version 8.13.3 - Revision 1.004:: HP-UX11.31 - 03rd February,2010/8.11.1; Wed, 20 May 2015 23:35:38 GMT</example>
     <param pos="0" name="service.family" value="Sendmail"/>
     <param pos="0" name="service.product" value="Sendmail"/>
@@ -1151,9 +1151,7 @@ The system or service fingerprint with the highest certainty overwrites the othe
     <param pos="4" name="system.time"/>
   </fingerprint>
   <fingerprint pattern="^([^ ]+) +ESMTP +Sendmail +([^ ]+)/UW([^ ]+) ready at *(.+) \(.+\) *$">
-    <description>
-       sendmail on unixware
-      </description>
+    <description>Sendmail - Unixware</description>
     <example>foo.bar.com ESMTP Sendmail 8.8.7/UW7.1.0 ready at Tue, 6 Feb 2001 16:39:30 -0300 (GMT-0300)</example>
     <param pos="0" name="service.family" value="Sendmail"/>
     <param pos="0" name="service.product" value="Sendmail"/>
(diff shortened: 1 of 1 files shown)


xml/smtp_banners.xml
```diff
@@ -69,28 +69,43 @@ The system or service fingerprint with the highest certainty overwrites the othe
          http://www.argosoft.com/applications/mailserver/
          Example: 220 ArGoSoft Mail Server, Version 1.4 (1.4.0.3)
       </description>
+    <param pos="0" name="os.vendor" value="Microsoft"/>
+    <param pos="0" name="os.family" value="Windows"/>
+    <param pos="0" name="os.device" value="General"/>
```
jhart-r7: We should not be adding the "General" fingerprint (here and elsewhere)

xml/smtp_banners.xml
```diff
@@ -1690,20 +1871,21 @@ The system or service fingerprint with the highest certainty overwrites the othe
...
     <example host.name="example.com">example.com ESMTP Perl</example>
     <param pos="0" name="service.product" value="Perl"/>
     <param pos="1" name="host.name"/>
   </fingerprint>
-  <fingerprint pattern="^([^ ]+) E?SMTP(?: (?:Service )?Ready\.?)?$" flags="REG_ICASE">
-    <description>
-        catch all for daemons that have no distinguishing fingerprint whatsoever
-      </description>
+  <fingerprint pattern="^(?i)(?:([^ ]+) )?E?SMTP(?: (?:Service )?Ready\.?)?$">
+    <description>catch all for daemons that have no distinguishing fingerprint whatsoever</description>
```
pdeardorff-r7: I remember there were some concerns for catch-alls.  Specifically around ordering and whether we should include them or not.  @jhart-r7 might have more info here...

xml/smtp_banners.xml
```diff
@@ -1690,20 +1871,21 @@ The system or service fingerprint with the highest certainty overwrites the othe
     <param pos="3" name="service.version.version"/>
     <param pos="4" name="system.time"/>
   </fingerprint>
-  <fingerprint pattern="^(\S+) E?SMTP Perl" flags="REG_ICASE">
+  <fingerprint pattern="^(?i)(\S+) E?SMTP Perl">
```
jhart-r7: `(?i)(\S+)` is a little odd.  Just make it `(\S+).

xml/smtp_banners.xml
tsellers-r7: This catch all isn't new, I've just made the hostname optional for it so that it will matched even shorter non-specific names such as 'ESMTP ready'

xml/smtp_banners.xml
tsellers-r7: I can't find this banner in our dataset so I set the whole regex to be case insensitive.

xml/smtp_banners.xml
jhart-r7: This may be labeled as a catch all but the regular expression is not.  It looks like it is intended to only match things that have an optional hostname followed by the string "ESMTP Service Ready" and a few minor variants.  A catch-all would do something like match "ESMTP Service Ready" with arbitrary text before and after.

xml/smtp_banners.xml
```diff
@@ -794,30 +829,47 @@ The system or service fingerprint with the highest certainty overwrites the othe
...
     <param pos="0" name="system.time.format" value="EEE, dd MMM yyyy HH:mm:ss zzz"/>
     <param pos="1" name="host.name"/>
     <param pos="2" name="service.version"/>
     <param pos="3" name="system.time"/>
   </fingerprint>
+  <fingerprint pattern="^(?:([^ ]+))? *ESMTP Service \(IBM Domino Release (\d+\.[\w.]+(?: HF\d+)?)\) ready at (.+) *$">
+    <description>IBM Domino SMTP MTA</description>
+    <example host.name="foo.bar" service.version="9.0.1FP8 HF475">foo.bar ESMTP Service (IBM Domino Release 9.0.1FP8 HF475) ready at Thu, 30 Nov 2017 17:55:48 +0900</example>
+    <example host.name="foo.bar" service.version="9.0.1">foo.bar ESMTP Service (IBM Domino Release 9.0.1) ready at Thu, 30 Nov 2017 10:12:26 +0100</example>
+    <example service.version="9.0.1FP8"> ESMTP Service (IBM Domino Release 9.0.1FP8) ready at Thu, 30 Nov 2017 13:51:59 -0800</example>
```
jhart-r7: what is `host.name` in this case?  You made that outer group non-capturing and optional, but the inner will capture.  I wonder if it is just set to the empty string?  Do we care?  I think the only solution would be to split these fingerprints up, one with a hostname and one with just spaces.

tsellers-r7: I went this route to reduce the number of fingerprints.  The result is an empty capture (at least in Ruby and (according to an online tester) Java. I'm ok with splitting them if you think it will make it easier to support or more cross-language compatible.

jhart-r7: I guess my concern is that the behavior of an empty field in the "fingerprint" that we returns is undefined.  It works one way when you use recog as a data source of fingerprints and an api/cli for interacting with them, it may work another way when used in something else, like metasploit-framework or other products.  We don't have a good way to control this today, AFAIK.

jhart-r7: We discussed this on the phone.  Good to go.

PR #162, opened 2018-04-21
Small regexp fixes for Go compatibility

hdm:
This change corrects two typos and removes two negative look aheads, allowing these expressions to be parsed by the Go standard library 'regexp' engine.

### DIFF
--- a/xml/dns_versionbind.xml
+++ b/xml/dns_versionbind.xml
@@ -304,7 +304,7 @@
     <param pos="0" name="service.family" value="PowerDNS"/>
     <param pos="0" name="service.product" value="Recursor"/>
   </fingerprint>
-  <fingerprint pattern="^PowerDNS Authoritative Server (\d\.[\d.]++(?:-rc\d)?) \(\w+@[\w.]+ built [\d\s]+\w*@[\w.-]*\)$">
+  <fingerprint pattern="^PowerDNS Authoritative Server (\d\.[\d\.]+(?:-rc\d)?) \(\w+@[\w.]+ built [\d\s]+\w*@[\w.-]*\)$">
     <description>PowerDNS Authoritative Server</description>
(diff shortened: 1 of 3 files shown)


xml/dns_versionbind.xml
```diff
@@ -304,7 +304,7 @@
     <param pos="0" name="service.family" value="PowerDNS"/>
     <param pos="0" name="service.product" value="Recursor"/>
   </fingerprint>
-  <fingerprint pattern="^PowerDNS Authoritative Server (\d\.[\d.]++(?:-rc\d)?) \(\w+@[\w.]+ built [\d\s]+\w*@[\w.-]*\)$">
+  <fingerprint pattern="^PowerDNS Authoritative Server (\d\.[\d\.]+(?:-rc\d)?) \(\w+@[\w.]+ built [\d\s]+\w*@[\w.-]*\)$">
```
tsellers-r7: This character is valid in the context of this character class group, similar to the one at the end of the pattern.  I've tested this both ways with an online regex tester for Golang but I haven't tested it natively with Golang. Unless you have a concern I'll revert this specific change.

hdm: Oh, the "." in the end of the pattern has the same issue, it matches everything, not just period.

tsellers-r7: It should not, see my comment just above your last comment.  I updated it to be more specific and to provide a link.  This probably didn't send an updated notice.

tsellers-r7:
Landed, thanks much HD.

xml/dns_versionbind.xml
hdm: Thank you, that was news to me.

hdm:
Awesome! Thank you for landing it. I should have a recog-go package available in the next few weeks (it works now, but is _REALLY_ ugly).

tsellers-r7:
Sounds great!  Also, your PR made me aware of the Golang lookahead/lookbehind limitations (in the native lib) which likely saved me tons of time on a project I was going to use to learn Golang so thanks for that too.

PR #163, opened 2018-06-25
Amazon Linux AMI fingerprinting via ftp banner

ncameron-r7:
Comes from an FTP banner seen on shodan here is the full output (with response codes)
`
220-Amazon Linux AMI release 2016.09
220-Kernel \r on an \m
220-
220
530 Permission denied.
530 Please login with USER and PASS.
211-Features:
 EPRT
 EPSV
 MDTM
 PASV
 REST STREAM
 SIZE
 TVFS
 UTF8
211 End`
nexpose PR - https://github.com/rapid7/nexpose/pull/5573

### DIFF
--- a/xml/ftp_banners.xml
+++ b/xml/ftp_banners.xml
@@ -1317,7 +1317,7 @@ more text</example>
     <param pos="2" name="service.version"/>
     <param pos="1" name="host.name"/>
   </fingerprint>
-  <fingerprint pattern="^Amazon Linux AMI release (.*)">
+  <fingerprint pattern="Amazon\sLinux\sAMI\srelease\s(\d+\.\d+)">
     <description>Amazon Linux AMI</description>
     <example os.version="2016.09">Amazon Linux AMI release 2016.09</example>
     <param pos="0" name="os.vendor" value="Amazon"/>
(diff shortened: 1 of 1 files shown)


xml/ftp_banners.xml
```diff
@@ -1317,4 +1317,12 @@ more text</example>
     <param pos="2" name="service.version"/>
     <param pos="1" name="host.name"/>
   </fingerprint>
+  <fingerprint pattern="^Amazon Linux AMI release (.*)">
```
kkuczkowski-r7: use `\s` to match white space?

cmccrisken-r7: Instead of match `(.*)`, I would match:
`^Amazon\sLinux\sAMI\srelease\s(\d+\.\d+)%`

cfleming-R7: @cmccrisken-r7 I assume you mean `^Amazon\sLinux\sAMI\srelease\s(\d+\.\d+)$`? With `$` rather than `%`?

ncameron-r7: Since the intention here is to match single whitespace characters then just a space in the pattern is more readable with no impact on matching. It also is more consistent with the rest of the patterns contained within the XML. But i'm open to either way

kkuczkowski-r7: :+1:

PR #164, opened 2018-06-26
Adding maria db max scale banner support

sdynes-r7:
### Description
Adds fingerprinting support for MariaDB MaxScale 

### Motivation and context
MariaDB Maxscale is a popular database proxy and adding it will increase coverage of the MariaDB line of products

### How Has This Been Tested?

- [x] Run rspec on branch
- [x] Run `bin/recog_verify xml/mysql_banners.xml`:

### DIFF
--- a/xml/mysql_banners.xml
+++ b/xml/mysql_banners.xml
@@ -998,13 +998,13 @@
     <param pos="0" name="service.product" value="MariaDB"/>
     <param pos="0" name="service.edition" value="Enterprise Edition"/>
   </fingerprint>
-  <fingerprint pattern="^(?:\d{1,2}\.\d{1,2}\.[\d]{1,2})(?:-\d{1,2}\.\d{1,2}\.[\d]{1,2})?\s(\d{1,2}\.\d{1,2}\.\d{1,2})-maxscale$">
+  <fingerprint pattern="^(?:\d{1,2}\.\d{1,2}\.[a-f\d]{1,2})(?:-\d{1,2}\.\d{1,2}\.[a-f\d]{1,2})?\s(\d{1,2}\.\d{1,2}\.\d{1,2})-maxscale$">
     <description>MariaDB MaxScale</description>
     <example service.version="2.1.14">5.5.5-10.0.0 2.1.14-maxscale</example> 
     <param pos="1" name="service.version"/>
     <param pos="0" name="service.vendor" value="MariaDB"/>
     <param pos="0" name="service.family" value="MySQL"/>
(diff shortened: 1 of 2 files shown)


xml/mysql_banners.xml
```diff
@@ -998,6 +998,15 @@
...
     <param pos="0" name="service.edition" value="Enterprise Edition"/>
   </fingerprint>
+  <fingerprint pattern="(\d{1,2}\.\d{1,2}\.\d{1,2})-maxscale">
+    <description>MariaDB MaxScale</description>
+    <example service.version="2.0.3">5.5.5-10.0.0 2.0.3-maxscale</example>
+    <example service.version="2.1.14">5.5.5-10.0.0 2.1.14-maxscale</example> 
+    <param pos="1" name="service.version"/>
+    <param pos="0" name="service.vendor" value="MariaDB"/>
+    <param pos="0" name="service.family" value="MySQL"/>
+    <param pos="0" name="service.product" value="MariaDB Maxscale"/>
```
ekelly-rapid7: MariaDB is redundant in the product field.

xml/mysql_banners.xml
```diff
@@ -998,6 +998,15 @@
     <param pos="0" name="service.product" value="MariaDB"/>
     <param pos="0" name="service.edition" value="Enterprise Edition"/>
   </fingerprint>
+  <fingerprint pattern="(\d{1,2}\.\d{1,2}\.\d{1,2})-maxscale">
+    <description>MariaDB MaxScale</description>
+    <example service.version="2.0.3">5.5.5-10.0.0 2.0.3-maxscale</example>
+    <example service.version="2.1.14">5.5.5-10.0.0 2.1.14-maxscale</example>
```
ekelly-rapid7: Are two examples really necessary? They exercise the same regex so not sure it's really necessary.

xml/mysql_banners.xml
```diff
@@ -998,6 +998,15 @@
     <param pos="0" name="service.product" value="MariaDB"/>
     <param pos="0" name="service.edition" value="Enterprise Edition"/>
   </fingerprint>
+  <fingerprint pattern="(\d{1,2}\.\d{1,2}\.\d{1,2})-maxscale">
```
ekelly-rapid7: We should consider adding the regex for the initial versions and anchoring on the full field if we know it is always in the format.

xml/mysql_banners.xml
amcclenaghan-r7: Scale needs to be capitalised as well 'MaxScale'

PR #168, opened 2018-08-17
Address several formatting and consistency issues

jhart-r7:
Each of the commits in this branch represent something that I ran into recently while working with the recog content.


xml/sip_user_agents.xml
```diff
@@ -1,8 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!--
-SIP User Agent header values are matched against these patterns to fingerprint SIP devices.
--->
 <fingerprints matches="sip_header.user_agent" protocol="sip" database_type="service">
+  <!--
+  SIP User Agent header values are matched against these patterns to fingerprint SIP devices.
+  -->
+
```
tsellers-r7: Should the extra line here and elsewhere be removed? It's not many other fps.

xml/ntp_banners.xml
```diff
@@ -599,137 +599,137 @@ NTP "banners", taken from a readvar response
...
   </fingerprint>
   <fingerprint pattern="^.*processor=&quot;([^ ]+)&quot;,.*system=&quot;SunOS/?5.([789]|10)&quot;" flags="REG_DOT_NEWLINE,REG_ICASE">
-   <description>Solaris 7-10 (SunOS/5.7 - SunOS/5.10) with no ntp version</description>
-   <example os.arch="sun4m" os.version="7">
+    <description>Solaris 7-10 (SunOS/5.7 - SunOS/5.10) with no ntp version</description>
+    <example os.arch="sun4m" os.version="7">
      processor="sun4m", system="SunOS/5.7"
    </example>
-   <example os.arch="sun4m" os.version="10">
+    <example os.arch="sun4m" os.version="10">
```
tsellers-r7: Indention here is too deep vs closing tag.  Should it just be moved to a single line for opening/closing tags + content like fingerprint examples in other files?

jhart-r7: It seems like the regex is meant to work over multiple lines, so I think we should just keep it as is.  The whitespace matters here, but I don't know why xmllint isn't handling the indention as we'd like.  For now, I say we just leave it as is.

tsellers-r7:
LGTM

PR #171, opened 2018-08-24
Correct F5 fingerprints

jhart-r7:
In other fingerprints the vendor is F5, which is more correct.  This makes the vendor name consistent.

This also uses the BIG-IP family and BIG-IP LTM product consistently, and corrects a handful of `os` fingerprints that should be `service`.

### DIFF
--- a/xml/http_wwwauth.xml
+++ b/xml/http_wwwauth.xml
@@ -287,8 +287,9 @@
   <fingerprint pattern="^.*(?:Basic|Digest) .*realm=&quot;BIG-IP&quot;.*$">
     <description>Generic F5 Big-IP</description>
     <example>Basic realm="BIG-IP"</example>
-    <param pos="0" name="os.vendor" value="F5"/>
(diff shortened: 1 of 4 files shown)


xml/http_wwwauth.xml
```diff
@@ -287,8 +287,9 @@
   <fingerprint pattern="^.*(?:Basic|Digest) .*realm=&quot;BIG-IP&quot;.*$">
     <description>Generic F5 Big-IP</description>
     <example>Basic realm="BIG-IP"</example>
-    <param pos="0" name="os.vendor" value="F5"/>
-    <param pos="0" name="os.product" value="BIG-IP"/>
+    <param pos="0" name="servicevendor" value="F5"/>
```
gwiseman-r7: missing `.` separator

jhart-r7:
Good catch.  I'll see about adding a test to catch possibly bad fingerprints like that.

jhart-r7:
This was landed, but for some reason it is not showing up here.

PR #172, opened 2018-08-24
Add CPEs to recog fingerprints

jhart-r7:
This is a proposal for adding CPEs to recog along with code to keep these values maintained.  The CPEs added are based off of `service`, `os`, and `hw` fingerprints in recog that have a vendor and product that map to a known valid combo in the CPE 2.3 dictionary.  

When present, a version will also be included in the CPE value:

```
    <param pos="0" name="service.cpe23" value="cpe:/a:vmware:zimbra_desktop:1"/>
```

In cases where the original fingerprint would use the interpolated value for the version, this CPE functionality follows the same convention, requiring consumers to interpolate the correct version when emitting the CPE fingerprints, for example:

```
    <param pos="0" name="service.cpe23" value="cpe:/a:vmware:zimbra_desktop:{service.version}"/>
```

The CPEs are 2.3 URIs, which is backwards compatible with 2.2.

This doesn't add CPEs to everything.  There are many factors that contribute to this:

  * Not every vendor or product name used in Recog is valid in the CPE 2.3 dictionary.  The maintenance code does support manual remapping of vendors and products.  
  * Similarly, not every vendor or product might be valid or applicable for the given fingerprint type.  For example, Microsoft Exchange is a valid vendor and product for a `service`, but not an `os`.  
  * ...

I don't believe this should have any impact on current consumers because this is just another `param` element with a new name.  

There are two parts to this PR:
  * Updates to the recog XML adding CPEs where possible
  * Python code to maintain the CPE values in recog XML, including a JSON file to support remapping between Rapid7-isms and CPE.  Note that the maintenance code is written in python 3 for lack of a ruby XML library that did what I needed, plus I just wanted to do this in python.  

Feel free to review both, however it may be better to examine the CPEs added to the recog XML first.

For using the updating code, you'll need to fetch the [CPE 2.3 XML dictionary](https://nvd.nist.gov/feeds/xml/cpe/dictionary/official-cpe-dictionary_v2.3.xml.gz):

```
curl -s  https://nvd.nist.gov/feeds/xml/cpe/dictionary/official-cpe-dictionary_v2.3.xml.gz | pigz -d > official-cpe-dictionary_v2.3.xml
```

Now run the update code against every XML file in recog, providing as arguments the path to the recog XML file, the path to the CPE 2.3 XML dictionary and the path to a JSON file that maps Rapid7 vendor and product names to their CPE equivalents.  I expedite this with `parallel` and save the error output for further remapping or recog bug fixing:

```
ls xml/*.xml | parallel --gnu "./update_cpes.py {} official-cpe-dictionary_v2.3.xml  remap.json && xmllint --format --noblanks {} > {}.bak &&  mv {}.bak {} || echo {}" 2> errors.txt
```


update_cpes.py
```diff
@@ -0,0 +1,201 @@
...
+            vendor = fp_data['vendor']
+            product = fp_data['product']
+            version = fp_data['version']
+
+            # build a reasonable looking CPE value from the vendor/product/version,
+            # lowercasing, replacing whitespace with _, and more
+            if vendor and product:
+                if not cpe_type in cpe_vp_map:
+                    print(cpe_type)
+                    logging.error("Didn't find CPE type %s", cpe_type)
```
tsellers-r7: Adding the vendor/product to this output would provide useful context.
For example:
#### Existing:
```
a
ERROR:root:Didn't find CPE type a
```
#### Possible output 
using `logging.error("Didn't find CPE type '%s' for '%s %s'", cpe_type, vendor, product)`
```
a
ERROR:root:Didn't find CPE type a for Oracle XML DB
```

update_cpes.py
```diff
@@ -0,0 +1,201 @@
...
+            cpe_type, vendor, product = cpe_match.group(1, 2, 3)
+            if not cpe_type in vp_map:
+                vp_map[cpe_type] = {}
+            if not vendor in vp_map[cpe_type]:
+                vp_map[cpe_type][vendor] = set()
+            vp_map[cpe_type][vendor].add(product)
+        else:
+            logging.error("Unexpected CPE %s", final_cpe_name)
+
+    return vp_map
```
tsellers-r7: May want to add a sanity check here.  If no records are imported from the XML then log an error and exit.  Ran into this because it seems that I don't have the correct XML file.

PR #173, opened 2018-08-29
Enforce <= 20 examples per fingerprint; cleanup

jhart-r7:
The purpose of examples is not to list every possible variant.  I added the unit test to enforce this, and then went through and chopped out large swaths of examples that didn't look to be providing any extra testing value and therefore just waste space, time and $ when running these tests.


tsellers-r7:
Jenkins test this

PR #175, opened 2018-08-31
Banner update: HTTP - 2018.08.13

tsellers-r7:
This PR is a banner update of the `http_servers.xml` file using data from a Project Sonar survey of HTTP on port 80 that occurred on 2018.08.13.

There are additional tweaks to support CPE generation as well as add `hw.device` and similar fields where possible.

*Note:* I am unsure as to why the CPE ordering changed.

### DIFF
--- a/xml/http_servers.xml
+++ b/xml/http_servers.xml
@@ -10,20 +10,20 @@
     <example>Stronghold/4.0 Apache/1.3.22 (Unix) mod_ssl/2.8.7 OpenSSL/0.9.6c mod_perl/1.26</example>
     <param pos="0" name="service.vendor" value="Apache"/>
     <param pos="0" name="service.product" value="HTTPD"/>
+    <param pos="0" name="service.cpe23" value="cpe:/a:apache:http_server:{service.version}"/>
     <param pos="0" name="service.family" value="Apache"/>
     <param pos="2" name="service.version"/>
-    <param pos="0" name="service.cpe23" value="cpe:/a:apache:http_server:{service.version}"/>
     <param pos="0" name="service.component.vendor" value="Red Hat"/>
+    <param pos="0" name="service.component.cpe23" value="cpe:/a:redhat:stronghold:{service.component.version}"/>
     <param pos="0" name="service.component.product" value="Stronghold"/>
     <param pos="0" name="service.component.family" value="Stronghold"/>
     <param pos="1" name="service.component.version"/>
-    <param pos="0" name="service.component.cpe23" value="cpe:/a:redhat:stronghold:{service.component.version}"/>
     <param pos="3" name="apache.info"/>
     <param pos="0" name="os.vendor" value="Red Hat"/>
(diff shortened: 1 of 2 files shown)


xml/http_servers.xml
```diff
@@ -1475,9 +1475,14 @@
     <param pos="0" name="service.family" value="Firewall-1"/>
     <param pos="0" name="service.cpe23" value="cpe:/a:checkpoint:firewall-1:-"/>
     <param pos="0" name="os.vendor" value="Check Point"/>
+    <param pos="0" name="os.cpe23" value="cpe:/o:checkpoint:gaia_os:-"/>
     <param pos="0" name="os.device" value="Firewall"/>
     <param pos="0" name="os.family" value="Firewall-1"/>
-    <param pos="0" name="os.product" value="Firewall-1"/>
+    <param pos="0" name="os.product" value="GAiA OS"/>
```
jhart-r7: Had to look this up to confirm.  👍

xml/http_servers.xml
```diff
@@ -1993,6 +1998,101 @@
...
+    <description>LiteSpeed</description>
+    <example>LiteSpeed</example>
+    <example>LiteSpeed/5.2.8 Enterprise</example>
+    <param pos="0" name="service.vendor" value="LiteSpeed Technologies"/>
+    <param pos="0" name="service.product" value="LiteSpeed Web Server"/>
+    <param pos="1" name="service.version"/>
+  </fingerprint>
+  <fingerprint pattern="^IdeaWebServer\/v?([\d.]+)$">
+    <description>Idea Web Server</description>
+    <example>IdeaWebServer/0.83</example>
```
jhart-r7: Minor, no example, not much different from what is already there.

xml/http_servers.xml
```diff
@@ -2380,6 +2475,18 @@
     <param pos="0" name="service.vendor" value="nginx"/>
     <param pos="0" name="service.cpe23" value="cpe:/a:nginx:nginx:-"/>
   </fingerprint>
+  <fingerprint pattern="^nginx\/?(:?[\d.]+)?">
+    <description>nginx with version info and/or mods</description>
+    <example service.version="0.8.53">nginx/0.8.53 + Phusion Passenger 3.0.0 (mod_rails/mod_rack)</example>
+    <example>nginx/0.8.53</example>
+    <example>nginx + Phusion Passenger 5.1.11</example>
```
jhart-r7: Strange.  So we get an empty string for the version here?  May be worth two fingerprints, one with a version, one without.

xml/http_servers.xml
tsellers-r7: Not sure I understand this comment.  Did you mean that the following two examples aren't different enough to warrant needing them?

xml/http_servers.xml
tsellers-r7: We have quite a few other fingerprints that can have an empty capture if the value isn't there.  If you like I can split this one up.

PR #178, opened 2018-09-11
TELNET: Initial commit

tsellers-r7:
Initial commit of TELNET fingerprint coverage.  This should be viewed with a very critical eye.

Processes leveraging the fingerprints *should* remove TELNET options control codes as well as strip CR, LF, and other whitespace from both ends of the banner.

A couple notes. 
1. Regex for leading CR/LF characters and trailing space characters has been left in place to enable a current project.  This will be removed the near future.

2. TELNET banners most often contain CR and LF. In order to properly create testable examples I've base64 encoded the banners.  The pattern I am using is a comment with the original banner containing escaped CR/LF followed by the encoded banner.

```xml
    <!-- MikroTik v5.2\r\nLogin: -->
    <example _encoding="base64" os.version="5.2">TWlrcm9UaWsgdjUuMg0KTG9naW46Cg==</example>
```

### DIFF
--- a/xml/telnet_banners.xml
+++ b/xml/telnet_banners.xml
@@ -148,14 +148,12 @@
   <fingerprint pattern="^(?:\r|\n)*ZyXEL login:$">
     <description>ZyXEL simple</description>
     <example>ZyXEL login:</example>
-    <param pos="0" name="service.vendor" value="ZyXEL"/>
     <param pos="0" name="hw.vendor" value="ZyXEL"/>
   </fingerprint>
   <fingerprint pattern="^ZyXEL \w?DSL Router\r\nLogin:$">
     <description>ZyXEL Router - simple</description>
     <!-- ZyXEL VDSL Router\r\nLogin: -->
     <example _encoding="base64">WnlYRUwgVkRTTCBSb3V0ZXINCkxvZ2luOgo=</example>
-    <param pos="0" name="service.vendor" value="ZyXEL"/>
     <param pos="0" name="hw.vendor" value="ZyXEL"/>
     <param pos="0" name="hw.device" value="Router"/>
   </fingerprint>
@@ -619,7 +617,6 @@
   <fingerprint pattern="^192.0.0.64 login:\s*$">
     <description>Hikvision cameras and NVRs (multiple)</description>
     <example>192.0.0.64 login:</example>
-    <param pos="0" name="service.vendor" value="Hikvision"/>
     <param pos="0" name="os.vendor" value="Hikvision"/>
     <param pos="0" name="hw.vendor" value="Hikvision"/>
   </fingerprint>
--- a/xml/telnet_banners.xml
+++ b/xml/telnet_banners.xml
@@ -248,11 +248,11 @@
     <example _encoding="base64" os.version="17.04" host.name="nginx">
       VWJ1bnR1IDE3LjA0DQpuZ2lueCBsb2dpbjoK
     </example>
-    <param pos="0" name="os.vendor" value="Debian"/>
+    <param pos="0" name="os.vendor" value="Ubuntu"/>
     <param pos="0" name="os.family" value="Linux"/>
-    <param pos="0" name="os.product" value="Linux"/>
+    <param pos="0" name="os.product" value="Ubuntu Linux"/>
     <param pos="1" name="os.version"/>
-    <param pos="0" name="os.cpe23" value="cpe:/o:debian:debian_linux:{os.version}"/>
+    <param pos="0" name="os.cpe23" value="cpe:/o:canonical:ubuntu_linux:{os.version}"/>
     <param pos="2" name="host.name"/>
   </fingerprint>
   <fingerprint pattern="(?:\r|\n)*Debian GNU\/Linux (2.\d)(?: [\w.-]+)?(?:\r|\n)+([\w.-]+) login:\s*">
--- a/xml/telnet_banners.xml
+++ b/xml/telnet_banners.xml
@@ -14,7 +14,7 @@
   Due to the multi-line nature of TELNET banners the regex are leveraging \A
   instead of ^ to prevent matching in the beginning of a 'line' (^) instead of
   at the beginning of the string (\A). This has been verified to work with
-  Ruby, Python, Javascript, and Golang.
+  Ruby, Python, Java, and Golang.
   -->
   <fingerprint pattern="\A(?i)(?:\r|\n)*login:\s*$">
     <description>bare 'login:' -- assert nothing.</description>
--- a/xml/telnet_banners.xml
+++ b/xml/telnet_banners.xml
@@ -15,8 +15,6 @@
     <example _encoding="base64" os.version="6.40.8">TWlrcm9UaWsgdjYuNDAuOCAoYnVnZml4KQ0KTG9naW46Cg==</example>
     <!-- MikroTik v6.36rc12 (testing)\r\nLogin: -->
     <example _encoding="base64" os.version="6.36rc12">TWlrcm9UaWsgdjYuMzZyYzEyICh0ZXN0aW5nKQ0KTG9naW46Cg==</example>
-    <param pos="0" name="service.vendor" value="MikroTik"/>
-    <param pos="0" name="service.product" value="telnet"/>
     <param pos="0" name="os.vendor" value="MikroTik"/>
     <param pos="0" name="os.device" value="Router"/>
     <param pos="0" name="os.product" value="RouterOS"/>
@@ -30,11 +28,9 @@
     <!-- ZXHN H108N\r\nLogin: -->
     <example _encoding="base64" hw.product="H108N">WlhITiBIMTA4Tg0KTG9naW46Cg==</example>
     <!-- ZXHN H298A V1.1\r\nLogin: -->
-    <example _encoding="base64"  hw.product="H298A" hw.version="1.1">WlhITiBIMjk4QSBWMS4xDQpMb2dpbjoK</example>
+    <example _encoding="base64" hw.product="H298A" hw.version="1.1">WlhITiBIMjk4QSBWMS4xDQpMb2dpbjoK</example>
     <!-- ZXHN H367N\r\n\rLogin: -->
     <example _encoding="base64" hw.product="H367N">WlhITiBIMzY3Tg0KDUxvZ2luOgo=</example>
-    <param pos="0" name="service.vendor" value="ZTE"/>
-    <param pos="0" name="service.product" value="telnet"/>
     <param pos="0" name="hw.vendor" value="ZTE"/>
     <param pos="0" name="hw.device" value="Router"/>
     <param pos="0" name="hw.family" value="ZXHN"/>
@@ -47,8 +43,6 @@
     <example _encoding="base64" hw.product="F668">RjY2OA0KDUxvZ2luOgo=</example>
     <!-- F612W\r\n\rLogin: -->
     <example _encoding="base64" hw.product="F612W">RjYxMlcNCg1Mb2dpbjoK</example>
-    <param pos="0" name="service.vendor" value="ZTE"/>
-    <param pos="0" name="service.product" value="telnet"/>
     <param pos="0" name="hw.vendor" value="ZTE"/>
     <param pos="0" name="hw.device" value="Router"/>
     <param pos="1" name="hw.product"/>
@@ -66,8 +60,6 @@
       DgvMDcvMTAgKFNWTiByZXZpc2lvbjogMTQ4OTYpDQoNClByb2xpYW50IERMOTgwUjA3IFg2NT
       UwIDgtY29yZSA0UCBTQVMgbG9naW46Cg==
     </example>
-    <param pos="0" name="service.vendor" value="DD-WRT"/>
-    <param pos="0" name="service.product" value="telnet"/>
     <param pos="0" name="os.vendor" value="DD-WRT"/>
     <param pos="0" name="os.product" value="DD-WRT"/>
     <param pos="0" name="os.device" value="Router"/>
@@ -83,8 +75,6 @@
       REQtV1JUIHYzLjAtcjM0ODg2TSBzdGQgKGMpIDIwMTggTmV3TWVkaWEtTkVUIEdtYkgNClJlb
       GVhc2U6IDAyLzEwLzE4DQoNCndpYnJhdGUgbG9naW46Cg==
     </example>
-    <param pos="0" name="service.vendor" value="DD-WRT"/>
-    <param pos="0" name="service.product" value="telnet"/>
     <param pos="0" name="os.vendor" value="DD-WRT"/>
     <param pos="0" name="os.product" value="DD-WRT"/>
     <param pos="0" name="os.device" value="Router"/>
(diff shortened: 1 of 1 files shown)


xml/telnet_banners.xml
```diff
@@ -0,0 +1,648 @@
...
+    <description>SMC Cable Modem</description>
+    <example>SMCD3GN2-BIZ login:</example>
+    <param pos="0" name="service.vendor" value="SMC Networks"/>
+    <param pos="0" name="service.product" value="telnet"/>
+    <param pos="0" name="hw.vendor" value="SMC Networks"/>
+    <param pos="0" name="hw.device" value="Cable Modem"/>
+    <param pos="1" name="hw.product"/>
+  </fingerprint>
+  <fingerprint pattern="^Welcome to Microsoft Telnet Service \r\n\n\rlogin:\s*$">
+    <description>Microsoft Windows</description>
```
jhart-r7: CE

xml/telnet_banners.xml
```diff
@@ -0,0 +1,648 @@
...
+    <param pos="1" name="hw.product"/>
+  </fingerprint>
+  <fingerprint pattern="^(BCM\d+) Broadband Router\r\nTelnet is Disabled in WAN Side$">
+    <description>OEM'd Broadcom Router</description>
+    <!-- BCM963268 Broadband Router\r\nTelnet is Disabled in WAN Side -->
+    <example _encoding="base64" hw.product="BCM963268">QkNNOTYzMjY4IEJyb2FkYmFuZCBSb3V0ZXINClRlbG5ldCBpcyBEaXNhYmxlZCBpbiBXQU4gU2lkZQo=</example>
+    <param pos="0" name="hw.device" value="Router"/>
+    <param pos="1" name="hw.product"/>
+  </fingerprint>
+  <fingerprint pattern="^(?:\r|\n)*login:\s*$" flags="REG_ICASE">
```
jhart-r7: IIRC we stopped using `flags` where possible, because these constants are language specific and complicate integrations just using recog content.  Instead we've been using case-insensitive regular expression modifiers (`(?i:blah)`)

xml/telnet_banners.xml
```diff
@@ -0,0 +1,648 @@
...
+    </example>
+    <!-- TiMOS-B-9.0.R9 both/mpc Nokia SAS-M 24F 2XFP 7210 Copyright (c) 2000-2017 Nokia.\r\r\nBanner Shortened For \r\r\nBrevity\r\nLogin: -->
+    <example _encoding="base64" hw.product="SAS-M 24F 2XFP 7210">
+      VGlNT1MtQi05LjAuUjkgYm90aC9tcGMgTm9raWEgU0FTLU0gMjRGIDJYRlAgNzIxMCBDb3B5c
+      mlnaHQgKGMpIDIwMDAtMjAxNyBOb2tpYS4NDQpCYW5uZXIgU2hvcnRlbmVkIEZvciANDQpCcm
+      V2aXR5DQpMb2dpbjoK
+    </example>
+    <param pos="0" name="os.vendor" value="Nokia"/>
+    <param pos="0" name="os.product" value="TimOS"/>
+    <param pos="0" name="os.device" value="Router"/>
```
jhart-r7: It says router here but switch for hw.

xml/telnet_banners.xml
```diff
@@ -0,0 +1,648 @@
...
+    <example hw.product="Air5650TT">Air5650TT login:</example>
+    <param pos="0" name="service.product" value="telnet"/>
+    <param pos="0" name="hw.vendor" value="Airties"/>
+    <param pos="0" name="hw.device" value="WAP"/>
+    <param pos="1" name="hw.product"/>
+  </fingerprint>
+  <fingerprint pattern="^Amazon Linux AMI release ([\d.]+)(?:\r|\n)+Kernel ([\w.-]+) on an (\w+)(?:\r|\n)+(?:([\w.-]+) )?login:\s*$">
+    <description>Amazon Linux AMI</description>
+    <!-- Amazon Linux AMI release 2013.09\r\nKernel 3.4.68-59.97.amzn1.x86_64 on an x86_64\r\nserver-01.2 login: -->
+    <example _encoding="base64" os.version="2013.09" linux.kernel.version="3.4.68-59.97.amzn1.x86_64">
```
jhart-r7: minor, but if possible any examples should test every extraction.  In this case `os.arch` and `host.name` need testing.  Here and maybe elsewhere.

xml/telnet_banners.xml
```diff
@@ -0,0 +1,648 @@
...
+    <param pos="0" name="os.vendor" value="Hikvision"/>
+    <param pos="0" name="hw.vendor" value="Hikvision"/>
+  </fingerprint>
+  <fingerprint pattern="^Remote Management Console\r\nlogin:\s*$">
+    <description>Juniper Netscreen</description>
+    <!-- Remote Management Console\r\nlogin: -->
+    <example _encoding="base64">UmVtb3RlIE1hbmFnZW1lbnQgQ29uc29sZQ0KbG9naW46Cg==</example>
+    <param pos="0" name="service.vendor" value="Juniper"/>
+    <param pos="0" name="service.family" value="NetScreen"/>
+    <param pos="0" name="service.product" value="NetScreen"/>
```
jhart-r7: why not telnet?  most others seem to use `telnet` except for things like jetdirect, which makes sense.

xml/telnet_banners.xml
tsellers-r7: Ah, thanks for reminding me!

xml/telnet_banners.xml
tsellers-r7: Doh, actually, I think the CE reference is incorrect. Fixed

xml/telnet_banners.xml
```diff
@@ -0,0 +1,735 @@
...
+    <param pos="2" name="hw.version"/>
+  </fingerprint>
+  <fingerprint pattern="^(F6\d+\w?)\r\n\rLogin:\s*$">
+    <description>ZTE F6xx series GPON router</description>
+    <!-- F668\r\n\rLogin: -->
+    <example _encoding="base64" hw.product="F668">RjY2OA0KDUxvZ2luOgo=</example>
+    <!-- F612W\r\n\rLogin: -->
+    <example _encoding="base64" hw.product="F612W">RjYxMlcNCg1Mb2dpbjoK</example>
+    <param pos="0" name="service.vendor" value="ZTE"/>
+    <param pos="0" name="service.product" value="telnet"/>
```
jhart-r7: I'm not sure we should be using `telnet` as the `product` here.  Yes, the service is telnet but the rest of the `service` fingerprint implies that ZTE wrote this telnet product, when in reality they (or others) are just using some open source/linux/GNU telnetd implementation.   

Looked at from another perspective, to even perform the recog fingerprinting, one must have already identified that the protocol is telnet and spoken it to obtain the banner.

tsellers-r7: Agreed, I'd been on the fence on this and was likely going to remove it.  We already get back  `"service.protocol":"telnet"` automatically for matches in this file where that value isn't overridden.

tsellers-r7:
I think this PR is ready for final review.  There are still an astounding number of telnet banners in our datasets but this should lay the groundwork for reviewing and adding them when resources permit.

xml/telnet_banners.xml
```diff
@@ -0,0 +1,820 @@
...
+  <!--
+  The following 'assert nothing' block is intended to handle banners so simple
+  that they cannot be attributed to a product or vendor. They are at the
+  beginning of the file as a performance tweak given how frequenty they occur.
+
+  NOTE:
+  Due to the multi-line nature of TELNET banners the regex are leveraging \A
+  instead of ^ to prevent matching in the beginning of a 'line' (^) instead of
+  at the beginning of the string (\A). This has been verified to work with
+  Ruby, Python, Java, and Golang.
```
gwiseman-r7: 👍

xml/telnet_banners.xml
```diff
@@ -0,0 +1,820 @@
...
+    </example>
+    <!-- Ubuntu 17.04\r\nnginx login: -->
+    <example _encoding="base64" os.version="17.04" host.name="nginx">
+      VWJ1bnR1IDE3LjA0DQpuZ2lueCBsb2dpbjoK
+    </example>
+    <param pos="0" name="os.vendor" value="Debian"/>
+    <param pos="0" name="os.family" value="Linux"/>
+    <param pos="0" name="os.product" value="Linux"/>
+    <param pos="1" name="os.version"/>
+    <param pos="0" name="os.cpe23" value="cpe:/o:debian:debian_linux:{os.version}"/>
```
gwiseman-r7: Looks like some copypasta, should the cpe here be `cpe:/o:canonical:ubuntu_linux:{os.version}`? Also the `os.vendor` should be `Ubuntu` I believe (initially thought Canonical but that seems inconsistent with the other recog files).

tsellers-r7: Ugh, thanks.
Also, good point Canonical.

xml/telnet_banners.xml
```diff
@@ -0,0 +1,820 @@
...
+  </fingerprint>
+  <fingerprint pattern="^(?:\r|\n)*DD-WRT v([\d.]+)(?:-(\w+))? ([\w-]+) \(c\) \d{4} NewMedia-NET GmbH(?:\r|\n)+Release: \d+\/\d+\/\d+ \(SVN revision: ([:\w]+)\)(?:\r|\n)+.* login:\s*$">
+    <description>DD-WRT - 24 family</description>
+    <!-- DD-WRT v24-sp2 mini (c) 2013 NewMedia-NET GmbH\r\nRelease: 05/27/13 (SVN revision: 21676)\r\n\r\nDD-WRT login: -->
+    <example _encoding="base64" os.version="24" os.version.version="sp2" os.edition="mini" os.build="21676">
+      REQtV1JUIHYyNC1zcDIgbWluaSAoYykgMjAxMyBOZXdNZWRpYS1ORVQgR21iSA0KUmVsZWFzZ
+      TogMDUvMjcvMTMgKFNWTiByZXZpc2lvbjogMjE2NzYpDQoNCkRELVdSVCBsb2dpbjoK
+    </example>
+    <!-- DD-WRT v24 micro (c) 2010 NewMedia-NET GmbH\r\nRelease: 08/07/10 (SVN revision: 14896)\r\n\r\nProliant DL980R07 X6550 8-core 4P SAS login: -->
+    <example _encoding="base64" os.version="24" os.edition="micro" os.build="14896">
```
jhart-r7: I thought this was missing a test for `os.version.version`, but maybe the point is that this example shouldn't be getting an `os.version.version` because that part of the regex is optional.  do you happen to know what happens if you added `os.version.version=""`?

xml/telnet_banners.xml
```diff
@@ -0,0 +1,820 @@
...
+  </fingerprint>
+  <fingerprint pattern="^ZyXEL \w?DSL Router\r\nLogin:$">
+    <description>ZyXEL Router - simple</description>
+    <!-- ZyXEL VDSL Router\r\nLogin: -->
+    <example _encoding="base64">WnlYRUwgVkRTTCBSb3V0ZXINCkxvZ2luOgo=</example>
+    <param pos="0" name="service.vendor" value="ZyXEL"/>
+    <param pos="0" name="hw.vendor" value="ZyXEL"/>
+    <param pos="0" name="hw.device" value="Router"/>
+  </fingerprint>
+  <fingerprint pattern="^Debian GNU\/Linux 9(?:\r|\n)+([\w.-]+) login:\s*$">
```
jhart-r7: The next two examples are missing the leading carriage return and newline part of the fingerprint.

tsellers-r7: I found no examples of particular banners with leading CR/LF.

jhart-r7: Ah, interesting.  For some reason I had assumed that all regexes in here were modified to account for the possible whitespace at the beginning of the banner.

xml/telnet_banners.xml
tsellers-r7: No `os.version.version` from that example.  In the past the tests fail when checking using that syntax.

xml/telnet_banners.xml
tsellers-r7: No, the current leading regex for CR/LF is left over from my pre-trim-the-data work.  It should only be present when I actually saw banners with leading/trailing CR, LF, nulls, and other whitespace.

xml/telnet_banners.xml
```diff
@@ -0,0 +1,817 @@
...
+      3cgYWJvdXQgbXlzZWxmOg0KTW9kZWw6ICAgICAgICAgICAgICAgVlNYIDYwMDBBDQpTZXJpYW
+      wgTnVtYmVyOiAgICAgICAwMDA3MDkwNkZDMzRGNg0KU29mdHdhcmUgVmVyc2lvbjogICAgUmV
+      sZWFzZSA5LjAuNi4yLTEwMyAtIDA0U2VwMjAxMSAyMToyNw0KQnVpbGQgSW5mb3JtYXRpb246
+      ICAgZWNvbW1hbgo=
+    </example>
+    <param pos="0" name="hw.vendor" value="Polycom"/>
+    <param pos="0" name="hw.family" value="VSX"/>
+    <param pos="0" name="hw.device" value="Video Conferencing"/>
+    <param pos="1" name="hw.product"/>
+    <param pos="2" name="hw.id"/>
```
jhart-r7: everywhere else, at least in this new file, `host.id` is used.

xml/telnet_banners.xml
```diff
@@ -0,0 +1,817 @@
...
+    <description>Huawei HG series Home Gateway routers</description>
+    <!-- Welcome Visiting Huawei Home Gateway\r\nCopyright by Huawei Technologies Co., Ltd.\r\n\r\nLogin: -->
+    <example _encoding="base64">
+      V2VsY29tZSBWaXNpdGluZyBIdWF3ZWkgSG9tZSBHYXRld2F5DQpDb3B5cmlnaHQgYnkgSHVhd
+      2VpIFRlY2hub2xvZ2llcyBDby4sIEx0ZC4NCg0KTG9naW46Cg==
+    </example>
+    <param pos="0" name="hw.vendor" value="Huawei"/>
+    <param pos="0" name="hw.device" value="Router"/>
+  </fingerprint>
+  <fingerprint pattern="^(?:\r|\n)*%connection closed by remote host!(?:\u0000)?$">
```
jhart-r7: minor, but may want to move this fingerprint up one, so that the Huawei ones are next to each other.  

(You said be critical!)

xml/telnet_banners.xml
```diff
@@ -0,0 +1,817 @@
...
+    </example>
+    <!-- DD-WRT v24 micro (c) 2010 NewMedia-NET GmbH\r\nRelease: 08/07/10 (SVN revision: 14896)\r\n\r\nProliant DL980R07 X6550 8-core 4P SAS login: -->
+    <example _encoding="base64" os.version="24" os.edition="micro" os.build="14896">
+      REQtV1JUIHYyNCBtaWNybyAoYykgMjAxMCBOZXdNZWRpYS1ORVQgR21iSA0KUmVsZWFzZTogM
+      DgvMDcvMTAgKFNWTiByZXZpc2lvbjogMTQ4OTYpDQoNClByb2xpYW50IERMOTgwUjA3IFg2NT
+      UwIDgtY29yZSA0UCBTQVMgbG9naW46Cg==
+    </example>
+    <param pos="0" name="os.vendor" value="DD-WRT"/>
+    <param pos="0" name="os.product" value="DD-WRT"/>
+    <param pos="0" name="os.device" value="Router"/>
```
jhart-r7: why no `hw` and/or `hw.device?   in many other cases where you are using `os.device` you are also using `hw.device`.

tsellers-r7: Thanks, I should had a `hw.device` but I can't add any other `hw.` params since the OS can be loaded on multiple brands.

PR #180, opened 2018-09-25
Add matches to all recog fingerprints

jhart-r7:
The current functionality is that this value will default to the `basename` of the fingerprint file in which it was found minus the .xml extension.  The problem with this is that you can't then easily have multiple files for the same protocol response, which might be useful for very large or complicated protocols like HTTP where we might want to keep fingerprints grouped in separate files for some reason.  Furthermore, I think it opens us up to more bugs with products that only integrate recog content which rely on specific recog XML file names but there are no corresponding tests in recog land to ensure that those XML files always exist.   Using the `matches` field is a more reliable and useful way of doing this.  Debating also making it required.

### DIFF
--- a/xml/h323_callresp.xml
+++ b/xml/h323_callresp.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<fingerprints protocol="h.323" database_type="service" preference="0.80">
+<fingerprints matches="h323.setup_response" protocol="h.323" database_type="service" preference="0.80">
   <!--
   Responses to H.323 call SETUP messages are matched against these patterns
   to fingerprint H.323 servers.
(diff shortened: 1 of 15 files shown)


tsellers-r7:
I'm ok with this in general though we will need to ensure this is called out in the Release notes and perhaps on Twitter since it'll be a breaking change for folks referencing the file name.

jhart-r7:
To clarify this should not impact anyone referencing the file by file name as those are not changing.  It will impact anyone who is using the fallback behavior when a `matches` is not present, in which case it takes the basename of the file minus the .xml extension.  This will change for files like `smtp_vrfy.xml`, because fallback-wise today this would be `smtp_vrfy`, but what I'm changing it to here is `smtp.verify` to be consistent with other current usages of `matches`.  Perhaps the best middle ground is to change those without `matches` to have a value that is equivalent to the basename minus .xml.

PR #181, opened 2018-09-25
Support interpolation of fingerprint values

jhart-r7:
When we added the CPE functionality to recog xml/product, we also introduced the concept of value interpolation such that if you had a `fingerprint` with a `params` like:

```
    ...
    <param pos="1" name="service.component.version"/>
    <param pos="0" name="service.component.cpe23" value="cpe:/a:redhat:stronghold:{service.component.version}"/>
   ...
```

The idea is that the value of `service.component.cpe23` would be composed of `cpe:/a:redhat:stronghold:` and the value of `service.component.version` which is extracted at fingerprint match time.

This PR adds the necessary functionality to recog "product" to enable this.  I updated one simple `param` as an example of this.  I added a check/test for this in `Fingerprint` itself, but part of me feels like it should be in the self test spec instead.  

@tsellers-r7 looking for your feedback here.  I believe we had discussed possibly allowing this interpolation to happen recursively and perhaps even supporting it in the `param` `name` themselves.

### DIFF
--- a/lib/recog/fingerprint.rb
+++ b/lib/recog/fingerprint.rb
@@ -100,13 +100,27 @@ class Fingerprint
       next if v.nil?
       # if this key's value uses interpolation of the form "foo{some.thing}",
       # if some.thing was "bar" then this keys value would be set to "foobar".
-      if matches = v.match(/\{(?<replace>[^\s{}]+)\}/)
-        replace = matches[:replace]
+      if /\{(?<replace>[^\s{}]+)\}/ =~ v
         if result[replace]
-          if recursive_match = result[replace].match(/\{(?<bad_replace>[^\s{}]+)\}/)
-            raise "Invalid recursive use of #{recursive_match[:bad_replace]} in #{replace}"
+          if /\{(?<bad_replace>[^\s{}]+)\}/ =~ result[replace]
+            raise "Invalid recursive use of #{bad_replace} in #{replace}"
           end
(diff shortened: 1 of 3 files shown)


jhart-r7:
The current travis failure is peculiar.  It seems like this is some fundamental bug in jruby -- it is somehow getting the extracted value from a totally different `example`:

```
  1) Recog::DB#http_servers.xml (?-mix:^Stronghold\/(\d\.\d) Apache\/([012][\d.]*)\s*(.*)$) Example 'Stronghold/4.0 Apache/1.3.22 (Unix) mod_ssl/2.8.7 OpenSSL/0.9.6c mod_perl/1.26' matches this regex
     Failure/Error: expect(match[k]).to eq(v), "Regex didn't extract expected value for fingerprint attribute #{k} -- got #{match[k]} instead of #{v}"
       Regex didn't extract expected value for fingerprint attribute service.cpe23 -- got cpe:/a:apache:http_server:1.3.19 instead of cpe:/a:apache:http_server:1.3.22
     # ./spec/lib/fingerprint_self_test_spec.rb:118:in `block in /home/travis/build/rapid7/recog/spec/lib/fingerprint_self_test_spec.rb'
     # ./spec/lib/fingerprint_self_test_spec.rb:116:in `block in /home/travis/build/rapid7/recog/spec/lib/fingerprint_self_test_spec.rb'
```

PR #188, opened 2018-10-17
Improve telnet fingerprinting

jpincombe-r7:
We have improved recog's telnet coverage by adding various fingerprints that were previously missing. Most notably the addition of Red Hat, Fedora, AIX and SuSE telnet coverage, as well as adding to the already Cisco coverage that was previously being used.

### DIFF
--- a/xml/telnet_banners.xml
+++ b/xml/telnet_banners.xml
@@ -993,7 +993,7 @@
   <fingerprint pattern=".*Nortel.*Passport ([^ ]*) .*Software Release ([^ ]*).*">
     <description>Nortel Passport</description>
     <!-- *********************************************\n\n\n* Copyright (c) 2003 Nortel Networks, Inc.  *\n\n\n* All Rights Reserved                       *\n\n\n* Passport 8010                             *\n\n\n* Software Release 3.5.0.0                  *\n\n\n*********************************************\n\n\n\n\nLogin: -->
-    <example _encoding="base64" hw.product="8010" os.version="3.5.0.0">
+    <example _encoding="base64" os.product="8010" os.version="3.5.0.0">
       KioqKioqKioqKioqKioqKioqKioqKioqKioqKioqKioqKioqKioqKioqKioqXG5cblxuKiBDb3B5cmlnaHQgKG
       MpIDIwMDMgTm9ydGVsIE5ldHdvcmtzLCBJbmMuICAqXG5cblxuKiBBbGwgUmlnaHRzIFJlc2VydmVkICAgICAg
       ICAgICAgICAgICAgICAgICAqXG5cblxuKiBQYXNzcG9ydCA4MDEwICAgICAgICAgICAgICAgICAgICAgICAgIC
--- a/xml/telnet_banners.xml
+++ b/xml/telnet_banners.xml
@@ -834,7 +834,7 @@
     </example>
     <param pos="0" name="hw.vendor" value="RedHat"/>
     <param pos="0" name="hw.family" value="Linux"/>
-    <param pos="0" name="hw.device" value="Linux"/>
+    <param pos="0" name="os.product" value="Linux"/>
     <param pos="1" name="os.version"/>
     <param pos="2" name="linux.kernel.version"/>
     <param pos="3" name="os.arch"/>
@@ -846,7 +846,7 @@
     </example>
     <param pos="0" name="hw.vendor" value="RedHat"/>
     <param pos="0" name="hw.family" value="Linux"/>
-    <param pos="0" name="hw.device" value="Linux"/>
+    <param pos="0" name="hw.device" value="General"/>
     <param pos="0" name="hw.product" value="RedHat Enterprise AS"/>
     <param pos="1" name="os.version"/>
     <param pos="2" name="linux.kernel.version"/>
@@ -860,7 +860,7 @@
     </example>
     <param pos="0" name="hw.vendor" value="RedHat"/>
     <param pos="0" name="hw.family" value="Linux"/>
-    <param pos="0" name="hw.device" value="Linux"/>
+    <param pos="0" name="hw.device" value="General"/>
     <param pos="0" name="hw.product" value="RedHat Enterprise WS"/>
     <param pos="1" name="os.version"/>
     <param pos="2" name="linux.kernel.version"/>
@@ -874,6 +874,7 @@
     <param pos="0" name="hw.vendor" value="Redhat"/>
     <param pos="0" name="hw.family" value="Linux"/>
     <param pos="0" name="hw.product" value="Fedora"/>
+    <param pos="0" name="hw.device" value="General"/>
     <param pos="1" name="os.version"/>
     <param pos="2" name="linux.kernel.version"/>
     <param pos="3" name="os.arch"/>
@@ -885,11 +886,14 @@
     </example>
     <param pos="0" name="hw.vendor" value="SUSE"/>
     <param pos="0" name="hw.family" value="Linux"/>
+    <param pos="0" name="os.product" value="Linux"/>
+    <param pos="0" name="hw.device" value="General"/>
     <param pos="1" name="os.version"/>
     <param pos="2" name="os.arch"/>
     <param pos="3" name="linux.kernel.version"/>
   </fingerprint>
   <fingerprint pattern="^Turbolinux ApplianceServer (\d+\.\d+).*">
+    <description>Turbolinux ApplianceServer</description>
     <example _encoding="base64" os.version="4.0">
      VHVyYm9saW51eCBBcHBsaWFuY2VTZXJ2ZXIgNC4wIChBdGxhczIpIExpbnV4IDIuNi4zMi00MzEuMjMuMy5lbDYueDg
      2XzY0IG9uIGEgeDg2XzY0IChzZW55bzE5MXg4OS5kaWdpdGFsaW5rLm5lLmpwKSBUVFk6IDEyOjE1IG9uIFR1ZXNkYX
@@ -897,6 +901,8 @@
     </example>
     <param pos="0" name="hw.vendor" value="Turbolinux"/>
     <param pos="0" name="hw.family" value="Linux"/>
+    <param pos="0" name="os.product" value="Linux"/>
+    <param pos="0" name="hw.device" value="General"/>
     <param pos="1" name="os.version"/>
   </fingerprint>
   <fingerprint pattern="^UnixWare ([^ ]+).*$">
@@ -907,6 +913,7 @@
     <param pos="0" name="hw.vendor" value="SCO"/>
     <param pos="0" name="hw.family" value="UnixWare"/>
     <param pos="0" name="hw.device" value="UnixWare"/>
+    <param pos="0" name="hw.product" value="UnixWare"/>
     <param pos="1" name="os.version"/>
   </fingerprint>
   <fingerprint pattern="^Telnet Server Build (5.*)">
@@ -917,13 +924,15 @@
     </example>
     <param pos="0" name="hw.vendor" value="Microsoft"/>
     <param pos="0" name="hw.family" value="Windows"/>
+    <param pos="0" name="hw.product" value="Windows 2000"/>
+    <param pos="0" name="hw.device" value="General"/>
     <param pos="1" name="os.version"/>
   </fingerprint>
   <fingerprint pattern="^Welcome. Type return, enter password at # prompt">
+    <description>Brother Printer</description>
     <example _encoding="base64">
      V2VsY29tZS4gVHlwZSByZXR1cm4sIGVudGVyIHBhc3N3b3JkIGF0ICMgcHJvbXB0Cg== 
     </example>
-    <description>Brother Printer</description>
     <param pos="0" name="hw.vendor" value="Brother"/>
     <param pos="0" name="hw.device" value="Printer" />
     <param pos="0" name="hw.product" value="Brother Printer"/>
@@ -935,6 +944,7 @@
     </example>
     <param pos="0" name="hw.vendor" value="Arescom"/>
     <param pos="0" name="hw.device" value="WAP" />
+    <param pos="0" name="hw.family" value="UNKNOWN"/>
     <param pos="1" name="hw.model"/>
   </fingerprint>
   <fingerprint pattern="Welcome to ViewStation">
@@ -943,25 +953,9 @@
     V2VsY29tZSB0byBWaWV3U3RhdGlvbgoKUGFzc3dvcmQ6
     </example>
     <param pos="0" name="hw.vendor" value="Polycom"/>
-    <param pos="0" name="hw.product" value="ViewStation"/>
-  </fingerprint>
-  <fingerprint pattern="Here is what I know about myself:">
-    <description>Polycom ViewStation Video Conference System</description>
-    <example _encoding="base64">
-     SGksIG15IG5hbWUgaXMgOiAgICAgUklER0VMQU5EIENBTVBVUwpIZXJlIGlzIHdoYXQgSSBrbm93IGFib3V0IG15c2VsZjo
-     KTW9kZWw6ICAgICAgICAgICAgICAgVlNYIDgwMDAKU2VyaWFsIE51bWJlcjogICAgICAgODIwODA4MDkzNTMyTjIKU29mdH
-     dhcmUgVmVyc2lvbjogICAgUmVsZWFzZSA5LjAuNi4yLTEwMyAtIDA0U2VwMjAxMSAyMToyNwpCdWlsZCBJbmZvcm1hdGlvb
-     jogICBlY29tbWFuZGVyIG9uIGVuZ2ludDAwLmF1c3Rpbi5wb2x5Y29tLmNvbQpGUEdBIFJldmlzaW9uOiAgICAgICA0LjQu
-     MgpNYWluIFByb2Nlc3NvcjogICAgICBCU1AxNSAgIHYwLjAgfiBDb3JlL01lbSBDbGtzIDQwNS8xMzUgIFszOjQgMDozXQp
-     UaW1lIEluIExhc3QgQ2FsbDogICAxOjU4OjAxClRvdGFsIFRpbWUgSW4gQ2FsbHM6IDE1NzowNDo1NQpUb3RhbCBDYWxscz
-     ogICAgICAgICAyMDYKU05UUCBUaW1lIFNlcnZpY2U6ICAgYXV0byBpbnN5bmMgMC5wb29sLm50cC5vcmcKTG9jYWwgVGltZ
-     SBpczogICAgICAgIFdlZCwgIDMgT2N0IDIwMTggMjE6MDg6MTcgLTExMDAKTmV0d29yayBJbnRlcmZhY2U6ICAgTk9ORQpJ
-     UCBWaWRlbyBOdW1iZXI6ICAgICAxNzIuMjAuMTkyLjI0OApJU0ROIFZpZGVvIE51bWJlcjogICAxLgpNUCBFbmFibGVkOiA
-     gICAgICAgICBLRUEwLTc5REEtNTZFMC0wMDAwLTAwODEKSDMyMyBFbmFibGVkOiAgICAgICAgVHJ1ZQpGVFAgRW5hYmxlZD
-     ogICAgICAgICBUcnVlCkhUVFAgRW5hYmxlZDogICAgICAgIFRydWUKU05NUCBFbmFibGVkOiAgICAgICAgRmFsc2U=
-    </example>
-    <param pos="0" name="hw.vendor" value="Polycom"/>
-    <param pos="0" name="hw.product" value="ViewStation"/>
+    <param pos="0" name="hw.device" value="ViewStation"/>
+    <param pos="0" name="hw.product" value="UNKNOWN"/>
+    <param pos="0" name="hw.family" value="UNKNOWN"/>
   </fingerprint>
   <fingerprint pattern="FlowPoint\/(.*) SDSL \[ATM\] Router .*v(.*) Ready">
     <description>iFlowPoint 2200 DSL router</description>
@@ -971,6 +965,7 @@
     <param pos="0" name="hw.vendor" value="Flowpoint"/>
     <param pos="0" name="hw.device" value="Broadband Router"/>
     <param pos="0" name="hw.product" value="DSL router"/>
+    <param pos="0" name="hw.family" value="UNKNOWN"/>
     <param pos="1" name="hw.model"/>
     <param pos="2" name="os.version"/>
   </fingerprint>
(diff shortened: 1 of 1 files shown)


xml/telnet_banners.xml
```diff
@@ -816,4 +816,589 @@
...
+    <example _encoding="base64">
+      RGlnaXRhbCBVTklYIChqb3VybmFsKSAodHR5cDIpCgoKCmxvZ2luOiA=
+    </example>
+    <param pos="0" name="hw.vendor" value="HP"/>
+    <param pos="0" name="hw.family" value="Digital Unix"/>
+    <param pos="0" name="hw.device" value="General"/>
+    <param pos="0" name="hw.product" value="Digital Unix"/>
+  </fingerprint>
+  <fingerprint pattern="Compaq Tru64 UNIX V(.*) \(Rev. (.*\d)\) .*" flags="REG_MULTILINE">
+    <description>System is Compaq Tru64 UNIX V</description>
```
jhart-r7: Here and in the other descriptions that start with "System is", remove that snippet.

xml/telnet_banners.xml
```diff
@@ -816,4 +816,589 @@
...
+      VMIG9uIGFuIHg4Nl82NApsb2dpbjo=
+    </example>
+    <param pos="0" name="hw.vendor" value="RedHat"/>
+    <param pos="0" name="hw.family" value="Linux"/>
+    <param pos="0" name="os.product" value="Linux"/>
+    <param pos="1" name="os.version"/>
+    <param pos="2" name="linux.kernel.version"/>
+    <param pos="3" name="os.arch"/>
+  </fingerprint>
+  <fingerprint pattern="^Red Hat Enterprise Linux AS release (.*) \(.*\).*Kernel (.*) on a[^ ]* ([^ ]*\d)" flags="REG_MULTILINE">
```
jhart-r7: I realize this is not documented and it is in use elsewhere, but please try to avoid using `flags`.  These are language-specific constants that makes it harder for the recog content to be integrated separately from the "product".  Instead, please make the entire regular expression case insensitive by adding a `(?i) to the beginning, just after the leading anchor `^`

xml/telnet_banners.xml
```diff
@@ -816,4 +816,589 @@
...
+  </fingerprint>
+  <fingerprint pattern="This copy of the Ataman TCP Remote Logon Services">
+    <description>Windows NT/2k/2k3 running Ataman telnet server</description>
+    <example _encoding="base64">
+      VGhpcyBjb3B5IG9mIHRoZSBBdGFtYW4gVENQIFJlbW90ZSBMb2dvbiBTZXJ2aWNlcyBpcyByZWdpc3RlcmVkIG
+      FzIGxpY2Vuc2VkIHRvOgoJRUNJMi9ERE1TCgpBY2NvdW50IE5hbWU6IA==
+    </example>
+    <param pos="0" name="hw.vendor" value="Microsoft"/>
+    <param pos="0" name="hw.family" value="Windows"/>
+    <param pos="0" name="hw.device" value="General"/>
```
jhart-r7: Here and everywhere, `hw.device` of `General` should be avoided.  It provides no value.

xml/telnet_banners.xml
```diff
@@ -816,4 +816,589 @@
...
+      U2VhdHRsZSBDb21tdW5pdHkgTmV0d29yayBTdW4gU29sYXJpcyAxLjEuMS5CClBsZWFzZSBsb2dpbiBhcyAndml
+      zaXRvcicgaWYgeW91IGFyZSBhIHZpc2l0b3IKCgpTdW5PUyBVTklYIChzY24pCgoKCmxvZ2luOg==
+    </example>
+    <param pos="0" name="hw.vendor" value="Sun"/>
+    <param pos="0" name="hw.family" value="Solaris"/>
+    <param pos="0" name="hw.device" value="General"/>
+    <param pos="0" name="hw.product" value="Solaris"/>
+    <param pos="1" name="os.version"/>
+  </fingerprint>
+  <fingerprint pattern="Digital UNIX">
```
jhart-r7: This seems overly loose.  What about times when the version appears here?  This will match instead.

xml/telnet_banners.xml
```diff
@@ -816,4 +816,589 @@
...
+  </fingerprint>
+  <fingerprint pattern="Data ONTAP">
+    <description>A NetApp apliance</description>
+    <example _encoding="base64">RGF0YSBPTlRBUCAoczUwMC4pCmxvZ2luOiA=</example>
+    <param pos="0" name="hw.vendor" value="NetApp"/>
+    <param pos="0" name="hw.family" value="Data ONTAP"/>
+    <param pos="0" name="hw.device" value="Storage"/>
+    <param pos="0" name="hw.product" value="Data ONTAP"/>
+  </fingerprint>
+  <fingerprint pattern="OpenVMS">
```
rkirk-r7: can we improve/tighten this? what about version data?

xml/telnet_banners.xml
```diff
@@ -816,4 +816,589 @@
...
+    <param pos="1" name="os.version"/>
+  </fingerprint>
+  <fingerprint pattern="% Username:  timeout expired!">
+    <description>Some kind of Cisco device</description>
+    <example _encoding="base64">
+      JSBVc2VybmFtZTogIHRpbWVvdXQgZXhwaXJlZCE=
+    </example>
+    <param pos="0" name="hw.vendor" value="Cisco"/>
+    <param pos="0" name="hw.family" value="IOS"/>
+    <param pos="0" name="hw.device" value="UNKNOWN"/>
```
rkirk-r7: should this be set if we don't know?

xml/telnet_banners.xml
```diff
@@ -816,4 +816,589 @@
...
+    <description>a DrayTek Vigor SOHO Router</description>
+    <example _encoding="base64">
+      U3lzdGVtIGFkbWluaXN0cmF0b3IgaXMgY29ubmVjdGluZyBmcm9tIDU0LjM5LjE3My44NgoKUmVqZWN0IH
+      RoZSBjb25uZWN0aW9uIHJlcXVlc3QgISEh
+    </example>
+    <param pos="0" name="hw.vendor" value="Draytek"/>
+    <param pos="0" name="hw.device" value="Broadband router"/>
+    <param pos="0" name="hw.product" value="Vigor"/>
+  </fingerprint>
+  <fingerprint pattern="OpenBSD">
```
rkirk-r7: versions? can we tighten this?

xml/telnet_banners.xml
```diff
@@ -816,4 +816,589 @@
...
+      RnJlZUJTRC9hbWQ2NCAobXMuZ3ltc3Bncy5jeikgKHB0cy8wKQoKCgpsb2dpbjo=
+    </example>
+    <param pos="0" name="hw.vendor" value="FreeBSD"/>
+    <param pos="0" name="hw.family" value="FreeBSD"/>
+    <param pos="0" name="hw.device" value="General"/>
+    <param pos="0" name="hw.product" value="FreeBSD"/>
+    <param pos="1" name="os.arch"/>
+    <param pos="2" name="host.name"/>
+  </fingerprint>
+  <fingerprint pattern="NetBSD">
```
rkirk-r7: versions? can we tighten this?

xml/telnet_banners.xml
```diff
@@ -816,4 +816,589 @@
...
+    <description>NetBSD</description>
+    <example _encoding="base64">
+      TmV0QlNEL21lc3NpbWlwcyAoKSAodHR5cDMpCgpsb2dpbjog
+    </example>
+    <param pos="0" name="hw.vendor" value="NetBSD"/>
+    <param pos="0" name="hw.family" value="NetBSD"/>
+    <param pos="0" name="hw.device" value="General"/>
+    <param pos="0" name="hw.product" value="NetBSD"/>
+  </fingerprint>
+  <fingerprint pattern="IRIX">
```
rkirk-r7: versions? can we tighten this?

xml/telnet_banners.xml
```diff
@@ -816,4 +816,589 @@
...
+    <description>SGI IRIX</description>
+    <example _encoding="base64">
+      SVJJWCAoYXJ0ZW1pcy5iaW9sLnVvYS5ncikKCgoKbG9naW46IA==
+    </example>
+    <param pos="0" name="hw.vendor" value="SGI"/>
+    <param pos="0" name="hw.family" value="IRIX"/>
+    <param pos="0" name="hw.device" value="General"/>
+    <param pos="0" name="hw.product" value="IRIX"/>
+  </fingerprint>
+  <fingerprint pattern="^(RS [^\s]+) System Software, Version ([^\s]+).*" flags="REG_MULTILINE">
```
rkirk-r7: should we use `(?s)` instead of the flag? as per Jon's previous comment

xml/telnet_banners.xml
```diff
@@ -816,4 +816,589 @@
...
+   </example>
+    <param pos="0" name="hw.vendor" value="RedHat"/>
+    <param pos="0" name="he.family" value="Linux"/>
+    <param pos="0" name="hw.device" value="Linux"/>
+    <param pos="1" name="os.version"/>
+  </fingerprint>
+  <fingerprint pattern="Red Hat Enterprise Linux ES release (.*) \(.*\).*Kernel (.*) on a[^ ]* ([^ ]*\d)" flags="REG_MULTILINE">
+    <description>RedHat Enterprise Linux ES</description>
+    <example _encoding="base64" os.version="3" linux.kernel.version="2.4.21-47.EL" os.arch="x86_64">
+      UmVkIEhhdCBFbnRlcnByaXNlIExpbnV4IEVTIHJlbGVhc2UgMyAoVGFyb29uIFVwZGF0ZSA5KQpLZXJuZWwgMi40LjIxLTQ3Lk
```
jhart-r7: Please include the plain-text version of the example in a comment above each example, here and everywhere.  Otherwise, review and maintenance of these files will become harder.

xml/telnet_banners.xml
```diff
@@ -816,4 +816,589 @@
...
+  </fingerprint>
+  <fingerprint pattern="Welcome to ViewStation">
+    <description>Polycom ViewStation Video Vonference System</description>
+    <example _encoding="base64">
+      V2VsY29tZSB0byBWaWV3U3RhdGlvbgoKUGFzc3dvcmQ6
+    </example>
+    <param pos="0" name="hw.vendor" value="Polycom"/>
+    <param pos="0" name="hw.device" value="ViewStation"/>
+    <param pos="0" name="hw.product" value="UNKNOWN"/>
+    <param pos="0" name="hw.family" value="UNKNOWN"/>
```
jhart-r7: I'm not sure we should be using UNKNOWN like this.  If we don't know a particular field, just leave it out.

xml/telnet_banners.xml
```diff
@@ -816,4 +816,589 @@
...
+    <example _encoding="base64">
+      VGhpcyBjb3B5IG9mIHRoZSBBdGFtYW4gVENQIFJlbW90ZSBMb2dvbiBTZXJ2aWNlcyBpcyByZWdpc3RlcmVkIG
+      FzIGxpY2Vuc2VkIHRvOgoJRUNJMi9ERE1TCgpBY2NvdW50IE5hbWU6IA==
+    </example>
+    <param pos="0" name="hw.vendor" value="Microsoft"/>
+    <param pos="0" name="hw.family" value="Windows"/>
+    <param pos="0" name="hw.device" value="General"/>
+    <param pos="0" name="hw.product" value="Windows"/>
+  </fingerprint>
+  <fingerprint pattern="Cobalt Linux release (.*)$">
```
jhart-r7: the example fingerprint makes for problematic versions because it includes whitespace, etc.  Maybe just catch the numeric version?

xml/telnet_banners.xml
```diff
@@ -816,4 +816,589 @@
...
+      UmFwdG9yIEZpcmV3YWxsIFNlY3VyZSBHYXRld2F5LgoKSG9zdG5hbWU6IA==
+    </example>
+    <param pos="0" name="hw.vendor" value="Symantec"/>
+    <param pos="0" name="hw.family" value="Raptor"/>
+    <param pos="0" name="hw.device" value="Firewall"/>
+    <param pos="0" name="hw.product" value="Raptor"/>
+  </fingerprint>
+  <fingerprint pattern="UNIX\(r\) System V Release 4.0">
+    <description>SunOS 5.x (Solaris)</description>
+    <example _encoding="base64">
```
jhart-r7: no version?

xml/telnet_banners.xml
jpincombe-r7: Just to clarify in this example do you mean to remove the "Shinkansen"?

xml/telnet_banners.xml
jhart-r7: Sonar also finds 6 hosts.  No version, but all have hostnames in them that we should extract:

```
Digital UNIX (C126927) (ttyp0)
Digital UNIX (some.host.name) (ttyp0)
```

xml/telnet_banners.xml
```diff
@@ -816,4 +816,680 @@
...
+    </example>
+    <param pos="0" name="hw.vendor" value="Cisco"/>
+    <param pos="0" name="hw.family" value="UCS"/>
+    <param pos="0" name="hw.device" value="General"/>
+    <param pos="0" name="hw.product" value="UCS Device"/>
+    <param pos="1" name="host.name"/>
+  </fingerprint>
+  <fingerprint pattern="^(?m)HP ProLiant.*v(\d+.\d+)">
+    <description>Sytem is HP ProLiant server</description>
+    <example _encoding="base64" os.version="2.00">
```
jhart-r7: Minor, but missing plain-text example.

xml/telnet_banners.xml
```diff
@@ -816,4 +816,680 @@
...
+    <description>an Imagistics device</description>
+    <!-- Imagistics im3511/im4511 Ver 01.00.20 TELNET server.\nCopyright(c) 2001-2005, silex technology, Inc.\nlogin: -->
+    <example _encoding="base64">
+      SW1hZ2lzdGljcyBpbTM1MTEvaW00NTExIFZlciAwMS4wMC4yMCBURUxORVQgc2VydmVyLgpDb3B5cmlnaH
+      QoYykgMjAwMS0yMDA1LCBzaWxleCB0ZWNobm9sb2d5LCBJbmMuCmxvZ2luOiA=
+    </example>
+    <param pos="0" name="hw.vendor" value="Imagistics"/>
+    <param pos="0" name="hw.family" value="Imagistics"/>
+    <param pos="0" name="hw.device" value="Malfunction Device"/>
+    <param pos="0" name="hw.product" value="im"/>
```
jhart-r7: is this product correct?   lowercase `im`?

xml/telnet_banners.xml
```diff
@@ -816,4 +816,680 @@
...
+    </example>
+    <param pos="0" name="hw.vendor" value="Riverstone"/>
+    <param pos="0" name="hw.family" value="ES"/>
+    <param pos="0" name="hw.device" value="Router"/>
+    <param pos="1" name="hw.product"/>
+    <param pos="2" name="os.version"/>
+  </fingerprint>
+  <fingerprint pattern="HP.*ProCurve Switch">
+    <description>HP ProCurve Switch</description>
+    <!-- ==============================================================================\nHP J4121A ProCurve Switch 4000M\n
```
jhart-r7: These appear to have product names in them.  Those should be extracted and tested.

xml/telnet_banners.xml
```diff
@@ -816,4 +816,680 @@
...
+      SVJJWCAoYXJ0ZW1pcy5iaW9sLnVvYS5ncikKCgoKbG9naW46IA==
+    </example>
+    <param pos="0" name="hw.vendor" value="SGI"/>
+    <param pos="0" name="hw.family" value="IRIX"/>
+    <param pos="0" name="hw.product" value="IRIX"/>
+    <param pos="1" name="host.name"/>
+  </fingerprint>
+  <fingerprint pattern="^(?m)(RS [^\s]+) System Software, Version ([^\s]+).*">
+    <description>RiverStone router</description>
+    <!-- ++++++++++++++++++++++++++++++++++++++++\nRS 8000 System Software, Version 9.3.0.5\n
```
jhart-r7: If there really are `+` at the beginning of the banners, your regex will not match.

xml/telnet_banners.xml
```diff
@@ -816,4 +816,680 @@
...
+  <fingerprint pattern="Digital UNIX \(([^)]+).*">
+    <description>Digital Unix</description>
+    <!-- Digital UNIX (journal) (ttyp2)\n\n\nlogin: -->
+    <example _encoding="base64" host.name="journal">
+      RGlnaXRhbCBVTklYIChqb3VybmFsKSAodHR5cDIpCgoKCmxvZ2luOiA=
+    </example>
+    <param pos="0" name="hw.vendor" value="HP"/>
+    <param pos="0" name="hw.family" value="Digital Unix"/>
+    <param pos="0" name="hw.product" value="Digital Unix"/>
+    <param pos="1" name="host.name"/>`
```
jhart-r7: trailing `

xml/telnet_banners.xml
```diff
@@ -816,4 +816,680 @@
...
+    \n++++++++++++++++++++++++++++++++++++++++ -->
+    <example _encoding="base64" hw.product="RS 8000" os.version="9.3.0.5">
+      LS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tL
+      S0tLS0tLQpSUyA4MDAwIFN5c3RlbSBTb2Z0d2FyZSwgVmVyc2lvbiA5LjMuMC41ClJpdmVyc3RvbmUgTmV0d2
+      9ya3MsIEluYy4sIENvcHlyaWdodCAoYykgMjAwMC0yMDA0LiBBbGwgcmlnaHRzIHJlc2VydmVkLgpTeXN0ZW0
+      gc3RhcnRlZCBvbiAyMDE4LTEwLTExIDIyOjAyOjAzCi0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0t
+      LS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS4uLg==
+    </example>
+    <param pos="0" name="hw.vendor" value="Riverstone"/>
+    <param pos="0" name="hw.family" value="Riverstone"/>
```
jhart-r7: Why Riverstone here but ES in the next fingerprint?

xml/telnet_banners.xml
```diff
@@ -816,4 +816,680 @@
...
+    <param pos="0" name="hw.device" value="Router"/>
+    <param pos="1" name="hw.product"/>
+    <param pos="2" name="os.version"/>
+  </fingerprint>
+  <fingerprint pattern="(ES [^\s]+) System Software, Version ([^\s]+)">
+    <description>a Riverstone router`</description>
```
jhart-r7: trailing `.

Also, the RS and ES fingerprints could be combined, extracting the family.

xml/telnet_banners.xml
```diff
@@ -816,4 +816,680 @@
...
+    <param pos="4" name="hw.series"/>
+    <param pos="5" name="hw.model"/>
+  </fingerprint>
+  <fingerprint pattern="Data ONTAP">
+    <description>A NetApp apliance</description>
+    <!-- Data ONTAP (s500.)\nlogin: -->
+    <example _encoding="base64">RGF0YSBPTlRBUCAoczUwMC4pCmxvZ2luOiA=</example>
+    <param pos="0" name="hw.vendor" value="NetApp"/>
+    <param pos="0" name="hw.family" value="Data ONTAP"/>
+    <param pos="0" name="hw.device" value="Storage"/>
```
kkuczkowski-r7: should this be `Server?

xml/telnet_banners.xml
jhart-r7: Ah, the reason this works is because of the anchor in combination with multi-line:

```
  ^      assert start of string (or line, in multiline mode)
```

xml/telnet_banners.xml
jpincombe-r7: No I do not believe as this is what we currently assert and when looking online it is described as a storage device

kkuczkowski-r7: We are classifying this as a `Server` device in authenticated fingerprinting, perhaps we want to revise that.

xml/telnet_banners.xml
```diff
@@ -816,4 +816,672 @@
...
+  <fingerprint pattern="Data ONTAP">
+    <description>A NetApp apliance</description>
+    <!-- Data ONTAP (s500.)\nlogin: -->
+    <example _encoding="base64">RGF0YSBPTlRBUCAoczUwMC4pCmxvZ2luOiA=</example>
+    <param pos="0" name="os.vendor" value="NetApp"/>
+    <param pos="0" name="os.family" value="Data ONTAP"/>
+    <param pos="0" name="os.device" value="Server"/>
+    <param pos="0" name="os.product" value="Data ONTAP"/>
+  </fingerprint>
+  <fingerprint pattern="OpenVMS.*Version\sV(\d*.\d*).*">
```
amcclenaghan-r7: There are versions of this which won't be matched by this regex, ie 7.3-1

https://www.shodan.io/host/203.164.10.116

xml/telnet_banners.xml
```diff
@@ -816,4 +816,679 @@
...
+    <example _encoding="base64">
+      SW1hZ2lzdGljcyBpbTM1MTEvaW00NTExIFZlciAwMS4wMC4yMCBURUxORVQgc2VydmVyLgpDb3B5cmlnaH
+      QoYykgMjAwMS0yMDA1LCBzaWxleCB0ZWNobm9sb2d5LCBJbmMuCmxvZ2luOiA=
+    </example>
+    <param pos="0" name="os.vendor" value="Imagistics"/>
+    <param pos="0" name="os.family" value="Imagistics"/>
+    <param pos="0" name="os.device" value="Malfunction Device"/>
+    <param pos="0" name="os.product" value="im"/>
+  </fingerprint>
+  <fingerprint pattern="NRG Maintenance Shell">
```
jhart-r7: Beginning of line anchor?

xml/telnet_banners.xml
```diff
@@ -816,4 +816,679 @@
...
+      iBwcm9kdWN0cyBub3cgYXQ6IHd3dy5Qcm9DdXJ2ZS5jb20KPT09PT09PT09PT09PT09PT09PT
+      09PT09PT09PT09PT09PT09PT09PT09PT09PT09PT09PT09PT09PT09PT09PT09PT09PT09PT0
+      9PT09CgoKVXNlcm5hbWU6IA==
+    </example>
+    <param pos="0" name="os.vendor" value="HP"/>
+    <param pos="0" name="os.family" value="ProCurve"/>
+    <param pos="0" name="os.device" value="Switch"/>
+    <param pos="1" name="os.product"/>
+  </fingerprint>
+  <fingerprint pattern="ConnectUPS">
```
jhart-r7: anchor?

xml/telnet_banners.xml
```diff
@@ -816,4 +816,678 @@
...
+    <param pos="0" name="os.vendor" value="Polycom"/>
+    <param pos="0" name="os.device" value="ViewStation"/>
+  </fingerprint>
+  <fingerprint pattern="^FlowPoint\/(.*) SDSL \[ATM\] Router .*v(.*) Ready">
+    <!--FlowPoint/2200 SDSL [ATM] Router fp2200-12 v3.0.2 Ready\nLogin:  -->
+    <description>FlowPoint 2200 DSL router</description>
+    <example _encoding="base64" hw.model="2200" os.version="3.0.2">
+      Rmxvd1BvaW50LzIyMDAgU0RTTCBbQVRNXSBSb3V0ZXIgZnAyMjAwLTEyIHYzLjAuMiBSZWFkeQpMb2dpbjog
+    </example>
+    <param pos="0" name="hw.vendor" value="Flowpoint"/>
```
ekelly-rapid7: Should these be `os.` params?

PR #194, opened 2018-11-15
Addition of attributes for Cisco and Windows fingerprints

jpincombe-r7:
## Description
Quick PR to add attributes that I had missed during my previous PR to the Cisco and Microsoft telnet regex's.

## How Has This Been Tested?
Running **rake tests** within recog and running **bin/recog_verify xml/telnet_banners.xml** within recog

## Types of changes

## Checklist:

### DIFF
--- a/xml/telnet_banners.xml
+++ b/xml/telnet_banners.xml
@@ -47,8 +47,6 @@
     <param pos="0" name="service.vendor" value="Cisco"/>
     <param pos="0" name="os.vendor" value="Cisco"/>
     <param pos="0" name="hw.vendor" value="Cisco"/>
-    <param pos="0" name="os.family" value="IOS"/>
-    <param pos="0" name="os.product" value="IOS"/>
   </fingerprint>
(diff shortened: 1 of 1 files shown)


xml/telnet_banners.xml
```diff
@@ -47,6 +47,8 @@
     <param pos="0" name="service.vendor" value="Cisco"/>
     <param pos="0" name="os.vendor" value="Cisco"/>
     <param pos="0" name="hw.vendor" value="Cisco"/>
+    <param pos="0" name="os.family" value="IOS"/>
+    <param pos="0" name="os.product" value="IOS"/>
```
tsellers-r7: IIRC this can match on Cisco switches as well which means the OS can be CAT OS for Cisco Catalyst gear.

jhart-r7: 👍.  We've definitely run into this before, and if it can be either, I think the pattern is to leave off product/family.

PR #195, opened 2018-11-19
Improve Squid fingerprinting

jhart-r7:
## Description

This improves the fingerprinting of Squid by:
 * Making the existing regex more case-insensitive
 * Adding a fingerprint for when Squid presents no version
 * Add a vendor to all Squid fingerprints
 * Add Squid CPE support

## Motivation and Context

Squid is a popular web cache/proxy with millions of endpoints exposed on the public Internet.

## How Has This Been Tested?

The existing examples are tested automatically.

## Types of changes

## Checklist:

### DIFF
--- a/cpe-remap.yaml
+++ b/cpe-remap.yaml
@@ -103,6 +103,8 @@ mappings:
       jboss_eap: jboss_enterprise_application_platform
       jbossweb: jboss_web_framework_kit
       red_hat_directory_server: directory_server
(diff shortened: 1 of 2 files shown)


jhart-r7:
@gwiseman-r7 agreed on Squid Cache.  The Squid website uses this name in the URL but almost nowhere else.  I made this decision purely based on CPE, which calls it "Squid Cache".

PR #198, opened 2018-12-11
Ensure consistency of os.device and hw.device when both present

jhart-r7:
## Description

Currently many fingerprints assert the device type through one or more of `os` and `hw`.  Only two use both but with inconsistent values and this seems like a mistake.  So, I've corrected the two problematic fingerprints and put a test in place to enforce this going forward.

## Motivation and Context

See description

## How Has This Been Tested?

Existing tests pass, new tests pass and were used to catch the two problematic fingerprints in the first place.

## Types of changes

## Checklist:


jhart-r7:
Wrong base, brb

PR #205, opened 2018-12-19
Improve UPnP Server fingerprinting

jhart-r7:
## Description

This adds a variety of new UPnP fingerprints is outlined in the individual commits.

## Motivation and Context

More fingerprints.

## How Has This Been Tested?

Examples for everything tested with GPRs.

## Types of changes

## Checklist:

### DIFF
--- a/xml/upnp_banners.xml
+++ b/xml/upnp_banners.xml
@@ -467,6 +467,11 @@
     <param pos="1" name="os.version"/>
     <param pos="0" name="os.device" value="Router"/>
   </fingerprint>
+  <fingerprint pattern="^TBS/R2 UPnP/[\d\.]+ MiniUPnPd/[\d\.]+$">
+    <description>D-Link generic</description>
+    <example>TBS/R2 UPnP/1.0 MiniUPnPd/1.2</example>
+    <param pos="0" name="hw.vendor" value="D-Link"/>
+  </fingerprint>
   <fingerprint pattern="^ipos/([\d\.]+) UPnP/[\d\.]+ (TL-\w+)/(\S+)$">
     <description>TP-Link WAP UPnP Server</description>
     <example>ipos/7.0 UPnP/1.0 TL-WR841N/6.0/7.0</example>
--- a/xml/upnp_banners.xml
+++ b/xml/upnp_banners.xml
@@ -1,6 +1,15 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- UPnP Server headers are matched against these patterns to fingerprint UPnP servers. -->
 <fingerprints matches="ssdp_header.server" protocol="ssdp" database_type="service" preference="0.70">
+  <fingerprint pattern="(?i)^(.*) UPnP/[\d\.]+\s+AVM FRITZ!(.*) ([\d\.]+)$">
+    <description>AVM FRITZ! devices of various types</description>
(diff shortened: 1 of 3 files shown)


xml/upnp_banners.xml
```diff
@@ -541,4 +587,20 @@
     <param pos="0" name="service.product" value="XBMC"/>
     <param pos="1" name="service.version"/>
   </fingerprint>
+  <fingerprint pattern="Synology/DSM/(\d+\.\d+\.\d+\.\d+)$">
+    <description>Synology DiskStation NAS with IP</description>
+    <example host.ip="192.168.1.100">Synology/DSM/192.168.1.100</example>
```
tsellers-r7: As a note, Synology uses a customized Linux distribution which they call DSM. That being said I don't see it being used else where except here:

https://github.com/rapid7/recog/blob/a0333ab8ab8bdecbad2bb7580b9e1d99bcaca981/xml/dns_versionbind.xml#L620-L624

Perhaps we can revisit this later and just open a ticket.

xml/upnp_banners.xml
```diff
@@ -531,6 +545,38 @@
...
+    <param pos="0" name="hw.product" value="Roku"/>
+    <param pos="0" name="hw.device" value="Media Server"/>
+    <param pos="1" name="hw.version"/>
+  </fingerprint>
+  <fingerprint pattern="^Roku UPnP/\S+ MiniUPnPd/\S+$">
+    <description>Roku without a version</description>
+    <example>Roku UPnP/1.0 MiniUPnPd/1.4</example>
+    <param pos="0" name="hw.vendor" value="Roku"/>
+    <param pos="0" name="hw.product" value="Roku"/>
+    <param pos="0" name="hw.device" value="Media Server"/>
```
tsellers-r7: I'm torn on the use of `Media Server` here.  It's just a player much like a Chromecast but perhaps we lump all such devices into this category to prevent label sprawl.

xml/upnp_banners.xml
```diff
@@ -531,6 +545,38 @@
     <param pos="1" name="os.version"/>
     <param pos="0" name="os.cpe23" value="cpe:/o:apple:mac_os_x:{os.version}"/>
   </fingerprint>
+  <fingerprint pattern="^Loxone Miniserver (.*) UPnP/1.0$">
+    <description>Loxone Miniserver Smart Home</description>
+    <example host.name="some name">Loxone Miniserver some name UPnP/1.0</example>
+    <param pos="0" name="hw.vendor" value="Loxone"/>
+    <param pos="0" name="hw.product" value="Miniserver"/>
+    <param pos="0" name="hw.device" value="Management"/>
```
tsellers-r7: `Management` seems vague. Perhaps `Smart Home Management`, `Home Automation`, etc?

xml/upnp_banners.xml
jhart-r7: Yes, my thinking was this devices like this serve media, and the fact that some play it vs others store and play it is an important distinction but one that it is outside recog currently, IMO :)

xml/upnp_banners.xml
jhart-r7: Yes, I had the same hesitation.  I think "Building Automation" or similar might work better given other work

tsellers-r7: Building Automation sounds like  a good compromise.

PR #207, opened 2018-12-20
OCTO-1870 Add Nero SIP device

pdeardorff-r7:
## Description
Add fingerprint 

## Motivation and Context
More FPs

## How Has This Been Tested?
Rspec

## Checklist:

### DIFF
--- a/xml/sip_user_agents.xml
+++ b/xml/sip_user_agents.xml
@@ -123,13 +123,13 @@
     <param pos="1" name="hw.model"/>
     <param pos="2" name="hw.version"/>
   </fingerprint>
-  <fingerprint pattern="^Nero SIPPS IP Phone Version ([\d\._]+)+$">
+  <fingerprint pattern="^Nero SIPPS IP Phone Version ([\d\.]+)+$">
     <description>Nero SIPPS IP Phone</description>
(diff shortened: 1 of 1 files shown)


xml/sip_user_agents.xml
```diff
@@ -124,4 +123,13 @@
     <param pos="1" name="hw.model"/>
     <param pos="2" name="hw.version"/>
   </fingerprint>
+  <fingerprint pattern="^Nero SIPPS IP Phone Version ([\d\._]+)+$">
+    <description>Nero SIPPS IP Phone</description>
+    <example hw.version="2.0.51.16">Nero SIPPS IP Phone Version 2.0.51.16</example>
+    <param pos="0" name="hw.vendor" value="Nero"/>
+    <param pos="0" name="hw.family" value="SIPPS"/>
+    <param pos="0" name="hw.device" value="VoIP"/>
+    <param pos="0" name="hw.product" value="SIPPS IP Phone"/>
```
tsellers-r7: Do we have other data from these IPs? I ask because this is likely a softphone as opposed to an actual hardware device.

pdeardorff-r7: Ah I think you're right, I'm not even sure if the software is offered anymore... most references to it are circa 2005.  

http://www.nero.com/enu/company/press/press-release.php?id=264

PR #210, opened 2018-12-23
Add basic support for identifying Huawei kit in cpe

brudis-r7:
## Description

While not perfect, `http_wwwauth.xml`, `ssh_banners.xml`, `telnet_banners.xml` will now enable measuring exposure of Huawei router-ish devices on well-known ports. 

## Motivation and Context

Needed primarily for CyberThreatAlliance "Edge" collab research paper but will be valuable for other research.

## How Has This Been Tested?

xmllint shows no errors and syntax matches other CPE

## Types of changes

Addition of basic CPE to Huawei signatures in three files

## Checklist:

- [X] I have updated the documentation accordingly (or changes are not required).
- [X] I have added tests to cover my changes (or new tests are not required).
- [X] All new and existing tests passed.

### DIFF
--- a/xml/http_wwwauth.xml
+++ b/xml/http_wwwauth.xml
@@ -211,6 +211,7 @@
     <param pos="0" name="os.device" value="Broadband router"/>
     <param pos="0" name="os.family" value="MT"/>
     <param pos="1" name="os.product"/>
+    <param pos="0" name="os.cpe23" value="cpe:2.3:o:huawei:-"/>
   </fingerprint>
   <fingerprint pattern="^(?:Basic|Digest) realm=&quot;HuaweiHomeGateway&quot;.*$">
     <description>Huawei Home Gateway Routers</description>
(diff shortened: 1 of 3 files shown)


jhart-r7:
Hmm, I'm all for more/better CPEs, but there are a couple of complications with this:

1. While not documented the best, the CPEs that are in here are the result of automation, not manual curation as I think this is.  We can work towards supporting manual CPE modifications in individual files but right now there is automation that keeps these all in place.  It *may* be the case that we just need to tweak or re-run this automation, because of huawei is a valid CPE, then our automation should support it.
2.  I think you are mixing CPE 2.2 and 2.3 formats in a field that is supposed to be 2.3
3. How exactly will you be using just these CPEs?  If your goal is just to identify Huawei devices, then you'd probably get much more by simply looking for `os.vendor` or `hw.vendor` for Huawei -- we have several across more fingerprint databases than just these

jhart-r7:
Sorry, there is also the issue of branching here.  Per https://github.com/rapid7/recog/blob/master/CONTRIBUTING.md, PRs need to come from a topic specific branch.  The only way to adjust this is to close this PR, rename your branch and then re-submit.  GIven the other issues with this PR, I'm going to close it now and we can discuss it elsewhere.  Thanks!

PR #214, opened 2018-12-26
Bug/golang compat

### LINKED ISSUE #209: Invalid escape sequence (unicode)
The file `telnet_banners.xml` contains a `\u0000` sequence which is invalid for certain Regexp implementations (for example, Go). This should probably be changed to `\x00` instead.

hdm:
## Description
This PR corrects a handful of minor issues that break Go regexp compatibility
* Replaces \u0000 with \x00 for null byte matching (fixes #209)
* Nested character classes for IP addresses worked with Ruby, but not Go, the fixed version works with both.
* Missing REG_MULTILINE flags on a handful of fingerprints. The author had prefixed (?m), but this isn't exactly the same thing. Ruby happens to match `.` to newline with (?m) but Go requires an explicit flag to be set.

## Motivation and Context
I have been working towards a public release of recog-go, which is a Go implementation of the recog ruby code for applying matches to service banners.

## How Has This Been Tested?
Post-change, the `recog_verify` script in ruby was used to verify that matching is working as expected. A Go test script performed a similar test as well.

### DIFF
--- a/xml/ldap_searchresult.xml
+++ b/xml/ldap_searchresult.xml
@@ -657,7 +657,7 @@
     <param pos="0" name="service.product" value="UnboundID Directory Proxy Server"/>
     <param pos="1" name="service.version"/>
   </fingerprint>
-  <fingerprint pattern="(?im:namingContexts1.\x04.cn=.?pbx.*\x04.ldapServiceName1.\x04.IPVA-\w+-)">
+  <fingerprint pattern="(?im:namingContexts1.\x04.cn=.?pbx.*\x04.ldapServiceName1.\x04.IPVA-\w+-)" flags="REG_MULTILINE">
     <description>innovaphone VoIP Gateway Virtual Appliance</description>
     <example _encoding="base64">
       Dm5hbWluZ0NvbnRleHRzMQoECGNuPUtQQlgwMCIED2xkYXBTZXJ2aWNlTmFtZTEPBA1JUFZBL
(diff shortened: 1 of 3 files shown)


xml/ldap_searchresult.xml
```diff
@@ -657,7 +657,7 @@
     <param pos="0" name="service.product" value="UnboundID Directory Proxy Server"/>
     <param pos="1" name="service.version"/>
   </fingerprint>
-  <fingerprint pattern="(?im:namingContexts1.\x04.cn=.?pbx.*\x04.ldapServiceName1.\x04.IPVA-\w+-)">
+  <fingerprint pattern="(?im:namingContexts1.\x04.cn=.?pbx.*\x04.ldapServiceName1.\x04.IPVA-\w+-)" flags="REG_MULTILINE">
```
jhart-r7: FYI, we haven't acted on this yet, but we have desires to remove support for `flags` entirely, but Go support and this specific newline problem is one of the complications we identified.  That shouldn't change what you are doing in this PR just yet, but just FYI.

Is another option to explicitly list the carriage return and/or newline characters that might separate parts of this rather than `.*`?

hdm: Specifying the newline in the match would work as well, but it might take a while to fix the existing fingerprints. Let me know if I can help either way.

hdm: Regarding newlines and matching, the Go syntax.Regexp documentation (page)[https://golang.org/src/regexp/syntax/parse.go?s=1700:1717#L43] goes into the distinction here. For the implementation in recog-go, I use REG_MULTILINE to set MatchNL (DotNL | ClassNL). The latter is needed for things like \s to match newlines.

hdm: Anther option might be to assume REG_MULTILINE is on for _all_ patterns. This seemed to break a small number of matches when I tested earlier, due to .* including whitespace, but I think the total was under 5. The REG_ICASE stuff seems to work just fine via (?i:pattern) prefixes, but (?m) gets wonky without `.` and `\s` matching newline.

hdm:
Howdy folks! Need anything from me to move this forward? I was waiting for this merge to start on the additional fingerprint and database PRs (my test stack is all Go atm).

PR #221, opened 2019-01-21
Feedback Request: First draft of x509 Subject & Issuer fingerprints

hdm:
This is a first hack at example fingerprints for x509 Subject and Issuer fields. The idea is to get feedback on what looks right, what needs work, and any thoughts for improving this before diving in for real. One of the oddities of fingerprinting the TLS field is that we don't have any information on the source service. Some of the recog match fields feel a bit off as well because of this. 

So, what do you think?

### DIFF
--- a/xml/x509_subjects.xml
+++ b/xml/x509_subjects.xml
@@ -1,0 +1,124 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<fingerprints matches="x509.subject" protocol="x509">
+  <!--
+  This fingerprint set matches the Subject field of x509 certificates. These x509
+  certificates may be sourced from any SSL or TLS service. If a particular system
+  has identical subject and issuer fields, the subject field should be preferred.
+  -->
+  <fingerprint pattern="^CN=([0-9a-f]{2}:[0-9a-f]{2}:[0-9a-f]{2}:[0-9a-f]{2}:[0-9a-f]{2}:[0-9a-f]{2}),OU=([^,]+),O=Araknis Networks \(TM\),L=Charlotte,C=US$">
+    <description>Araknis Networks Router</description>
+    <example host.mac="d4:6a:91:7a:a3:c4" hw.product="AN-300-RT-4L2W">CN=d4:6a:91:7a:a3:c4,OU=AN-300-RT-4L2W,O=Araknis Networks (TM),L=Charlotte,C=US</example>
+    <!-- no service or OS information -->
+    <!-- this fingerprint may also match WAPs and Switches, but we don't have examples -->
+    <param pos="0" name="hw.vendor" value="Araknis Networks"/>
+    <param pos="0" name="hw.family" value="Araknis Networks"/>
+    <param pos="0" name="hw.device" value="Router"/>
+    <param pos="2" name="hw.product"/>
+    <param pos="1" name="host.mac"/>
+  </fingerprint>
+
+  <fingerprint pattern="^CN=([^,]+),OU=Products,O=Cambium Networks Inc,L=San Jose,ST=CA,C=US$">
+    <description>Cambium Networks Router</description>
+    <example hw.product="R190V">CN=R190V,OU=Products,O=Cambium Networks Inc,L=San Jose,ST=CA,C=US</example>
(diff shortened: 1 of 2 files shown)


xml/x509_subjects.xml
```diff
@@ -0,0 +1,124 @@
...
+
+  <fingerprint pattern="^CN=ASA Temporary Self Signed Certificate$">
+    <description>Cisco ASA</description>
+    <example>CN=ASA Temporary Self Signed Certificate</example>
+    <param pos="0" name="os.vendor" value="Cisco"/>
+    <param pos="0" name="os.family" value="Adaptive Security Appliance"/>
+    <param pos="0" name="os.product" value="VPN"/>
+    <param pos="0" name="hw.vendor" value="Cisco"/>
+    <param pos="0" name="hw.family" value="Adaptive Security Appliance"/>
+    <param pos="0" name="hw.product" value="Adaptive Security Appliance"/>
```
jhart-r7: Why is this product different than `os`?

hdm: This matches the current fingerprints in http_servers.xml (Cisco Aware), as well as the http_cookies.xml entry. It seems weird though. I could use a primer on what these fields should be set to and how os and hw should relate for hardware appliances.

xml/x509_subjects.xml
```diff
@@ -0,0 +1,124 @@
...
+    <description>Super Micro IPMI Controller</description>
+    <example>CN=IPMI,OU=Software,O=Super Micro Computer,ST=California,C=US</example>
+    <param pos="0" name="service.vendor" value="Super Micro"/>
+    <param pos="0" name="service.product" value="IPMI"/>
+    <param pos="0" name="service.family" value="IPMI"/>
+    <param pos="0" name="hw.vendor" value="Super Micro"/>
+    <param pos="0" name="os.vendor" value="Super Micro"/>
+    <param pos="0" name="os.product" value="IPMI"/>
+    <param pos="0" name="os.family" value="IPMI"/>
+    <!-- should we standardize on BMC for all controllers? -->
```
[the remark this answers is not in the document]
hdm: 👍

xml/x509_subjects.xml
```diff
@@ -0,0 +1,124 @@
...
+
+  <fingerprint pattern="^CN=([^ ,]+),OU=Products,O=Cambium Networks Inc,L=San Jose,ST=CA,C=US$">
+    <description>Cambium Networks Router</description>
+    <example hw.product="R190V">CN=R190V,OU=Products,O=Cambium Networks Inc,L=San Jose,ST=CA,C=US</example>
+    <example hw.product="R190W">CN=R190W,OU=Products,O=Cambium Networks Inc,L=San Jose,ST=CA,C=US</example>
+    <example hw.product="R200">CN=R200,OU=Products,O=Cambium Networks Inc,L=San Jose,ST=CA,C=US</example>
+    <example hw.product="R200P">CN=R200P,OU=Products,O=Cambium Networks Inc,L=San Jose,ST=CA,C=US</example>
+    <example hw.product="R201">CN=R201,OU=Products,O=Cambium Networks Inc,L=San Jose,ST=CA,C=US</example>
+    <example hw.product="R201P">CN=R201P,OU=Products,O=Cambium Networks Inc,L=San Jose,ST=CA,C=US</example>
+    <example hw.product="R201W">CN=R201W,OU=Products,O=Cambium Networks Inc,L=San Jose,ST=CA,C=US</example>
```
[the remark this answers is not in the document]
hdm: 👍

xml/x509_subjects.xml
jhart-r7: It is a tricky area largely because there are lots of bad counterexamples in current recog that need to be cleaned up.  In general, as you might guess, `os` is for anything we can reasonably and usefully say about the operating system and `hw` is for the hardware or physical device itself.

jhart-r7:
@hdm let me know if we can assist with this effort.

hdm:
Thanks for the feedback. I had been letting this stew for a bit, but plan to circle back soon with a real PR based on your comments.

hdm:
Closing this and replacing with a real PR soon.

PR #225, opened 2019-04-04
TLS issuer and subject fingerprints

hdm:
This is a first run at fingerprints for TLS issuer and subject fields. I plan to keep updating this for a while, but would appreciate any feedback on the current state. Coverage-wise, this is about 50% of the systems that I am trying to match, but that doesn't include any Sonar scans.

### DIFF
--- a/xml/x509_issuers.xml
+++ b/xml/x509_issuers.xml
@@ -6,18 +6,18 @@
   has identical subject and issuer fields, the subject field should be preferred.
   -->
 
-    <fingerprint pattern="CN=Chromecast ICA (\d+)\s*([^,]*),OU=Cast,O=Google Inc,L=Mountain View,ST=California,C=US$">
+    <fingerprint pattern="CN=Chromecast ICA (\d+)\s*\(?([^,\)]*)\)?,OU=Cast,O=Google Inc,L=Mountain View,ST=California,C=US$">
         <description>Google Chromecast</description>
 
         <example chromecast.generation="3">CN=Chromecast ICA 3,OU=Cast,O=Google Inc,L=Mountain View,ST=California,C=US</example>
-        <example chromecast.generation="4" chromecast.capabilities="(Audio)">CN=Chromecast ICA 4 (Audio),OU=Cast,O=Google Inc,L=Mountain View,ST=California,C=US</example>
-        <example chromecast.generation="5" chromecast.capabilities="(4K)">CN=Chromecast ICA 5 (4K),OU=Cast,O=Google Inc,L=Mountain View,ST=California,C=US</example>
-        <example chromecast.generation="6" chromecast.capabilities="(Audio Assist)">CN=Chromecast ICA 6 (Audio Assist),OU=Cast,O=Google Inc,L=Mountain View,ST=California,C=US</example>
-        <example chromecast.generation="7" chromecast.capabilities="(Audio Assist 2)">CN=Chromecast ICA 7 (Audio Assist 2),OU=Cast,O=Google Inc,L=Mountain View,ST=California,C=US</example>
-        <example chromecast.generation="9" chromecast.capabilities="(Audio Assist)">CN=Chromecast ICA 9 (Audio Assist),OU=Cast,O=Google Inc,L=Mountain View,ST=California,C=US</example>
-        <example chromecast.generation="11" chromecast.capabilities="(Video Assist)">CN=Chromecast ICA 11 (Video Assist),OU=Cast,O=Google Inc,L=Mountain View,ST=California,C=US</example>
+        <example chromecast.generation="4" chromecast.capabilities="Audio">CN=Chromecast ICA 4 (Audio),OU=Cast,O=Google Inc,L=Mountain View,ST=California,C=US</example>
+        <example chromecast.generation="5" chromecast.capabilities="4K">CN=Chromecast ICA 5 (4K),OU=Cast,O=Google Inc,L=Mountain View,ST=California,C=US</example>
+        <example chromecast.generation="6" chromecast.capabilities="Audio Assist">CN=Chromecast ICA 6 (Audio Assist),OU=Cast,O=Google Inc,L=Mountain View,ST=California,C=US</example>
+        <example chromecast.generation="7" chromecast.capabilities="Audio Assist 2">CN=Chromecast ICA 7 (Audio Assist 2),OU=Cast,O=Google Inc,L=Mountain View,ST=California,C=US</example>
+        <example chromecast.generation="9" chromecast.capabilities="Audio Assist">CN=Chromecast ICA 9 (Audio Assist),OU=Cast,O=Google Inc,L=Mountain View,ST=California,C=US</example>
+        <example chromecast.generation="11" chromecast.capabilities="Video Assist">CN=Chromecast ICA 11 (Video Assist),OU=Cast,O=Google Inc,L=Mountain View,ST=California,C=US</example>
         <example chromecast.generation="12">CN=Chromecast ICA 12,OU=Cast,O=Google Inc,L=Mountain View,ST=California,C=US</example>
-        
+
         <param pos="0" name="hw.device" value="Media Server"/>
         <param pos="0" name="hw.vendor" value="Google"/>
         <param pos="0" name="hw.product" value="Chromecast"/>
--- a/xml/x509_issuers.xml
+++ b/xml/x509_issuers.xml
@@ -101,12 +101,12 @@
     <param pos="0" name="hw.product" value="TelePresence"/>
   </fingerprint>
   <fingerprint pattern="^O=VMware Installer$">
-    <description>VMWare ESX</description>
+    <description>VMWare ESXi w/Installer</description>
     <example>O=VMware Installer</example>
     <param pos="0" name="os.vendor" value="VMWare"/>
-    <param pos="0" name="os.product" value="ESX"/>
+    <param pos="0" name="os.product" value="ESXi"/>
     <param pos="0" name="os.device" value="Hypervisor"/>
-    <param pos="0" name="os.cpe23" value="cpe:/o:vmware:esx:-"/>
+    <param pos="0" name="os.cpe23" value="cpe:/o:vmware:esxi:-"/>
   </fingerprint>
   <fingerprint pattern="^CN=CA,OU=VMware Engineering,O=vCenter,ST=California,C=US$">
     <description>VMWare vCenter</description>
--- a/xml/x509_issuers.xml
+++ b/xml/x509_issuers.xml
@@ -6,7 +6,8 @@
   has identical subject and issuer fields, the subject field should be preferred.
   -->
 
-    <fingerprint pattern="CN=Chromecast ICA (\d+)\s*\(?([^,\)]*)\)?,OU=Cast,O=Google Inc,L=Mountain View,ST=California,C=US$">
+    <!-- Chromecast and various devices that support the Cast protocol -->
+    <fingerprint pattern="^CN=Chromecast ICA (\d+)\s*\(?([^,\)]*)\)?,OU=Cast,O=Google Inc,L=Mountain View,ST=California,C=US$">
         <description>Google Chromecast</description>
 
         <example chromecast.generation="3">CN=Chromecast ICA 3,OU=Cast,O=Google Inc,L=Mountain View,ST=California,C=US</example>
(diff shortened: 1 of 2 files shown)


hdm:
I should have the rspec failures sorted out later today.

tsellers-r7:
Thanks HD. I'll try to take a look at this later today. @ me to catch my attention earlier

hdm:
@tsellers-r7 thanks! it should be good for a first review

xml/x509_subjects.xml
```diff
@@ -0,0 +1,1019 @@
...
+	<fingerprint pattern="^CN=IPMI,OU=Software,O=Super Micro Computer,ST=California,C=US$">
+        <description>Super Micro IPMI Controller</description>
+        <example>CN=IPMI,OU=Software,O=Super Micro Computer,ST=California,C=US</example>
+        <param pos="0" name="hw.device" value="Lights Out Management"/>
+        <param pos="0" name="hw.vendor" value="Super Micro"/>
+        <param pos="0" name="hw.product" value="IPMI"/>
+    </fingerprint>
+
+
+    <!-- iDRAC with a MAC -->
```
jhart-r7: for these (and perhaps other fingerprint groups) it would be better to put this differentiation in the `description` field as it will help more readily identify the difference between two fingerprints that would otherwise just be described as "Dell iDRAC Remote Access Controller"

xml/x509_subjects.xml
```diff
@@ -0,0 +1,1019 @@
...
+    <!-- iDRAC with real service tag -->
+	<fingerprint pattern="^CN=idrac-([A-Z0-9]{7}),OU=Remote Access Group,O=Dell Inc\.,L=Round Rock,ST=Texas,C=US$">
+        <description>Dell iDRAC Remote Access Controller</description>
+        <example dell.service_tag="JXJJC02">CN=idrac-JXJJC02,OU=Remote Access Group,O=Dell Inc.,L=Round Rock,ST=Texas,C=US</example>
+        <example dell.service_tag="FMDSJH3">CN=idrac-FMDSJH3,OU=Remote Access Group,O=Dell Inc.,L=Round Rock,ST=Texas,C=US</example>
+        <param pos="0" name="hw.device" value="Lights Out Management"/>
+        <param pos="0" name="hw.vendor" value="Dell"/>
+        <param pos="0" name="hw.product" value="iDRAC"/>
+        <param pos="1" name="dell.service_tag"/>
+    </fingerprint>
```
jhart-r7: minor, here and elsewhere there is some inconsistent indentation.

xml/x509_issuers.xml
```diff
@@ -0,0 +1,128 @@
...
+    
+	<fingerprint pattern="^CN=Asus fugu Cast ICA,OU=Widevine,O=Google Inc,L=Kirkland,ST=Washington,C=US$">
+        <description>ASUS Nexus Player (Android) with Google Cast</description>
+        <example>CN=Asus fugu Cast ICA,OU=Widevine,O=Google Inc,L=Kirkland,ST=Washington,C=US</example>
+        <param pos="0" name="os.vendor" value="Google" />
+        <param pos="0" name="os.family" value="Linux" />
+        <param pos="0" name="os.product" value="Android" />
+        <param pos="0" name="hw.device" value="Media Server" />
+        <param pos="0" name="hw.vendor" value="ASUS" />
+        <param pos="0" name="hw.product" value="Nexus Player" />
```
jhart-r7: minor, but the style regarding spaces before the close of a tag seems to vary in this file.  I'd suggest a `xmllint --format`

xml/x509_issuers.xml
```diff
@@ -0,0 +1,128 @@
...
+    </fingerprint>
+    
+	<fingerprint pattern="^CN=[a-zA-Z0-9]+,OU=Internally Generated Certificate,O=American Power Conversion Corp,L=Default Locality,ST=Default State,C=US$">
+        <description>APC UPS</description>
+        <example>CN=ZA1117619249,OU=Internally Generated Certificate,O=American Power Conversion Corp,L=Default Locality,ST=Default State,C=US</example>
+        <param pos="0" name="hw.device" value="Power Device" />
+        <param pos="0" name="hw.vendor" value="APC" />
+    </fingerprint>
+    
+	<fingerprint pattern="^CN=Temporary CA [a-fA-F0-9]{8}\-[a-fA-F0-9]{4}\-[a-fA-F0-9]{4}\-[a-fA-F0-9]{4}\-[a-fA-F0-9]{12},OU=Temporary CA">
```
jhart-r7: Anchor at end with `$`?

xml/x509_issuers.xml
```diff
@@ -0,0 +1,128 @@
...
+        <param pos="0" name="os.product" value="ChromeOS" />
+        <param pos="0" name="hw.device" value="Media Server" />
+        <param pos="0" name="hw.vendor" value="Google" />
+        <param pos="0" name="hw.product" value="Chromecast" />
+        <param pos="0" name="hw.cpe23" value="cpe:/a:google:chromecast:-" />
+        <param pos="1" name="chromecast.generation" />
+        <param pos="2" name="chromecast.capabilities" />
+    </fingerprint>
+    
+	<fingerprint pattern="^CN=Asus fugu Cast ICA,OU=Widevine,O=Google Inc,L=Kirkland,ST=Washington,C=US$">
```
jhart-r7: Can you comment on how you are generating these DN strings?  I thought they were going to be ordered alphabetically by field name but that does not appear to be the case.  If there is some other order we should standardize around then we should discuss it.

hdm: These are the Issuer and Subject functions of https://golang.org/pkg/crypto/x509/#Certificate

hdm:
The feedback provided so far should be resolved, thanks!

xml/x509_issuers.xml
jhart-r7: The problem is that [RFC 2253](http://www6.ietf.org/rfc/rfc2253.txt) doesn't define an order:

>    When converting from an ASN.1 RelativeDistinguishedName to a string,
   the output consists of the string encodings of each
   AttributeTypeAndValue (according to 2.3), in any order.

IMO we should either agree upon a specific order and list the field names and their order in these XML files as a comment or require that the field names be ordered alphabetically.  

I agree that the format used by Go is the most common and will probably work for many other languages and would likely result in little/no conversion in most cases, however.

hdm: Thanks! I added comments for now.

xml/x509_issuers.xml
```diff
@@ -0,0 +1,131 @@
...
+    <param pos="0" name="hw.vendor" value="Cisco"/>
+    <param pos="0" name="hw.product" value="TelePresence"/>
+  </fingerprint>
+  <fingerprint pattern="^O=VMware Installer$">
+    <description>VMWare ESX</description>
+    <example>O=VMware Installer</example>
+    <param pos="0" name="os.vendor" value="VMWare"/>
+    <param pos="0" name="os.product" value="ESX"/>
+    <param pos="0" name="os.device" value="Hypervisor"/>
+    <param pos="0" name="os.cpe23" value="cpe:/o:vmware:esx:-"/>
```
tsellers-r7: Any indication that this is ESX `cpe:/o:vmware:esx:-` and not ESXi `cpe:/o:vmware:esxi:-`?

As a note, once this lands I'll try to add the ESXi host from my lab.

xml/x509_subjects.xml
```diff
@@ -0,0 +1,950 @@
...
+    <param pos="0" name="hw.vendor" value="Cisco"/>
+    <param pos="0" name="hw.product" value="Adaptive Security Appliance"/>
+    <param pos="0" name="hw.cpe23" value="cpe:/h:cisco:adaptive_security_appliance:-"/>
+  </fingerprint>
+  <fingerprint pattern="^CN=synology\.com.*,O=Synology Inc\.,L=Taipei.*,C=TW$">
+    <description>Synology NAS</description>
+    <example>CN=synology.com,OU=FTP Team,O=Synology Inc.,L=Taipei,ST=Taiwan,C=TW</example>
+    <example>CN=synology.com,O=Synology Inc.,L=Taipei,C=TW</example>
+    <param pos="0" name="hw.vendor" value="Synology"/>
+    <param pos="0" name="hw.family" value="DiskStation"/>
```
tsellers-r7: I wonder if we can find an example of this from one of their routers to see if this pattern is `Diskstation` only.

xml/x509_subjects.xml
```diff
@@ -0,0 +1,950 @@
...
+    <param pos="0" name="hw.device" value="Firewall"/>
+    <param pos="0" name="hw.product" value="PoliWall"/>
+  </fingerprint>
+  <fingerprint pattern="^CN=pfSense-[a-zA-Z0-9]+,O=pfSense webConfigurator Self-Signed Certificate">
+    <description>pfSense Firewall</description>
+    <example>CN=pfSense-58fb5b0b06777,O=pfSense webConfigurator Self-Signed Certificate,L=Locality,ST=State,C=US</example>
+    <example>CN=pfSense-5c76251e4d3e4,O=pfSense webConfigurator Self-Signed Certificate</example>
+    <param pos="0" name="hw.vendor" value="pfSense"/>
+    <param pos="0" name="hw.device" value="Firewall"/>
+    <param pos="0" name="hw.product" value="Firewall"/>
```
tsellers-r7: I think the pfSense OS is FreeBSD for all of their gear.

xml/x509_subjects.xml
hdm: The security gateways use as TLS subject like `CN=UBNT Router UI,O=Ubiquiti Networks,L=San Jose,ST=CA,C=US` instead.

xml/x509_issuers.xml
hdm: These were indeed ESXi not ESX, updating.

xml/x509_subjects.xml
hdm: Thanks, updating (default ttl of 64 made it look like Linux at first)

tsellers-r7:
@hdm - Unless you have anything you'd like to add to this before we land this I expect we will land this and release this afternoon.

hdm:
Excellent, thanks! I have piles of more fingerprints, but happy to let this one land first.

tsellers-r7:
Up to you

hdm:
Probably good to have this land first and then keep rolling things up for a bigger future PR (following any standardization changes were making, etc).

xml/x509_subjects.xml
```diff
@@ -0,0 +1,953 @@
...
+        }
+
+    All names are separated by commas and any commas inside a name are escaped with a
+    single backslash character. See RFC 2253 for additional details on formatting.
+
+    Practically, most Subjects start with the Common Name (CN=) and then step through
+    Organization Unit (OU), Organization (O), and then some level of location, but
+    typically Locality (L) and Country (C). Names are guaranteed to be listed in
+    the order described above, but may start at any point in the list. For example,
+    Subjects may start with a Serial Number (SERIALNUMBER=) or even Extra Names, but
```
jhart-r7: Somewhat unrelated to this PR but do you know why the serial numbers are not in every DN?  My understanding is that each certificate must have a serial number.  Is it just that the serial number is not normally part of the DN but some implementations also put it there?

tsellers-r7: If I understand correctly the text above isn't the TLS certificate number, but instead a logical label applied to the context similar to Name, OU, etc.

xml/x509_subjects.xml
```diff
@@ -0,0 +1,953 @@
...
+    <param pos="0" name="hw.vendor" value="Aruba Networks"/>
+    <param pos="0" name="hw.device" value="Wireless Controller"/>
+    <param pos="0" name="hw.product" value="Captive Portal"/>
+  </fingerprint>
+  <fingerprint pattern="^CN=Fireware web CA,OU=Fireware,O=WatchGuard$">
+    <description>WatchGuard Firewall</description>
+    <example>CN=Fireware web CA,OU=Fireware,O=WatchGuard</example>
+    <param pos="0" name="hw.vendor" value="WatchGuard"/>
+    <param pos="0" name="hw.device" value="Firewall"/>
+    <param pos="0" name="os.vendor" value="WatchGuard"/>
```
jhart-r7: `<param pos="0" name="os.cpe23" value="cpe:/o:watchguard:fireware:-"/>`?  Here and elsewhere in these files there appear to be some CPE issues.  How did you add the CPEs?

hdm: Mostly guess work where I happened to see a similar cpe somewhere else. Happy to remove them all and let the scripts sort it out, if that is a better approach.

xml/x509_subjects.xml
jhart-r7: Yes that aligns with the one example in these files.

xml/x509_subjects.xml
jhart-r7: You should be able to use the script against just these files and it will remove/correct CPEs.  It isn't perfect but should correct most of the issues related to CPEs being introduced here.  We can also do this when landing, which ever you prefer.

hdm: If yall could handle this while landing, I would appreciate it, thanks!

jhart-r7: 👍

PR #227, opened 2019-04-30
RCG-30: Description deduplication

tsellers-r7:
## Description
This PR:

- Reduces the number of duplicate `description` fields. Duplicates in this field can cause issues when processing large number of banners and generating metrics on the fingerprint matches.
- Cleans up a few multi-line `description` fields, these are undesirable for similar metrics reasons.
- Adds a few `service.version` tests for `example` fields. Every regex should have every extracted value tested at least once. One of these new tests found an error in the regex for a fingerprint.

Expect a few more of these as time permits. I'm trying to avoid asking folks to review a wall of changes all at once.

## Motivation and Context
These changes improve the usability of `recog`

## How Has This Been Tested?
`rspec`

## Checklist:

### DIFF
--- a/xml/x509_subjects.xml
+++ b/xml/x509_subjects.xml
@@ -72,7 +72,7 @@
     <param pos="1" name="host.mac"/>
   </fingerprint>
   <fingerprint pattern="^SERIALNUMBER=PID:([^ ]+) SN:([^,]+),CN=(?:[a-zA-Z0-9\-]+)-SEP([a-fA-F0-9]{12}),OU=CTG,O=Cisco Systems Inc\.$">
-    <description>Cisco / Linksys Router</description>
+    <description>Cisco / Linksys Router with serial number</description>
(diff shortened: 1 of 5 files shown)


jhart-r7:
Does this fix all duplication?  If so, we should put a test in place to enforce it.

tsellers-r7:
@jhart-r7  - No. Expect a few more of these as time permits. I'm trying to avoid asking folks to review a wall of changes all at once.

xml/x509_subjects.xml
```diff
@@ -878,7 +878,7 @@
     <param pos="0" name="os.device" value="WAP"/>
   </fingerprint>
   <fingerprint pattern="^CN=axis-([a-fA-F0-9]{12}),O=Axis Communications AB$">
-    <description>OpenWRT WAP</description>
+    <description>OpenWRT WAP - Axis Communications variant</description>
```
jhart-r7: I suspect this was a copy-past problem from the fingerprint above.  I don't believe this is OpenWRT.

tsellers-r7: Thanks, I should have caught that. I caught similar in one or two other places.

hdm:
Thanks for catching it!

PR #228, opened 2019-04-30
RCG-30: Description / Example cleanup

tsellers-r7:
## Description
This PR is very similar to the previous PR.  It:

- Reduces the number of duplicate `description` fields. Duplicates in this field can cause issues when processing large number of banners and generating metrics on the fingerprint matches.
- Cleans up a few multi-line `description` fields, these are undesirable for similar metrics reasons.
- Adds a few `service.version` tests for `example` fields. Every regex should have every extracted value tested at least once. One of these new tests found an error in the regex for a fingerprint.
- Adds some examples and relocates some examples from the `description` field.

Most of the remaining changes are in the `snmp_sysdescr.xml` and I'll handle them as well as the `rspec` tests in the next PR. 

## Motivation and Context
These changes improve the usability of `recog`

## How Has This Been Tested?
`rspec`

## Checklist:

### DIFF
--- a/xml/apache_os.xml
+++ b/xml/apache_os.xml
@@ -37,7 +37,7 @@
     <param pos="0" name="os.cpe23" value="cpe:/o:mandriva:linux:-"/>
   </fingerprint>
   <fingerprint pattern=".*\(Mandrakelinux/.*">
-    <description>Mandriva (formerly Mandrake) Linux unknown version</description>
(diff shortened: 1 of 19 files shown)


tsellers-r7:
Note for self: `grep description *.xml | sort | uniq -c | sort -k 1 -brn | grep -v "   1 "`

PR #231, opened 2019-05-02
Add examples to http_wwwauth.xml

tsellers-r7:
## Description
I'm submitting this PR since I happened to have the data handy.  This PR only updates `xml/http_wwwauth.xml` and makes the following changes:

- Adds some missing examples
- Adds 1 fingerprint for SpeedTouch and reworks the WRT54G fingerprint to be more global
- Replaces the wildcard `.` with `&quot;` where appropriate in the regex
- Adds `hw.*` params in select cases where I was already touching a fingerprint.
- Moves `Fujitsu Siemens Primergy with BMC RemoteView on an iRMC card` lower in the file behind more frequently seen matches.

## Motivation and Context
Improved ability test.

## How Has This Been Tested?
`rspec`

## Types of changes

## Checklist:

### DIFF
--- a/xml/http_wwwauth.xml
+++ b/xml/http_wwwauth.xml
@@ -18,7 +18,7 @@
     <param pos="0" name="hw.vendor" value="Cisco"/>
     <param pos="0" name="hw.device" value="Router"/>
   </fingerprint>
-  <fingerprint pattern="^(?:Basic|Digest) realm=.level[ _]15[ _]or[ _]view[ _]access.$">
+  <fingerprint pattern="^(?:Basic|Digest) realm=&quot;level[ _]15[ _]or[ _]view[ _]access&quot;$">
     <description>Cisco IOS 12.x - view access variant</description>
     <example>Basic realm="level_15 or view_access"</example>
     <example>Basic realm="level_15_or_view_access"</example>
@@ -36,7 +36,7 @@
     <param pos="0" name="hw.vendor" value="Cisco"/>
     <param pos="0" name="hw.device" value="Router"/>
   </fingerprint>
-  <fingerprint pattern="^(?:Basic|Digest) realm=.level[ _]\d\d?[ _]access.$">
+  <fingerprint pattern="^(?:Basic|Digest) realm=&quot;level[ _]\d\d?[ _]access&quot;$">
     <description>Cisco IOS 12.x</description>
     <example>Basic realm="level_15_access"</example>
     <example>Basic realm="level 15 access"</example>
@@ -59,7 +59,7 @@
     <example>Basic realm="Cisco_CCSP_CWMP_TCPCR"</example>
     <param pos="0" name="hw.vendor" value="Cisco"/>
   </fingerprint>
-  <fingerprint pattern="^(?:Basic|Digest) realm=.FW-1.      Reason: no user      Server .$">
+  <fingerprint pattern="^(?:Basic|Digest) realm=&quot;FW-1.      Reason: no user      Server &quot;$">
     <description>Check Point FireWall-1</description>
     <example>Basic realm="FW-1.      Reason: no user      Server "</example>
     <param pos="0" name="service.vendor" value="Check Point"/>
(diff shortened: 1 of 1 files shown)


xml/http_wwwauth.xml
```diff
@@ -38,9 +33,13 @@
     <param pos="0" name="os.product" value="IOS"/>
     <param pos="0" name="os.version" value="12"/>
     <param pos="0" name="os.cpe23" value="cpe:/o:cisco:ios:12"/>
+    <param pos="0" name="hw.vendor" value="Cisco"/>
+    <param pos="0" name="hw.device" value="Router"/>
   </fingerprint>
   <fingerprint pattern="^(?:Basic|Digest) realm=.level[ _]\d\d?[ _]access.$">
```
jhart-r7: Why no quote change here?

xml/http_wwwauth.xml
```diff
@@ -1,18 +1,9 @@
...
-    <param pos="0" name="service.vendor" value="Fujitsu Siemens"/>
-    <param pos="0" name="service.product" value="RemoteView"/>
-    <param pos="0" name="service.family" value="RemoteView"/>
-    <param pos="0" name="hw.vendor" value="Fujitsu Siemens"/>
-    <param pos="0" name="hw.family" value="Primergy"/>
-    <param pos="0" name="hw.product" value="Primergy"/>
-    <param pos="1" name="host.name"/>
-  </fingerprint>
-  <fingerprint pattern="^(?:Basic|Digest) realm=.access.$">
+  <fingerprint pattern="^(?:Basic|Digest) realm=&quot;access&quot;$">
```
jhart-r7: Hmm, will it always be double quotes?  Perhaps that it is why historically `.` has been used to may be match single or double quotes?

tsellers-r7: I should have worded that differently. I had replaced them opportunistically.  I've gone back now and replaced them all. Thanks for the feedback.

xml/http_wwwauth.xml
tsellers-r7: I should have worded that differently. I had replaced them opportunistically. I've gone back now and replaced them all. Thanks for the feedback.

xml/http_wwwauth.xml
tsellers-r7: Good point. We actually have examples of this in our data set but none that are currently covered by the existing database.

tsellers-r7: And I immediately see that I'm wrong.

tsellers-r7: Should be sorted out now

xml/http_wwwauth.xml
```diff
@@ -167,18 +170,35 @@
...
     <param pos="0" name="os.device" value="Broadband router"/>
     <param pos="0" name="os.family" value="SpeedTouch"/>
     <param pos="0" name="os.product" value="SpeedTouch"/>
+    <param pos="0" name="hw.vendor" value="Thomson"/>
+    <param pos="0" name="hw.family" value="SpeedTouch"/>
+    <param pos="0" name="hw.device" value="Broadband router"/>
     <param pos="1" name="host.mac"/>
   </fingerprint>
-  <fingerprint pattern="^(?:Basic|Digest) realm=.SpeedTouch., nonce=.[0-9A-Z]+:([0-9A-F]{2}-[0-9A-F]{2}-[0-9A-F]{2}-[0-9A-F]{2}-[0-9A-F]{2}-[0-9A-F]{2}):\d+:\d+., qop=.auth.$">
+  <fingerprint pattern="^(?:Basic|Digest) realm=&quot;SpeedTouch., nonce=.[0-9A-Z]+:([0-9A-F]{2}-[0-9A-F]{2}-[0-9A-F]{2}-[0-9A-F]{2}-[0-9A-F]{2}-[0-9A-F]{2}):\d+:\d+., qop=.auth&quot;$">
```
jhart-r7: This one and the one above it are identical with the exception of the `qop` part which isn't reflected in the resulting fingerprint anyway.  Combine into one.

tsellers-r7: Yup, looks like one was derived from Digest and the other basic.

PR #236, opened 2019-06-17
FTP: stop matching empty banners

tsellers-r7:
## Description
This PR adjusts an FTP fingerprint so as to no longer match on empty strings.

#### Before
```shell
echo '' |  ./bin/recog_match xml/ftp_banners.xml 

MATCH: {"matched"=>"Generic FTP fingerprint without a hostname", "service.protocol"=>"ftp", "fingerprint_db"=>"ftp.banner", "data"=>""}
```

#### After
```shell
echo '' |  ./bin/recog_match xml/ftp_banners.xml 
FAIL: 
```

The trade off is that we no longer match `Server Ready`. Given that this banner is not product or service specific I consider this an acceptable trade off. Our most Sonar data set from May 1 showed 0 hits for this banner.

## How Has This Been Tested?
rspec

## Types of changes

## Checklist:

### DIFF
--- a/xml/ftp_banners.xml
+++ b/xml/ftp_banners.xml
@@ -1359,12 +1359,11 @@ more text</example>
     <param pos="1" name="host.name"/>
     <param pos="2" name="service.version"/>
   </fingerprint>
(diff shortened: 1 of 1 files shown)


tsellers-r7:
As a note, I checked some of the other more common Recog databases and didn't see any similar behavior matching empty strings.

PR #239, opened 2019-10-11
Add patterns for vxworks FTP

ekelly-rapid7:
## Description
A number of new patterns, as well as updating some others, to match a number of vxworks FTP banner messages. 

## Motivation and Context
Finding more VxWorks Operating Systems.

## How Has This Been Tested?
Added examples.
Ran `rake test`.

## Types of changes

## Checklist:

### DIFF
--- a/xml/ftp_banners.xml
+++ b/xml/ftp_banners.xml
@@ -678,7 +678,7 @@ more text</example>
     <param pos="1" name="os.version"/>
     <param pos="0" name="os.cpe23" value="cpe:/o:windriver:vxworks:-{os.version}"/>
   </fingerprint>
-  <fingerprint pattern="^[\w&lt;&gt;]+\s*Tenor\sMultipath\sSwitch\sFTP\sserver\s\(Version VxWorks([\d\.]+)\) ready\.$" flags="REG_ICASE">
+  <fingerprint pattern="^[\w&lt;&gt;]+\s*Tenor Multipath Switch FTP server \(Version VxWorks([\d\.]+)\) ready\.$" flags="REG_ICASE">
     <description>VxWorks on Tenor MultiPath with version information</description>
     <example os.version="5.4.2"><![CDATA[<38785ca0>  Tenor Multipath Switch FTP server (Version VxWorks5.4.2) ready.]]></example>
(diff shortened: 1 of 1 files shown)


xml/ftp_banners.xml
```diff
@@ -668,6 +669,32 @@ more text</example>
...
+  <fingerprint pattern="^[\w\-\.]* FTP server \((?:VxWorks\s?)+([\d\.]+)\) ready.$" flags="REG_ICASE">
+    <description>VxWorks 6 with version information</description>
+    <example os.version="6.6">NanoDAC FTP server (VxWorks VxWorks 6.6) ready.</example>
+    <example os.version="6.4">BVS-MR-BSC2 FTP server (VxWorks 6.4) ready.</example>
+    <param pos="0" name="os.vendor" value="Wind River"/>
+    <param pos="0" name="os.product" value="VxWorks"/>
+    <param pos="1" name="os.version"/>
+    <param pos="0" name="os.cpe23" value="cpe:/o:windriver:vxworks:-{os.version}"/>
+  </fingerprint>
+  <fingerprint pattern="^[\w&lt;&gt;]+\s*Tenor\sMultipath\sSwitch\sFTP\sserver\s\(Version VxWorks([\d\.]+)\) ready\.$" flags="REG_ICASE">
```
tsellers-r7: Why `\s` here for single spaces? Is it to allow for tabs?

ekelly-rapid7: Ah no I was just keeping it consistent with the initial `\s`. I'll change it to single spaces to be readable.

xml/ftp_banners.xml
```diff
@@ -668,6 +669,32 @@ more text</example>
...
+    <description>VxWorks on Tenor MultiPath with version information</description>
```
tsellers-r7: Is this hardware specific, can we add hw.product, etc?

ekelly-rapid7: This is from sonar data, I wouldn't be confident of getting the correct hw.product etc set. For this I just wanted to make sure we got the vxworks version, and fairly confident from the banner alone that this will be correct.

PR #243, opened 2019-10-25
Add a few ftp and ssh banners and a little cleanup

rhodgman-r7:
## Description
This PR cleans up a few banners and adds some new ones:

* ftp:
  * 3 ProFTPD variants
  * 1 TP-Link banner
  * 3 banners for FTP services providers (Amazon, DreamHost, home.pl)
* ssh:
  * OpenSSH for Windows
  * WeOnlyDo

## Motivation and Context
These banners provide more coverage for several banners seen in the wild that have not already been fingerprinted.

## How Has This Been Tested?
I ran `rspec`, `rake tests` and `recoc_verify` none of which reported errors.

## Types of changes

## Checklist:
- [x] I have updated the documentation accordingly (or **changes are not required**).
- [x] I have added tests to cover my changes (or **new tests are not required**).

### DIFF
--- a/xml/ssh_banners.xml
+++ b/xml/ssh_banners.xml
@@ -1150,6 +1150,9 @@
     <param pos="0" name="os.product" value="RouterOS"/>
     <param pos="0" name="os.cpe23" value="cpe:/o:mikrotik:routeros:-"/>
   </fingerprint>
+  <!-- xlightftpd is an ftp server that also supports SFTP. The SFTP
+       server appears in ssh studies, thus this banner is here, and
+       not in ftp_banners.xml-->
   <fingerprint pattern="^xlightftpd_release_([\d.]+)$">
     <description>Xlight FTP Server</description>
     <example service.version="3.8.3.6.1">xlightftpd_release_3.8.3.6.1</example>
--- a/xml/ssh_banners.xml
+++ b/xml/ssh_banners.xml
@@ -644,6 +644,18 @@
     <param pos="0" name="os.family" value="Linux"/>
     <param pos="0" name="os.product" value="HipServ"/>
   </fingerprint>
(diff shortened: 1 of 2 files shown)


xml/ssh_banners.xml
```diff
@@ -1129,17 +1141,18 @@
     <param pos="0" name="os.product" value="Tru64 Unix"/>
     <param pos="0" name="os.cpe23" value="cpe:/o:hp:tru64:-"/>
   </fingerprint>
-  <fingerprint pattern="^(?:SSH-(\d\.\d)-)?ROSSSH$">
+  <fingerprint pattern="^ROSSSH$">
     <description>MikroTik RouterOS sshd</description>
```
tsellers-r7: Verified in a sample data set that Roy's changes are correct.  The raw banner that we get back is `SSH-2.0-ROSSSH` and the processed banner that Recog sees is `ROSSSH`

gschneider-r7: doesn't this differ by implementation of Recog consumer? e.g. I think raw banners are used for recog matching in Nexpose and InsightVM

rhodgman-r7: This was the only banner in the entire `ssh_banners.xml` file that had an optional `SSH-x.x` prefix, so to keep this fingerprint in line with all the others, we decided to take it out.

Also the comment at the very top of `ssh_banners.xml` indicates that these fingerprints should be for the part of the banner after that prefix, so it looked ok to take this out.

https://github.com/rapid7/recog/blob/master/xml/ssh_banners.xml#L3-L7

```
  <!--
  SSH "software revision and comment" strings (official RFC nomenclature for the part of
  the identification string after "SSH-x.x-") are matched against these patterns to
  fingerprint SSH servers.
  -->
```

PR #259, opened 2020-04-10
Ruby gem touchups

todb-r7:
## Description
This fixes Gemfile, .ruby-version, and .ruby-gemset in order to encourage a common Ruby developer environment. I arbitrarily chose Ruby 2.6.5, which is what metasploit-framework is at right now.

## Motivation and Context
In order to keep gemsets and Ruby versions consistent across developers and to make sure tests run consistently, we should encourage a common developer environment.

## How Has This Been Tested?
`cd . && bundle install && rake spec`

Output:

```
Pending: (Failures listed here are expected and do not affect your suite's status)

  1) Recog::Fingerprint value interpolation
     # No reason given
     # ./spec/lib/recog/fingerprint_spec.rb:109


Finished in 53.16 seconds (files took 14.53 seconds to load)
131815 examples, 0 failures, 1 pending

Randomized with seed 50949

Coverage report generated for RSpec to /home/todb/git/recog/coverage. 916 / 980 LOC (93.47%) covered.
```

## Types of changes

If you're running some local hodgepodge of gems, then this will break your dev environment. After all, I've removed some lines from `.gitignore`.

## Checklist:

- [ x ] I have updated the documentation accordingly (or changes are not required).

This doesn't seem to warrant docs or tests.

- [ x ] I have added tests to cover my changes (or new tests are not required).

This doesn't seem to warrant docs or tests.

- [ x ] All new and existing tests passed.

Since there was no way to say what old Ruby version was expected, it's hard to say if the Pending failure is expected or not.


gschneider-r7:
Probably should update travis ruby versions while you're at it 😉

tsellers-r7:
@todb-r7 - I'll toss a review on it after I double check some internal code that leverages recog.

busterb:
Seems this PR kinda lost stream, will push a few extra updates to it.

tsellers-r7:
FYI, I'l review this and likely land today.

bcook-r7:
Thanks @todb-r7. There's a force push here just because I pushed a commit bumping a few other dependencies, and subsequently reverted it. Unfortunately, Travis tests showed there are quite a few things broken if we update to newer than some of the pinned gems, but fixing them requires more changes than probably makes sense for this PR.

todb-r7:
Really? Huh. Travis says the checks all pass.... would love a fail in there.

bcook-r7:
Easy peasey, just modify Gemfile, remove the pinned versions, do a 'bundle update', then run 'rake spec'. About 140 tests will fail, probably because they're using incorrect regexes.

todb-r7:
But, but, that's the point of committing a Gemfile, so you don't run into exactly that problem!

busterb:
Agree, that's convenient for a while. It's probably time to address the problems here though.The fixes were pretty simple in the end (all in the test harnesses), and according to Snyk we get at least one security vulnerability fixed as a bonus: https://app.snyk.io/vuln/SNYK-RUBY-CUCUMBER-20442 

Note that automatic vuln checking only works if you check in Gemfile.lock, which might also be worth considering.

PR #265, opened 2020-05-17
Add the initial "known identifiers" lists.

hdm:
The goal is to drive standardization of the hw, os, product, service, and vendor fields across Recog.

## Description

This adds a new directory full of text files. These lists were contributed by Rapid7 with sign-off from Brent Cook and converted from CSV to per-column text files. This is the initial work needed to begin standardization around known identifiers across the platform. 

## Motivation and Context

Having known identifiers formalized will make it easier to keep new fingerprints aligned with expected values and prevent changes to existing fingerprints from breaking products that depend on these identifiers.

## How Has This Been Tested?

It's a list of lists =D

## Types of changes
 - Documentation and known identifier lists
 - Adds bin/recog_standardize to compare against identifiers

## Checklist:

### DIFF
--- a/bin/recog_standardize
+++ b/bin/recog_standardize
@@ -1,0 +1,118 @@
+#!/usr/bin/env ruby
+
+$:.unshift(File.expand_path(File.join(File.dirname(__FILE__), "..", "lib")))
+require 'optparse'
+require 'ostruct'
+require 'recog'
+
+def load_identifiers(path)
+  res = {}
+  File.readlines(path).map{|line| line.strip}.each do |ident|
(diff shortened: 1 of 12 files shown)


tsellers-r7:
@hdm - Should be close this PR in favor of #266? It seems #265 is an subset of the commits in #266

hdm:
Either way works here; if the first example in #266 is good enough, feel free to merge that and close this one. If you think this idea needs more work, maybe merge #265 and comment/close #266 instead.

hdm:
Closing this in favor of #266

PR #266, opened 2020-05-17
Example: os_architecture.txt Normalization

hdm:
## Description

This is a starter PR to demonstrate how the normalization towards known identifiers can work.

1) Fingerprints which asserted similar, but non-standard identifiers were fixed.
2) The identifier list was updated with the net-new entries from fingerprints.

Any feedback on this? This is a small example compared to what the vendors, os, or service updates will look like. The idea is that these changes are safe because they push more fingerprints to the identifiers already used by existing consumers. 

## Motivation and Context

The goal is to standardize on as few unique identifiers as possible for all fingerprints.

## How Has This Been Tested?

- [x] recog_verify of modified files
- [x] rspec

## Types of changes
- Non-breaking change to move existing fingerprints to known architecture constants.

## Checklist:

### DIFF
--- a/xml/architecture.xml
+++ b/xml/architecture.xml
@@ -6,7 +6,7 @@
   <fingerprint pattern="x64|amd64|x86_64" flags="REG_ICASE">
     <description>x64 (x86_x64)</description>
     <example>Linux claw 3.11.0-15-generic #23-Ubuntu SMP Mon Dec 9 18:17:04 UTC 2013 x86_64 x86_64 x86_64 GNU/Linux</example>
-    <param pos="0" name="os.arch" value="x64"/>
+    <param pos="0" name="os.arch" value="x86_64"/>
   </fingerprint>
   <fingerprint pattern="x86|i[3456]86" flags="REG_ICASE">
     <description>x86</description>
@@ -15,18 +15,18 @@
   </fingerprint>
   <fingerprint pattern="PowerPC|PPC|POWER|ppc">
     <description>PowerPC</description>
-    <param pos="0" name="os.arch" value="ppc"/>
+    <param pos="0" name="os.arch" value="PowerPC"/>
   </fingerprint>
   <fingerprint pattern="SPARC" flags="REG_ICASE">
     <description>SPARC</description>
-    <param pos="0" name="os.arch" value="sparc"/>
+    <param pos="0" name="os.arch" value="Sparc"/>
   </fingerprint>
   <fingerprint pattern="mips" flags="REG_ICASE">
     <description>MIPS</description>
-    <param pos="0" name="os.arch" value="mips"/>
+    <param pos="0" name="os.arch" value="MIPS"/>
   </fingerprint>
   <fingerprint pattern="arm" flags="REG_ICASE">
     <description>ARM</description>
-    <param pos="0" name="os.arch" value="arm"/>
+    <param pos="0" name="os.arch" value="ARM"/>
   </fingerprint>
 </fingerprints>
--- a/xml/architecture.xml
+++ b/xml/architecture.xml
@@ -25,6 +25,10 @@
     <description>MIPS</description>
     <param pos="0" name="os.arch" value="MIPS"/>
   </fingerprint>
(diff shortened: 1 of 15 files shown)


tsellers-r7:
Thanks for the contribution @hdm It's going to be later this week before I can engage on this.

@gwiseman-r7 - Want to take a look as well

hdm:
@tsellers-r7 thanks! we can go either way on this, either just feedback and close this one, or close #265 and try to merge this one instead.

xml/snmp_sysdescr.xml
```diff
@@ -5866,7 +5866,7 @@ Copyright (c) 1995-2005 by Cisco Systems
     <param pos="0" name="os.vendor" value="Sun"/>
     <param pos="0" name="os.family" value="Solaris"/>
     <param pos="0" name="os.product" value="Solaris"/>
-    <param pos="0" name="os.arch" value="sun4u"/>
+    <param pos="0" name="os.arch" value="Sparc"/>
```
busterb, line 5869: This looks reasonable. Nobody's 32-bit Sparc systems work anymore anyway, except for many a few classic computer collectors.

xml/architecture.xml
```diff
@@ -15,18 +15,18 @@
...
-    <param pos="0" name="os.arch" value="sparc"/>
+    <param pos="0" name="os.arch" value="Sparc"/>
   </fingerprint>
   <fingerprint pattern="mips" flags="REG_ICASE">
     <description>MIPS</description>
-    <param pos="0" name="os.arch" value="mips"/>
+    <param pos="0" name="os.arch" value="MIPS"/>
   </fingerprint>
   <fingerprint pattern="arm" flags="REG_ICASE">
     <description>ARM</description>
```
busterb, line 29: We recently added 'aarch64' as well for 64-bit ARM.

busterb:
Having taken an initial look, nothing sticks out to me as problematic. Thanks for moving this forward, I really appreciate it!

hdm:
Thanks! I'll get rolling on the related work this week once this merges (all the other OS and Service fields)

xml/architecture.xml
sranjit-r7: Yes, I think it would be good to add 'aarch64' for ARM64 (Ref: https://github.com/rapid7/assessment-core/blob/0007d245ded9275ef495800227b155d822bc0237/fingerprinting-core/src/main/java/com/rapid7/fingerprint/system/IArchitectureConstants.java#L18)
Regarding the case, from my understanding, I don't think there's any dependency from here. But, I think it would be better to keep it alike so we won't have to check multiple places for any dependency and will be consistent.

hdm: You might want to delete this private repository URL from the comment. I went ahead and pushed arm64/aarch64 matches to "ARM64".

hdm:
Would it make sense to merge in all the other changes and handle this at once, versus a drip of PRs?

xml/architecture.xml
rkirk-r7: Looks like Nexpose has arm64/aarch64 mapping to "aarch64" not to "ARM64" with that recent update

hdm:
Thanks folks!

xml/architecture.xml
sranjit-r7: Yes, In nexpose, ARM64 is mapped to "aarch64".

hdm: Does this impact this PR?

hdm: Ex: all patterns used in this xml match to an uppercase value. Do you mean Nexpose maps only aarch64 to lowercase?

sranjit-r7: Yes, we had recently added this one in lowercase.

hdm: That sounds like a Nexpose bug =D 

Does this PR need any changes before it can be merged?

hdm:
This looks super-approved, any chance someone can merge it? I can get to work on the rest.

tsellers-r7:
@hdm I'm going to try to land this today.

PR #268, opened 2020-06-12
Upgrade lxml, improve fingerprint readability

tsellers-r7:
## Description
This PR:
- bumps the version of `lxml`
- Updates the CPE `Updating CPE` section of CONTRIBUTING.MD so that newlines are added after every fingerprint in a fingerprint database when updating CPEs.
   - The `sed` command was constructed so as add newlines after each fingerprint but avoid these compounding if the command was run multiple times. It also leaves the comments immediately above the fingerprints the comments apply to.
   - The command has been run on the XML databases and these changes are in this PR.
- Documentation
   - Updates README.MD so as to point users to the CONTRIBUTING.MD document.
   - Adds Table of Contents and re-organizes information for better flow
   - Performs some general Markdown linting/cleanup on README.ME and CONTRIBUTING.MD

## Motivation and Context
- Fix a security issue with `lxml`. This library is used in the CPE code and the vulnerability isn't reachable by any normal use of this code.
- Improve readability of the fingerprints by adding a newline between them.

## How Has This Been Tested?
- multiple runs of the CPE updating command
- `rspec`

## Types of changes

## Checklist:


tsellers-r7:
@hdm  - FYI, I think this should address your request related to whitespace around fingerprints.  It was being removed by the CPE process since we rebuild the XML document. `xmllint` is what is doing the final formatting.  I've added a `sed` command to handle the whitespace changes.  It's not my favorite way to handle this so if you've a cleaner way to do it let me know.

CONTRIBUTING.md
```diff
@@ -73,48 +96,95 @@ Please note that changes to [lib/recog/version.rb](https://github.com/rapid7/rec
...
+```
 
 Run the CPE automation against every XML file, using GNU `parallel` to speed things up:
 
-  ```
-  ls xml/*.xml | parallel --gnu "./update_cpes.py {} official-cpe-dictionary_v2.3.xml cpe-remap.yaml && xmllint --format --noblanks {} > {}.bak &&  mv {}.bak {} || echo {}" 2> errors.txt
-  ```
+```bash
+ls xml/*.xml | parallel --gnu "./update_cpes.py {} official-cpe-dictionary_v2.3.xml cpe-remap.yaml && xmllint --format --noblanks {} | sed  ':a;N;$ba;s|</fingerprint>\n  <|</fingerprint>\n\n  <|g;P;D' > {}.bak &&  mv {}.bak {} || echo {}" 2> errors.txt
+```
```
tsellers-r7, line 165: This and a similar section below it are the commands to update the CPEs and then format the resulting XML documents.  I had to create a section for `macOS` because I couldn't figure out to do use `sed` to do what I needed in a portable way.

PR #269, opened 2020-06-14
New fingerprints and completed normalization

hdm:
## Description

This PR completes the normalization work for all the other field types (os.product, vendors, etc) and introduces a bunch of new fingerprints.

## Motivation and Context

Creates a baseline for known identifiers and improves coverage.

## How Has This Been Tested?

Real-world data, rspec, and individual test cases.

## Types of changes
- Non-breaking changes (no existing identifiers modified)
- New fingerprints

## Checklist:


busterb:
Wow, nice additions! Thanks, HD.

hdm:
Apologies for the noise, GH complained about post-merge conflicts, this should resolve them.

tsellers-r7:
@hdm -
Is there anything in this process that should be added to the CONTRIBUTING.md document or should that be later?

Also, can you run the update CPE process as described here: https://github.com/rapid7/recog/blob/master/CONTRIBUTING.md#updating-cpes

Alternatively, if you just want to add the new fingerprint whitespacing (added #268) you can just run one of the following for reformatting:

**LInux** with `sed`
```bash
ls xml/*.xml | parallel --gnu "xmllint --format --noblanks {} | sed  ':a;N;$ba;s|</fingerprint>\n  <|</fingerprint>\n\n  <|g;P;D' > {}.bak &&  mv {}.bak {} || echo {}" 2> errors.txt
```

**macOS** with `gsed` (see CONTRIBUTING.md)
```bash
ls xml/*.xml | parallel --gnu "xmllint --format --noblanks {} | gsed  ':a;N;$ba;s|</fingerprint>\n  <|</fingerprint>\n\n  <|g;P;D' > {}.bak &&  mv {}.bak {} || echo {}"
```

hdm:
@tsellers-r7 Sure thing! see the latest push. The notes were updated to be less mac-specific (Windows uses a different path for venv scripts). A non-parallel example was provided.  The use of `xmllint` and `sed` was swapped for a quick ruby script. It would be awesome of the cpe script was ported to ruby to keep the dependency spread down in this repo.

bin/recog_cleanup
```diff
@@ -0,0 +1,16 @@
...
+
+$:.unshift(File.expand_path(File.join(File.dirname(__FILE__), "..", "lib")))
+require 'optparse'
+require 'ostruct'
+require 'recog'
+
+# Cleanup trailing whitespace an
+Dir[ File.expand_path(File.join(File.dirname(__FILE__), "..", "xml")) + "/*.xml" ].each do |f|
+    data = File.read(f).
+        gsub(/\s+$/, '').                           # Trailing whitespace and empty lines
```
tsellers-r7, line 11: This is removing whitespace inside of comments. I think these should be left for readability. In the `sed` command I checked for a single newline after `</fingerprint>` so that we could safely replace without cumulative line additions.

hdm: This only seemed to impact a couple comments was much less complex than the xmllint+sed line. For my dev case at least (running this on Windows), getting a working cli xmllint is a pain, not to mention lack of support for parallel, etc.

Maybe we can switch to a REXML or Nokogiri-based formatted instead?

tsellers-r7: +++ For platform independent commands. Might be able to tweak the regex in the following command, I'll take a look later today.

tsellers-r7:
@hdm - Landed, thanks for the contributions. I'll try to find some time to revisit the the regex to allow the ability to preserve white space in comments. I also totally agree w/ moving the CPE process to Ruby.

hdm:
Thanks yall!

PR #276, opened 2020-07-14
DNS tweaks

tsellers-r7:
## Description
This PR adds coverage for an additional DNS fingerprint of DNS running on Microsoft Windows 2008 SP2.  There are other misc DNS tweaks as well and some performance related changes.

The main thing I'm concerned about is the 'assert nothing' results that end up assigning a `service.protocol` value which is inherited from the XML database. 

```shell
$ echo '' | bin/recog_match xml/dns_versionbind.xml 

MATCH: {
  "matched"=>"empty string -- assert nothing.", 
  "service.protocol"=>"dns", 
  "fingerprint_db"=>"dns.versionbind", 
  "data"=>""}
```

 I'm addressing this by setting `service.certainty` to `0.0.  It this logic makes sense we can apply it elsewhere when the same assert nothing style matches are used.

```shell
$ echo '' | bin/recog_match xml/dns_versionbind.xml 

MATCH: {
  "matched"=>"empty string -- assert nothing.", 
  "service.certainty"=>"0.0", 
  "service.protocol"=>"dns",
  "fingerprint_db"=>"dns.versionbind", 
  "data"=>""}
```

## Motivation and Context
Coverage

## How Has This Been Tested?
`rspec`, local testing

## Types of changes

## Checklist:

### DIFF
--- a/xml/dns_versionbind.xml
+++ b/xml/dns_versionbind.xml
@@ -8,6 +8,32 @@
     dnsmasq-2.76-1-ubnt2
   -->
 
+  <!--
+  The following 'assert nothing' block is intended to handle banners so simple
+  that they cannot be attributed to a product or vendor. They are at the
+  beginning of the file as a performance tweak given how frequenty they occur.
+  -->
+
+  <fingerprint pattern="^none$">
+    <description>bare 'none' -- assert nothing.</description>
+    <example>none</example>
+  </fingerprint>
+
+  <fingerprint pattern="(?i)^unknown$">
+    <description>bare 'unknown' -- assert nothing.</description>
+    <example>unknown</example>
+  </fingerprint>
+
+  <fingerprint pattern="^no version$">
+    <description>bare 'no version' -- assert nothing.</description>
+    <example>no version</example>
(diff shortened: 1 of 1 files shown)


xml/dns_versionbind.xml
```diff
@@ -8,6 +8,40 @@
...
+  </fingerprint>
+
+  <fingerprint pattern="^no version$">
+    <description>bare 'no version' -- assert nothing.</description>
+    <example>no version</example>
+    <param pos="0" name="service.certainty" value="0.0"/>
+    <param pos="0" name="service.protocol" value=""/>
+  </fingerprint>
+
+  <fingerprint pattern="^$">
```
dabdine-r7, line 38: move this to the top? debatable, as project sonar has a ton of empty replies on version.bind, but other commercial products may have non-empty replies first, in which case this move is not a real optimization for them. 

On the other hand, it's just a single, simple regular expression...

tsellers-r7: Actually, no, I should have done that.  It's a high volume match and computationally cheap. Thanks.

PR #272, opened 2020-07-13
Adjust banner for $ProjectRevision

aluan-r7:
## Description
Updated $ProjectRevision banner so that it description is changed to Treck TCP/IP instead of HP printers.
Added examples of other version formats found and added example from running the Treck demo. Treck demo HTTP server included extra words Last Checkpoint e.g. $Project Revision: Last Checkpoint: 4.2.2.13
Changed regex to capture all versions.
added service and software param name values and added capture group for version

## Motivation and Context
Changes were made to fingerprint Treck TCP/IP software running on devices in order to run vulnerability checks.

## How Has This Been Tested?
A clear and concise description of your changes were tested.
Rake test passed
Screen Shot 2020-07-13 at 3 19 25 PM

## Checklist:

- [x] I have updated the documentation accordingly (or changes are not required).
- [x] I have added tests to cover my changes (or new tests are not required).
- [x] All new and existing tests passed.

PR #278, opened 2020-07-15
CPE: Correct CPE generating automation

tsellers-r7:
## Description
PR #272 re-raised an issue where `update_cpes.py` wasn't correctly mapping products which had a `/` in the name. The new PR correctly escapes the forward slash.

This:
```xml
<param pos="0" name="service.product" value="TCP/IP"/>
<param pos="0" name="service.vendor" value="Treck"/>`
```

should result in following.
```xml
<param pos="0" name="service.cpe23" value="cpe:/a:treck:tcp\/ip:{service.version}"/>
```

It was failing because it used `tcp/ip` instead of `tcp\/ip` which isn't a valid NIST issued CPE.  See: https://nvd.nist.gov/products/cpe/search/results?keyword=Treck&status=FINAL&orderBy=CPEURI&namingFormat=2.3

The PR also contains misc formatting fixes that were automatically handled by the scripts.

## Motivation and Context
Correct CPE generating automation, few work arounds in `cpe-mapping.xml`

## How Has This Been Tested?
`rspec`

## Types of changes

## Checklist:

### DIFF
--- a/cpe-remap.yaml
+++ b/cpe-remap.yaml
@@ -48,7 +48,6 @@ mappings:
     products:
       lotus_domino: lotus_domino_server
       os/400: os_400
-      z/os: z\/os
   jamf:
     products:
       jamf_pro: jamf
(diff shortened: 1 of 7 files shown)


tsellers-r7:
@aluan-r7 - FYI, this should fix the lack of CPE generation for the fingerprint you added.

PR #283, opened 2020-08-04
Project Sonar: SonicWall/Cisco Expressway

tsellers-r7:
## Description
Fingerprint additions / tweaks for SonicWall and Cisco Expressway products

## Motivation and Context
Improved coverage

### DIFF
--- a/xml/http_servers.xml
+++ b/xml/http_servers.xml
@@ -1724,7 +1724,7 @@
     <param pos="0" name="hw.device" value="VPN"/>
   </fingerprint>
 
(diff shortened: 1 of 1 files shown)


xml/http_servers.xml
```diff
@@ -1724,7 +1724,7 @@
     <param pos="0" name="hw.device" value="VPN"/>
   </fingerprint>
 
-  <fingerprint pattern="^SonicWALL (SSL-?VPN(?: (?:\d+))?) Web Server\.?$">
+  <fingerprint pattern="^SonicWALL SSL-?VPN Web Server$">
```
tsellers-r7, line 1727: I observed no variants where the capture would have provided value and we don't have any examples.  I removed this part of the regex and will monitor for future banners.

PR #286, opened 2020-08-30
Fix wrong values in snmp_sysdescr.xml

nstylepro:
Changed wrong values of 2 fingerprints

### DIFF
--- a/xml/snmp_sysdescr.xml
+++ b/xml/snmp_sysdescr.xml
@@ -6318,7 +6318,7 @@ Copyright (c) 1995-2005 by Cisco Systems
     <example>Siemens, SIMATIC HMI, KTP600 Basic Mono PN, 6AV6647-0AB11-3AX0, HW:1, FW:V01.06.00</example>
     <example>Siemens, SIMATIC HMI, KTP600 Basic color PN, 6AV6 647-0AD11-3AX0, HW:1, FW:V11.00.02.00</example>
     <param pos="0" name="os.vendor" value="Siemens"/>
-    <param pos="0" name="os.device" value="Monitoring"/>
+    <param pos="0" name="os.device" value="HMI Controller"/>
     <param pos="0" name="os.family" value="Simatic HMI"/>
     <param pos="1" name="os.product"/>
     <param pos="2" name="os.version"/>
@@ -6332,7 +6332,7 @@ Copyright (c) 1995-2005 by Cisco Systems
     <example>Siemens, SIMATIC HMI, unknown, 6AV2 124-0GC01-0AX0, HW: 0, SW: V 11 0 2</example>
     <example>Siemens, SIMATIC HMI, unknown, 6AV2 124-0JC01-0AX0, HW: 0, SW: V 11 0 0</example>
     <param pos="0" name="os.vendor" value="Siemens"/>
-    <param pos="0" name="os.device" value="Monitoring"/>
+    <param pos="0" name="os.device" value="HMI Controller"/>
     <param pos="0" name="os.family" value="Simatic HMI"/>
     <param pos="1" name="os.product"/>
     <param pos="2" name="os.version"/>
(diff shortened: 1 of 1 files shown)


hdm:
👍 LGTM

tsellers-r7:
Thanks for the contribution. We don't currently have anything else that uses `os.device` of `HMI Controller`. Keeping in mind that `os.device` is a generic classification for the device, what's the value of adding the specificity here? Is this a label we should apply to other fingerprints?

tsellers-r7:
As a note, I see that it is listed in https://github.com/rapid7/recog/blob/master/identifiers/hw_device.txt and that we have one occurrence of it's use;

https://github.com/rapid7/recog/blob/c0deef7c674144239fd675a9a6c14e05ec079a17/xml/html_title.xml#L1316-L1325

nstylepro:
well, that's incorrect. there is this one:
https://github.com/rapid7/recog/blob/c2beb75492c2938a4006bb8781214181c4edb457/xml/html_title.xml#L1316

the specificity is not so important, but Monitoring is not the proper classification for HMI devices, while they are usually old Windows box

tsellers-r7:
I made a comment on that one about a split second before you did.  Make sense, particularly if this is a firmware image and not stock Windows with some service running on it.  This seems like it'd be a good time to had `hw.*` fields as well since it appears that Siemens makes these and the banner contains model information. Does that sound correct?  

If it is correct I can follow this up with a PR to add them if needed.

nstylepro:
Sound like a good Idea.

nstylepro:
actually did some further research on these devices. turned out this is running WindowsCE! 
(not   <param pos="0" name="os.family" value="Simatic HMI"/> )
https://support.industry.siemens.com/cs/images/thumbnails/28263711/protool_image_versionen_01_e.gif

should fix that as well. LMK if you want me to push the change as well.

e.g. here:
https://github.com/rapid7/recog/blob/c2beb75492c2938a4006bb8781214181c4edb457/xml/snmp_sysdescr.xml#L6322

tsellers-r7:
@nstylepro - Yes, if you could add these changes that'd be great. Thanks!

nstylepro:
> @nstylepro - Yes, if you could add these changes that'd be great. Thanks!

review again :)

xml/snmp_sysdescr.xml
```diff
@@ -6318,10 +6318,12 @@ Copyright (c) 1995-2005 by Cisco Systems
     <example>Siemens, SIMATIC HMI, KTP600 Basic Mono PN, 6AV6647-0AB11-3AX0, HW:1, FW:V01.06.00</example>
     <example>Siemens, SIMATIC HMI, KTP600 Basic color PN, 6AV6 647-0AD11-3AX0, HW:1, FW:V11.00.02.00</example>
     <param pos="0" name="os.vendor" value="Siemens"/>
```
tsellers-r7, line 6320: `os.vendor` here and elsewhere should be Microsoft.
`hw.vendor` should, I'm guessing, be `Siemens`?
Can you group the `os.*` and `hw.*` together?

xml/snmp_sysdescr.xml
```diff
@@ -6369,9 +6377,10 @@ Copyright (c) 1995-2005 by Cisco Systems
     <example os.product="IM151-8" os.version="3.2.3">Siemens, SIMATIC S7, IM151-8, 6ES7 151-8AB01-0AB0 , HW: 2, FW: V3.2.3, S C-B3UC78192011</example>
     <param pos="0" name="os.vendor" value="Siemens"/>
     <param pos="0" name="os.device" value="Monitoring"/>
-    <param pos="0" name="os.family" value="Simatic S7"/>
-    <param pos="1" name="os.product"/>
-    <param pos="2" name="os.version"/>
+    <param pos="0" name="os.family" value="Windows"/>
+    <param pos="0" name="hw.family" value="Simatic S7"/>
+    <param pos="1" name="hw.product"/>
```
tsellers-r7, line 6382: The changes from `os.product` to `hw.product` are what broke the tests. You can change the field in the example and retest using `rspec`

tsellers-r7:
I'll land this and fix the issues. Thanks much for the feedback and pr @nstylepro

tsellers-r7:
FYI, here is a PR with the changes: https://github.com/rapid7/recog/pull/291

PR #296, opened 2020-10-27
CPE: use safe_load, update CPEs

tsellers-r7:
## Description
This PR changes `update_cpes.py` so that it uses `yaml.safe_load()` instead of `yaml.load()`. This should reduce the risk when loading YAML.  I've also run it against our fingerprint databases using the latest data from NIST.

I've also added a `.snyk` file in order to suppress the Synk warning on PyYAML. We don't currently use a vulnerable code path, Snyk doesn't check to see if you are using the vulnerable code path, and there is no full library level fix for this other than to use `safe_load()` (which this PR does) or the `SafeLoader` loader.

https://app.snyk.io/vuln/SNYK-PYTHON-PYYAML-590151

## Motivation and Context
Risk reduction

## How Has This Been Tested?
`rspec`, Github PR hooks

## Types of changes

## Checklist:

### DIFF
--- a/xml/x509_issuers.xml
+++ b/xml/x509_issuers.xml
@@ -14,7 +14,8 @@
     <description>Google Chromecast Gen 1</description>
     <example>CN=Eureka Gen1 ICA,OU=Google TV,O=Google Inc,L=Mountain View,ST=California,C=US</example>
     <param pos="0" name="os.vendor" value="Google"/>
-    <param pos="0" name="os.product" value="ChromeOS"/>
+    <param pos="0" name="os.product" value="Chrome OS"/>
+    <param pos="0" name="os.cpe23" value="cpe:/o:google:chrome_os:-"/>
     <param pos="0" name="hw.device" value="Media Server"/>
     <param pos="0" name="hw.vendor" value="Google"/>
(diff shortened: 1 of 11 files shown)


xml/x509_issuers.xml
```diff
@@ -32,7 +33,8 @@
     <example chromecast.generation="11" chromecast.capabilities="Video Assist">CN=Chromecast ICA 11 (Video Assist),OU=Cast,O=Google Inc,L=Mountain View,ST=California,C=US</example>
     <example chromecast.generation="12">CN=Chromecast ICA 12,OU=Cast,O=Google Inc,L=Mountain View,ST=California,C=US</example>
     <param pos="0" name="os.vendor" value="Google"/>
-    <param pos="0" name="os.product" value="ChromeOS"/>
+    <param pos="0" name="os.product" value="Chrome OS"/>
```
tsellers-r7, line 36: Note, this change aligns the name with the Google usage of the term (space vs no space) and allows the NIST CPE mapping to automatically match.  If this causes issue I can handle it with a remapping instead..

CC @hdm

rkirk-r7 (approved):

PR #300, opened 2020-12-04
CPE: cpe_update.py - Address matching fall through logic

tsellers-r7:
## Description
The PR updates the CPE generation script to address an issue that occurs when a vendor exists in the official CPE list but you still need to remap the vendor for a particular product.

We have a remap in place to map `ubiquiti` to `ui` (the correct NIST value) but it was never tried since `ubiquiti` was in the official CPE vendor list as well.  The new logic will try the original logic to find a vendor/product match but if this fails and a remap exists it will check those values as well.  

The priority is:
1. Original vendor / product
1. Original vendor / remap product
1. Remap vendor / original product
1. Remap vendor / remap product

### This PR also
1. Cleans up some newly unnecessary `vendor` entries in the remap file.
1. Performs some linting

## Motivation and Context
Improved CPE auto-matching

## How Has This Been Tested?
Multiple tests locally across all fingerprint databases and ensure that only the expected entries are changed and no additional errors are generated.

**ALL CPE values changed in this PR were done via this script**.

```shell
# Do this once
curl -o official-cpe-dictionary_v2.3.xml.gz https://nvd.nist.gov/feeds/xml/cpe/dictionary/official-cpe-dictionary_v2.3.xml.gz && \
gunzip official-cpe-dictionary_v2.3.xml.gz

# Run the following to test logic
touch errors.txt && \
rm errors.txt && \
ls xml/*.xml | parallel --gnu "python update_cpes.py {} official-cpe-dictionary_v2.3.xml cpe-remap.yaml"  2>>errors.txt && \
ruby bin/recog_cleanup
```

## Types of changes

## Checklist:


update_cpes.py
```diff
@@ -20,9 +20,9 @@ def parse_cpe_vp_map(file):
         cpe_match = re.match('^cpe:/([aho]):([^:]+):([^:]+)', cpe_name)
         if cpe_match:
             cpe_type, vendor, product = cpe_match.group(1, 2, 3)
-            if not cpe_type in vp_map:
+            if cpe_type not in vp_map:
                 vp_map[cpe_type] = {}
-            if not vendor in vp_map[cpe_type]:
+            if vendor not in vp_map[cpe_type]:
```
tsellers-r7, line 25: This and just above are just lintng changes.

requirements.txt
```diff
@@ -1,2 +1,2 @@
-lxml==4.5.1
+lxml==4.6.2
```
tsellers-r7, line 1: Bumped to address an `lxml` vulnerability as identified by Snyk. I have retested to ensure that this doesn't impact the CPEs that are generated or the values added to the XML.

image

cpe-remap.yaml
```diff
@@ -2,7 +2,6 @@ mappings:
   alpine:
     vendor: alpinelinux
   apache:
-    vendor: apache
     products:
```
tsellers-r7, line 5: The new changes make entries mapping the vendor back to the same string unnecessary.

tsellers-r7:
Went ahead and landed this to unlock some further CPE updates.

PR #302, opened 2020-12-06
recog_standardize: adjust logic

tsellers-r7:
## Description
This PR addresses what I believe to a bug in `bin/recog_standardize` in which a string index is compared to `-1` instead of `nil`. Since `-1` shouldn't be a valid response to this check all entries were skipped.

I noticed this when working on PR #301 and not seeing some of the changes I made. I then added a random vendor name and `recog_standardize` didn't report it.

**NOTE**: In this PR I have not included the changes to the files in `identifiers/` since I need some of the changes in PR #301 to land first. When the issue is addressed I'll open a PR to clean up.

You can see the results either of the following commands before and after the changes in this PR.

CC @hdm 

```
# Across http_servers.xml
ruby bin/recog_standardize xml/http_servers.xml

# Across all of the files
for db in xml/*.xml; do ruby bin/recog_standardize $db; done
```

### Output when running the updated script against `http_servers.xml`

```
$ ruby bin/recog_standardize xml/http_servers.xml 
VENDOR MISSING: CentOS WebPanel
SERVICE PRODUCT MISSING: Web Cache
SERVICE PRODUCT MISSING: Application Server Web Cache
SERVICE PRODUCT MISSING: Serv-U FTP Server
VENDOR MISSING: WFTPServer
SERVICE FAMILY MISSING: Wing FTP
SERVICE PRODUCT MISSING: Wing FTP Server
SERVICE PRODUCT MISSING: GlassFish Server
VENDOR MISSING: lighttpd
SERVICE PRODUCT MISSING: NetWeaver Application Server Java
SERVICE PRODUCT MISSING: NetWeaver Application Server
SERVICE PRODUCT MISSING: NetWeaver Internet Communication Manager
SERVICE PRODUCT MISSING: NetWeaver AS ABAP
SERVICE PRODUCT MISSING: Internet Graphics Server
SERVICE PRODUCT MISSING: SAP Message Server
SERVICE PRODUCT MISSING: SQL Anywhere
SERVICE PRODUCT MISSING: OpenVPN Access Server
SERVICE PRODUCT MISSING: Expressway
SERVICE PRODUCT MISSING: STUN Server
VENDOR MISSING: TigerVNC
SERVICE PRODUCT MISSING: TigerVNC
VENDOR MISSING: EmbedThis
VENDOR MISSING: Zed Shaw
VENDOR MISSING: Ruby-Lang
VENDOR MISSING: Aspen
VENDOR MISSING: Boa
VENDOR MISSING: Genivia
[...]

PR #303, opened 2020-12-07
Update standard identifiers

tsellers-r7:
## Description
The goal of this PR is to reset the standard identifiers after the change in PR #302.  There are quite a few changes here that are due to updates being made while `bin/recog_standardize` wasn't working as expected. There are also cleanup of records that appear to be from the original PR.

**NOTE**: Most consumers of Recog will only care about the tweaks to the fingerprint files that standardized device types, vendor names, etc. which are toward the end of the diff.

The process to perform the update was to:
1. Remove the contents of all `.txt` files in `/identifiers`
2. Run `for db in xml/*.xml; do ruby bin/recog_standardize $db -w ; done` to sequentially work through each database file

**CHANGES OF CONCERN**
1. Standardization on Title Case for `os.device` and `hw.device` since most values already were.
1. All instances of `Web cam` or `Web Cam` in `*.device` fields are now `IP Camera`
2. Removal of `software_class.txt`, `software_family.txt`, and `software_product.txt` in the `identifiers` directory since this codebase doesn't use these files at all.

## How Has This Been Tested?
Local testing and review
rspec

## Types of changes

## Checklist:


identifiers/hw_device.txt
```diff
@@ -71,7 +71,7 @@ VoIP Switch
 Voice Appliance
 WAP
 WLAN Repeater
-Web cam
+Web Cam
```
tsellers-r7, line 74: Standardize `Web Cam` on title case.

identifiers/vendor.txt
```diff
@@ -223,36 +177,31 @@ Dell
...
-EasyTel
 Eaton
 Ecelerity
 Eclipse
-Edimax
-Eicon
 Ektron
 Elastic
 Eltek
+EmbedThis
```
tsellers-r7, line 204: The name / case of this has changed over time.

xml/favicons.xml
```diff
@@ -423,7 +423,7 @@
     <example>5856edf7bcbea0817312d9e535e5eb2a</example>
     <example>f4f3cb900258441d5dbc9105b7ab9b44</example>
     <example>c6acedaff906029fc5455d9ec52c7f42</example>
-    <param pos="0" name="service.vendor" value="VMWare"/>
+    <param pos="0" name="service.vendor" value="VMware"/>
```
tsellers-r7, line 426: Here and elsewhere: Correcting the case of `VMware`.

bin/recog_standardize
```diff
@@ -111,7 +111,7 @@ ARGV.each do |arg|
             puts "HW DEVICE MISSING: #{val}"
             hw_device[val] = true
           end          
-        when "service.product"
+        when "service.product", "service.component.product"
```
tsellers-r7, line 114: Collect `service.component.product` since the script already does the same for `service.component.vendor`.

identifiers/os_device.txt
```diff
@@ -47,48 +40,36 @@ Network
 Network Audio
 Network Management Device
 Network Scanner
-Network management device
 Onboard Administrator
-PBX
-PDA
 PDU
 PLC
 POS
```
tsellers-r7, line 46: @hdm - We've both `POS` and `Point of Sale` for device values.  I think we should move to one or the other. Any opinion from you before I switch from `POS` to `Point of Sale`?

tsellers-r7:
CCing @hdm  since you're one of the primary users/contributors for this much of these files, the changed `*.device` fields, and Recog in general.

hdm:
No issues on this side, I was more worried about changes to this affecting the Nexpose folks. If they are good to go, even better =D

hdm:
"Web cam" -> "Web Cam" is something we do internally in Rumble today, but we left the original variant in place because that was how Nexpose tracked things. That whole category should probably be "IP Camera" these days.

tsellers-r7:
I think this is a great idea particularly since we already use IP Camera elsewhere (and I missed it). I'll reach out to a few folks internally.

identifiers/hw_device.txt
tsellers-r7: To HD's point we should just adopt `IP Camera` everywhere.

tsellers-r7:
I have switch all instances `Web Cam` in `*.device` to `IP Camera`.

tsellers-r7:
The single instance of `POS` as a device type has been changed to `Point of Sale`. This brings the number of `Point of Sale` fingerprints to ... 2.

xml/favicons.xml
abunn-r7: Verified this is how we match in product.

tsellers-r7: Thanks. It was mixed use (`VMWare` vs `VMware`) throughout the fingerprints but, IIRC, most of our fingerprints were using the correct case.  This is a good example of  why this tool has value.

tsellers-r7:
@hdm - In this PR I deleted the `indentifiers/software_*.txt` files since they weren't used by `bin/recog_standardize` or Recog in general.  In the case of `software_class.txt` I wasn't sure where you might have pulled the data from unless it was from non-Recog data such as Rumble.

Did you have a vision for how these should be used?
Would you like to see them returned? If so, how would you generate them?

hdm:
Thanks for the note! I think these came from the original email from the Nexpose team that started this.

hdm:
We can probably omit from the standardize code and identifier list, but the source of those was Rapid7 originally.

tsellers-r7:
> We can probably omit from the standardize code and identifier list..

AFAIK, there wasn't any code in `bin/recog_standardize` that dealt with or generated the `software_*.txt` files.  I removed them because I assumed they were just artifacts left over from the development process.

> I think these came from the original email from the Nexpose team that started this.

Do you know if I was on that email thread or PRs that contain them?  Also, we can take this to another medium if it's easier for you.

hdm:
No PR, but I forwarded the thread by email. Thanks!

PR #305, opened 2020-12-10
Additional x509 banners

psupel-r7:
## Description
Recog updated with fingerprints from Jess

## Motivation and Context
More functionality

## Types of changes

## Checklist:


xml/x509_subjects.xml
```diff
@@ -474,6 +482,14 @@
     <param pos="0" name="service.product" value="vmauthd"/>
   </fingerprint>
 
+  <fingerprint pattern="^CN=localhost,OU=Support,O=Palo Alto Networks$">
+    <description>Palo Alto Networks</description>
+    <example>CN=localhost,OU=Support,O=Palo Alto Networks</example>
```
tsellers-r7, line 487: Please use the standardized values for this vendor, product, etc. 
You can see an example here - https://github.com/rapid7/recog/blob/fe52bbcc54fdf25c11869c526b75158761de0884/xml/x509_subjects.xml#L1194-L1202

xml/x509_subjects.xml
```diff
@@ -168,6 +168,14 @@
     <param pos="0" name="os.product" value="iDRAC Linux"/>
   </fingerprint>
 
+  <fingerprint pattern="^CN=opensnort\.example.com,O=Sourcefire,emailaddress=support@sourcefire.com">
+    <description>Sourcefire Snort HTTPS console w/Service tag</description>
+    <example>CN=opensnort.example.com,O=Sourcefire,emailaddress=support@sourcefire.com</example>
+    <param pos="0" name="service.vendor" value="SOURCEFIRE"/>
```
tsellers-r7, line 174: After looking at the NIST database and Snort website I'd recommend using `Snort` for the vendor and product. It's accurate for today (Cisco purchased Sourcefire, and Snort is no longer branded as Sourcefire) and will enable autogeneration of CPEs.  

See: https://github.com/rapid7/recog/blob/master/CONTRIBUTING.md#updating-cpes

xml/x509_subjects.xml
```diff
@@ -168,6 +168,14 @@
     <param pos="0" name="os.product" value="iDRAC Linux"/>
   </fingerprint>
 
+  <fingerprint pattern="^CN=opensnort\.example.com,O=Sourcefire,emailaddress=support@sourcefire.com">
+    <description>Sourcefire Snort HTTPS console w/Service tag</description>
```
tsellers-r7, line 172: `w/Service tag` - Is that something in the fingerprint or context that didn't get copied here?

tsellers-r7 (changes requested):
See comments.

dabdine:
@tsellers-r7 we should probably add tests against the fields for the files in the 'identifiers' dir of this project (such as hw_product.txt). 

Also, I know R7 VM expects explicit vendor/product names to trigger downstream checks, so there may be some additional integration testing necessary so these additions/changes don't regress functionality there.

tsellers-r7:
@dabdine - Yup, after we move the tests from Travis-CI (tests that take 18+ hours to run aren't useful) we can just have it run `bin/recog_standardize` to check the results.

xml/x509_subjects.xml
rkirk-r7: We are using PAN-OS currently so this is something to consider.  Would also move this to be with the existing PANOS definition

xml/x509_subjects.xml
```diff
@@ -168,6 +168,14 @@
     <param pos="0" name="os.product" value="iDRAC Linux"/>
   </fingerprint>
 
+  <fingerprint pattern="^CN=opensnort\.example.com,O=Sourcefire,emailaddress=support@sourcefire.com">
+    <description>Sourcefire Snort HTTPS console</description>
+    <example>CN=opensnort.example.com,O=Sourcefire,emailaddress=support@sourcefire.com</example>
+    <param pos="0" name="service.vendor" value="Snort"/>
+    <param pos="1" name="service.family" value="Snort"/>
```
rkirk-r7, line 175: should this be pos 1? We are not using a capture group

xml/x509_subjects.xml
```diff
@@ -168,6 +168,14 @@
     <param pos="0" name="os.product" value="iDRAC Linux"/>
   </fingerprint>
 
+  <fingerprint pattern="^CN=opensnort\.example.com,O=Sourcefire,emailaddress=support@sourcefire.com">
+    <description>Sourcefire Snort HTTPS console</description>
+    <example>CN=opensnort.example.com,O=Sourcefire,emailaddress=support@sourcefire.com</example>
```
rkirk-r7, line 173: @tsellers-r7 thoughts on the ordering of email address here? Would we expect it before the CN component?

Also should it be `emailAddress`?

rkirk-r7 (commented):
Looks like we still need to port
* VMWare - we have os.family values, and looks like one regex match is missing
* iLo - some existing ones but do not look like they match ours

xml/x509_subjects.xml
tsellers-r7: In looking at the CPE data and the Palo Alto website, the value for `os.product` probably should be `PAN-OS`.  Lemme know what works for you and we can bring everything to a consistent state and add CPE mappings if needed.

xml/x509_subjects.xml
tsellers-r7: Hurm, I'm unsure. Based on the intro block in this file I'd *guess* that it'd be last but it's not clear.

xml/x509_subjects.xml
tsellers-r7: Correct

xml/x509_subjects.xml
rkirk-r7: `PAN-OS` is what Nexpose is currently using/expecting for our fingerprinting, so if we think changing recog to `PAN-OS` is correct then that would be perfect for us

hdm:
Hello! The EMAILADDRESS change is problematic today because the x509 representation differs by implementation. For example, Go is putting the email address at the end of the line and using the OID instead "EMAILADDRESS".

This comment is relevant:
```
<!--
  This fingerprint set matches the Subject field of x509 certificates. These x509
  certificates may be sourced from any SSL or TLS service. If a particular system
  has identical subject and issuer fields, the subject field should be preferred.
  The format of the Subject field is built from the x509 distinguished names using
  a specific order. This order matches the Go implementation at the URL:
    https://golang.org/src/crypto/x509/pkix/pkix.go#203
  The ToRDNSequence() function builds the string in reverse order:
        func (n Name) ToRDNSequence() (ret RDNSequence) {
            ret = n.appendRDNs(ret, n.Country, oidCountry)
            ret = n.appendRDNs(ret, n.Province, oidProvince)
            ret = n.appendRDNs(ret, n.Locality, oidLocality)
            ret = n.appendRDNs(ret, n.StreetAddress, oidStreetAddress)
            ret = n.appendRDNs(ret, n.PostalCode, oidPostalCode)
            ret = n.appendRDNs(ret, n.Organization, oidOrganization)
            ret = n.appendRDNs(ret, n.OrganizationalUnit, oidOrganizationalUnit)
            if len(n.CommonName) > 0 {
                ret = n.appendRDNs(ret, []string{n.CommonName}, oidCommonName)
            }
            if len(n.SerialNumber) > 0 {
                ret = n.appendRDNs(ret, []string{n.SerialNumber}, oidSerialNumber)
            }
            for _, atv := range n.ExtraNames {
                ret = append(ret, []AttributeTypeAndValue{atv})
            }
            return ret
        }
    All names are separated by commas and any commas inside a name are escaped with a
    single backslash character. See RFC 2253 for additional details on formatting.
    Practically, most Subjects start with the Common Name (CN=) and then step through
    Organization Unit (OU), Organization (O), and then some level of location, but
    typically Locality (L) and Country (C). Names are guaranteed to be listed in
    the order described above, but may start at any point in the list. For example,
    Subjects may start with a Serial Number (SERIALNUMBER=) or even Extra Names, but
    these are somewhat rare. Keep this name order in mind when working on these
    fingerprints.
    The same constraints also apply to the x509 Issuers (x509_issuers.xml).
  -->
```

Supporting EMAILADDRESS (and other attributes is great), but we need to agree on how to represent them first and make sure Go, Ruby, Python, Java all agree when building the string.

hdm:
The relevant Go code for extra name handling:
```
			oidString := tv.Type.String()
			typeName, ok := attributeTypeNames[oidString]
			if !ok {
				derBytes, err := asn1.Marshal(tv.Value)
				if err == nil {
					s += oidString + "=#" + hex.EncodeToString(derBytes)
					continue // No value escaping necessary.
				}

				typeName = oidString
			}
```

In practice this looks like:

```
CN=device.corp.com,OU=VMware Engineering,O=VMware,L=Palo Alto,ST=California,C=US,1.2.840.113549.1.9.1=#0c0f766d636140766d776172652e636f6d
```

1.2.840.113549.1.9.1 is the OID of the EMAILADDRESS attribute. The value includes the DER bytes, including the Type and Length before the value (vmca@vmware.com).

hdm:
Go also places these LAST where the example above has the EMAILADDRESS first. 

Anywho, this gets tricky because we normally use ^ anchors and this will different by language.

xml/x509_subjects.xml
```diff
@@ -475,7 +483,7 @@
   </fingerprint>
 
   <fingerprint pattern="^CN=([a-zA-Z0-9\.\-\_]+),OU=VMware ESX Server Default Certificate,O=VMware\\, Inc,L=Palo Alto,ST=California,C=US$">
-    <description>VMware ESX</description>
+    <description>VMWare ESX</description>
```
tsellers-r7, line 486: This PR will likely be closed, but note for self - this change is incorrect. `VMware` is correct.

xml/x509_subjects.xml
```diff
@@ -484,6 +492,16 @@
...
 
+  <fingerprint pattern="^CN=([a-zA-Z0-9\.\-\_]+),OU=VMware ESX Server Certificate,O=VMware\\, Inc,L=Palo Alto,ST=California,C=US$">
+    <description>VMWare ESX</description>
+    <example>CN=server99.,OU=VMware ESX Server Certificate,O=VMware\, Inc,L=Palo Alto,ST=California,C=US</example>
+    <param pos="0" name="os.vendor" value="VMware"/>
+    <param pos="0" name="os.product" value="ESX"/>
+    <param pos="0" name="os.device" value="Hypervisor"/>
+    <param pos="0" name="os.cpe23" value="cpe:/o:vmware:esx:-"/>
+    <param pos="1" name="host.name"/>
+  </fingerprint>
```
tsellers-r7, line 503: Already landed elsewhere

mkienow-r7:
Closing this pull request since x.509 normalization has not been worked out. See issue #360.

PR #340, opened 2021-02-04
Housekeeping / Fixup after recent merges

tsellers-r7:
## Description
This PR is cleanup and normalization after the last couple of PRs.

## How Has This Been Tested?
local tests, `rspec`, etc.

## Checklist:

### DIFF
--- a/xml/favicons.xml
+++ b/xml/favicons.xml
@@ -1810,7 +1810,7 @@
     <description>PBX in a Flash</description>
     <example>f5c62ea4c4e9f9a8606400becc01375e</example>
     <param pos="0" name="hw.vendor" value="PIAF"/>
-    <param pos="0" name="hw.device" value="PBX"/>
+    <param pos="0" name="hw.device" value="SIP Gateway"/>
     <param pos="0" name="hw.product" value="PIAF Virtual Appliance"/>
   </fingerprint>
 
--- a/xml/snmp_sysdescr.xml
+++ b/xml/snmp_sysdescr.xml
@@ -1040,6 +1040,8 @@
     <param pos="0" name="os.product" value="Fabric OS"/>
     <param pos="0" name="os.device" value="Switch"/>
     <param pos="0" name="os.cpe23" value="cpe:/o:brocade:fabric_os:-"/>
+    <param pos="0" name="hw.vendor" value="Brocade"/>
+    <param pos="0" name="hw.device" value="Switch"/>
   </fingerprint>
 
   <fingerprint pattern="^Brocade ((?:\S+ )?SAN Switch) Module for IBM eServer BladeCenter$">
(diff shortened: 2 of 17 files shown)


xml/favicons.xml
```diff
@@ -1810,7 +1810,7 @@
     <description>PBX in a Flash</description>
     <example>f5c62ea4c4e9f9a8606400becc01375e</example>
     <param pos="0" name="hw.vendor" value="PIAF"/>
-    <param pos="0" name="hw.device" value="PBX"/>
+    <param pos="0" name="hw.device" value="SIP Gateway"/>
```
tsellers-r7, line 1813: We've standardized on a few variants on `SIP` and `VoIP`. If PBX needs to be its own class we should do this across products.

xml/snmp_sysdescr.xml
```diff
@@ -1079,6 +1082,47 @@
...
+  <fingerprint pattern="^Brocade VDX Switch, BR-(\S+), Network Operating System Software Version (\S+)\.$">
+    <description>Brocade VDX Switch</description>
+    <example hw.product="VDX6740" os.version="7.0.2b">Brocade VDX Switch, BR-VDX6740, Network Operating System Software Version 7.0.2b.</example>
+    <param pos="0" name="os.vendor" value="Brocade"/>
+    <param pos="0" name="os.device" value="Switch"/>
+    <param pos="0" name="os.family" value="NetworkOS"/>
+    <param pos="2" name="os.version"/>
+    <param pos="0" name="hw.vendor" value="Brocade"/>
+    <param pos="0" name="hw.family" value="VDX"/>
+    <param pos="1" name="hw.product"/>
```
tsellers-r7, line 1094: I moved these Rumble related fingerprints (from the recently landed PR #334 ) in order to group them with the other Brocade fps. Unfortunately, this hid the changes I made.  They are:
1. Adding `hw.*` fields
2. Moving `os.product` to `hw.product`.

CC @hdm in the event that this is a breaking change for Rumble.

xml/ssh_banners.xml
```diff
@@ -2110,7 +2110,6 @@
     <param pos="0" name="os.product" value="Data ONTAP"/>
     <param pos="0" name="os.cpe23" value="cpe:/o:netapp:data_ontap:-"/>
     <param pos="0" name="hw.vendor" value="NetApp"/>
-    <param pos="0" name="hw.device" value="File server"/>
```
tsellers-r7, line 2113: Also from the PR #334 - Here and elsewhere I have removed `File Server`.  This would be a new value for `hw.device` and, with these fingerprints, you can't tell if the device is a SAN, NAS, FCoIP, etc.

Also CC @hdm

hdm: NAS is certainly a better default these days for ONTAP

xml/snmp_sysdescr.xml
hdm: 👍

tsellers-r7:
I really should have caught most of this when handling PR #334  
We need to make running `bin/recog_standardize` and CPE updates easier for contributors and improve the docs around running them and how to handle exceptions.

xml/ssh_banners.xml
tsellers-r7: Groovy, I'm comfortable with that label as well.

PR #344, opened 2021-02-25
Add a vCenter fingerprint based on HTML title

dabdine:
## Description
Adds a fingerprint for VMware vCenter. 

This is sort of experimental. The VMware vCenter page content looks like this:

```
<script type="text/javascript">document.write("<title>" + ID_VC_Welcome + "</title>");</script>
```

At least Shodan and Censys both grab the page title as `" + ID_VC_Welcome + "`.  However, it's obviously meant to be interpreted, as it's javascript (likely because of internationalization -- if you view what the page title ends up being after this is interpreted, it is "Welcome to VMware vSphere").

If we're OK with HTML title tag matching to be defined as simply "whatever is in the static text representation of two enclosing title tags" I think this fingerprint is valid. If we think it that title tag retrieval is defined by some other means (interpreted, etc.) then we should reject this change.

Thoughts?

## Motivation and Context
A 9.8 unauth remote file write that leads to RCE was announced.

## How Has This Been Tested?
Simply using `bin/recog_verify xml/html_title.xml`

## Types of changes
- New feature: Add fingerprint for generic VMware vCenter detection (versionless)

## Checklist:

### DIFF
--- a/xml/html_title.xml
+++ b/xml/html_title.xml
@@ -1547,6 +1547,14 @@
     <param pos="0" name="service.product" value="Site Recovery Manager"/>
   </fingerprint>
 
+  <fingerprint pattern="^&quot; \+ ID_VC_Welcome \+ &quot;$">
+    <description>VMware vCenter</description>
+    <example>&quot; + ID_VC_Welcome + &quot;</example>
+    <param pos="0" name="service.vendor" value="VMware"/>
+    <param pos="0" name="service.product" value="vCenter"/>
+    <param pos="0" name="service.cpe23" value="cpe:/a:vmware:vcenter_server:-"/>
+  </fingerprint>
+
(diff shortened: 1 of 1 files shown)


hdm:
I have been think about this too, we extract it the same way Shodan and others do today, even though it's not quite accurate.
I guess the downside is limited? If the implementation "does it right", this fingerprint won't, but most others will.

dabdine:
@hdm fair enough -- agreed downside is limited, in which case merging def makes sense here. We probably should "standardize" more and provide more canonical examples long term on what title tag extraction should look like / tests around it too, at some point outside of this PR :)

tsellers-r7:
Thanks @dabdine

dabdine:
thanks @hdm & @tsellers-r7 !

PR #348, opened 2021-04-23
Add Ubuntu match from nginx server header

ClintonCollins:


hdm:
This is a WIP from our fork (feel free to pull) and was PRd a bit early.

PR #351, opened 2021-04-26
Update http_cookies

tsellers-r7:
## Description
This PR updates `http_cookies` by:
- Adding missing examples
- Re-ordering some of the matches to improve metrics
- Removing unnecessary captures where the value doesn't change.

## Motivation and Context
Improving fingerprints

## How Has This Been Tested?
`rspec`

## Checklist:

### DIFF
--- a/xml/http_cookies.xml
+++ b/xml/http_cookies.xml
@@ -5,9 +5,49 @@
   servers.
   -->
 
+  <fingerprint pattern="^__cfd?uid=">
+    <description>CloudFlare web load balancer endpoint</description>
+    <example>__cfuid=1337</example>
+    <example>__cfduid=dd450f2431e1e611a61a15f68974de9a41618794671; expires=Wed, 19-May-21 01:11:11 GMT; path=/; domain=.foo.bar; HttpOnly; SameSite=Lax</example>
+    <param pos="0" name="service.vendor" value="CloudFlare"/>
+    <param pos="0" name="service.product" value="CloudFlare Load Balancer"/>
+    <param pos="0" name="service.family" value="CloudFlare"/>
+  </fingerprint>
+
+  <fingerprint pattern="^(PHPSESSI(?:D|ON))=.*">
+    <description>PHP - http://www.php.net/ref.session</description>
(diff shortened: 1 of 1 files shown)


xml/http_cookies.xml
```diff
@@ -5,9 +5,49 @@
   servers.
   -->
 
+  <fingerprint pattern="^__cfd?uid=">
+    <description>CloudFlare web load balancer endpoint</description>
+    <example>__cfuid=1337</example>
+    <example>__cfduid=dd450f2431e1e611a61a15f68974de9a41618794671; expires=Wed, 19-May-21 01:11:11 GMT; path=/; domain=.foo.bar; HttpOnly; SameSite=Lax</example>
```
tsellers-r7, line 11: Fingerprint moved up due to frequency of observation.  Coverage for `__cfduid` added

xml/http_cookies.xml
```diff
@@ -5,9 +5,49 @@
...
+    <example cookie="PHPSESSID">PHPSESSID=deleted; expires=Thu, 01-Jan-1970 00:00:01 GMT; Max-Age=0; path=/</example>
+    <example cookie="PHPSESSION">PHPSESSION=vt2ag6n7t6ngvlg8adk4860h46; path=/</example>
+    <param pos="1" name="cookie"/>
+    <param pos="0" name="service.vendor" value="PHP"/>
+    <param pos="0" name="service.family" value="PHP"/>
+    <param pos="0" name="service.product" value="PHP"/>
+    <param pos="0" name="service.cpe23" value="cpe:/a:php:php:-"/>
+  </fingerprint>
+
+  <fingerprint pattern="^(ASPSESSIONID[A-Z]+|ASP\.NET_SessionId|\.ASPXANONYMOUS)=.*">
```
tsellers-r7, line 28: This and above moved up for performance reasons

tsellers-r7:
FYI @hdm

hdm:
Thanks!

hdm:
This LGTM at first glance. We may want to consider normalizing cookies in the future so that we parse each cookie separately vs the entire Set-Cookie header, as the order of cookies can change. Any anchors (^) might not be valid when there is an application that is set first in the set.

PR #355, opened 2021-05-05
Palo Alto Tweaks

tsellers-r7:
## Description
This PR tweaks the fields for Palo Alto so as to return better data and/or improve the ability to generate CPEs.

I would have preferred to change `PANOS` to `PAN-OS` for the `os.product` but I wasn't sure what this would break downstream.

I also reset and regenerated the contents of the `identifiers/` directory.

## Checklist:

### DIFF
--- a/cpe-remap.yaml
+++ b/cpe-remap.yaml
@@ -160,8 +160,6 @@ mappings:
       owncloud_server: owncloud
   palo_alto_networks:
     vendor: paloaltonetworks
-    products:
-      panos: pan-os
   parallels:
     products:
       plesk: parallels_plesk_panel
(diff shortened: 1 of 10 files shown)


tsellers-r7:
CC @hdm  since I've adjusted some OS values and names.

hdm:
Thanks! Looks good to me!

rkirk-r7 (commented):
Nexpose currently expects `PAN-OS` for product and family so would be a good change for us at least. Could we add `os.family` field too? Nexpose checks depend on that field.

tsellers-r7:
FYI @hdm  we're going ahead with the switch from `PANOS` to `PAN-OS`.

rkirk-r7 (approved):

PR #356, opened 2021-05-06
Add new database dhcp_vendor_class

modoyle-r7:
Enables device identification using broadcast DHCP messages.

Uses strings from the vi_vendor_class field in DHCP client
discover, request and inform messages.

Added sample fingerprints with examples.
Associated changes to the identifiers directory

## Checklist:


xml/dhcp_vendor_class.xml
```diff
@@ -0,0 +1,215 @@
...
+    The vi_vendor_class field can be found in client discover (1), request (3)
+    and inform (8) messages.
+  -->
+
+  <fingerprint pattern="^Mfg=(?:Fuji)?(?i:Xerox);Typ=(?:MFP|printer);Mod=(?:Xerox )?(\S+) ([a-zA-Z0-9]+)">
+    <description>Xerox Multifunction Printer</description>
+    <example hw.family="VersaLink" hw.model="C405">Mfg=Xerox;Typ=MFP;Mod=VersaLink C405;Ser=</example>
+    <example hw.family="AltaLink" hw.model="C8035">Mfg=Xerox;Typ=MFP;Mod=Xerox AltaLink C8035 Multifunction Printer;Ser=</example>
+    <example hw.family="WorkCentre" hw.model="3345">Mfg=XEROX;Typ=MFP;Mod=WorkCentre 3345;Ser=3BT844742;Loc=</example>
+    <example hw.family="Phaser" hw.model="6500DN">Mfg=FujiXerox;Typ=printer;Mod=Phaser 6500DN;Ser=YXE135427</example>
```
tsellers-r7, line 15: We will probably want to capture the serial numbers here but that can wait until this Issue is sorted out:
https://github.com/rapid7/recog/issues/364

xml/dhcp_vendor_class.xml
```diff
@@ -0,0 +1,215 @@
...
+    <param pos="0" name="hw.device" value="Printer"/>
+    <param pos="0" name="hw.vendor" value="Xerox"/>
+    <param pos="1" name="hw.family"/>
+    <param pos="2" name="hw.model"/>
+    <param pos="0" name="hw.product" value="{hw.family} {hw.model}"/>
```
tsellers-r7, line 20: Constructing `hw.product` will work well for these printers since there is so much variation. That being said, it will prevent CPE values for being generated dynamically.  If we see specific models that are seen at high volumes and/or that have high risk vulnerabilities we may wish to construct a specific fingerprint for them with a static product. That will allow CPE values to be generated.

xml/dhcp_vendor_class.xml
```diff
@@ -0,0 +1,215 @@
...
+    <param pos="0" name="os.vendor" value="Xerox"/>
+  </fingerprint>
+
+  <fingerprint pattern="^Mfg=DELL;Typ=Printer;Mod=Dell (\S+) Laser Printer;">
+    <description>DELL Laser Printer</description>
+    <example hw.model="B2360dn">Mfg=DELL;Typ=Printer;Mod=Dell B2360dn Laser Printer;Ser=</example>
```
tsellers-r7, line 26: If you have an example with a serial number please go ahead and add the serial number value so that we can capture it in the future.  

My preference would be that all serial number values be modified so that they aren't a serial number that we've actually observed but ARE representative in content and length.

xml/dhcp_vendor_class.xml
```diff
@@ -0,0 +1,215 @@
...
+  </fingerprint>
+
+  <fingerprint pattern="^Mfg=LEXMARK;Typ=(?:MFP|Printer);Mod=Lexmark (\S+);Ser=">
+    <description>Lexmark Printer</description>
+    <example hw.model="MX410de">Mfg=LEXMARK;Typ=MFP;Mod=Lexmark MX410de;Ser=70155PLM0Y5YD;</example>
+    <example hw.model="MS310dn">Mfg=LEXMARK;Typ=Printer;Mod=Lexmark MS310dn;Ser=451432LM0X7XY;</example>
+    <param pos="0" name="hw.device" value="Printer"/>
+    <param pos="0" name="hw.vendor" value="Lexmark"/>
+    <param pos="1" name="hw.model"/>
+    <param pos="0" name="os.vendor" value="Lexmark"/>
```
tsellers-r7, line 70: For devices like printers the `os` is going to be a device or model specific firmware. In those cases go ahead and set `os.device` to `Printer` please.

xml/dhcp_vendor_class.xml
```diff
@@ -0,0 +1,215 @@
...
+    <param pos="0" name="hw.device" value="Printer"/>
+    <param pos="0" name="hw.vendor" value="Canon"/>
+    <param pos="0" name="hw.family" value="imageRunner"/>
+    <param pos="1" name="hw.model"/>
+    <param pos="0" name="hw.product" value="{hw.family} {hw.model}"/>
+    <param pos="0" name="os.vendor" value="Canon"/>
+  </fingerprint>
+
+  <fingerprint pattern="Canon (D\d+) Series">
+    <description>Canon imageClas Printer</description>
```
tsellers-r7, line 88: Typo, missing an `s` in `imageClass`

xml/dhcp_vendor_class.xml
```diff
@@ -0,0 +1,215 @@
...
+    <param pos="0" name="hw.vendor" value="Canon"/>
+    <param pos="0" name="hw.family" value="imageClass"/>
+    <param pos="1" name="hw.model"/>
+    <param pos="0" name="hw.product" value="{hw.family} {hw.model}"/>
+    <param pos="0" name="os.vendor" value="Canon"/>
+  </fingerprint>
+
+  <fingerprint pattern="^Polycom-(VVX\S+)">
+    <description>Polycom IP Phone</description>
+    <example hw.model="VVX410">Polycom-VVX410</example>
```
tsellers-r7, line 100: In this case please go ahead and set `hw.product` to be the same capture you are using for `hw.model`. That will enable auto population of the CPE data and result in values such as

```text
h:polycom:vvx400
h:polycom:vvx401
h:polycom:vvx410
h:polycom:vvx411
h:polycom:vvx450
h:polycom:vvx500
h:polycom:vvx501
h:polycom:vvx600
h:polycom:vvx601
```

xml/dhcp_vendor_class.xml
```diff
@@ -0,0 +1,215 @@
...
+    <param pos="0" name="hw.vendor" value="Polycom"/>
+    <param pos="0" name="hw.family" value="VVX"/>
+    <param pos="1" name="hw.model"/>
+    <param pos="0" name="os.vendor" value="Polycom"/>
+  </fingerprint>
+
+  <fingerprint pattern="^Cisco Systems, Inc. IP Phone (CP-\d+G?)\.*">
+    <description>Cisco IP Phone</description>
+    <example hw.model="CP-7821">Cisco Systems, Inc. IP Phone CP-7821\u0000</example>
+    <example hw.model="CP-7911G">Cisco Systems, Inc. IP Phone CP-7911G\u0000</example>
```
tsellers-r7, line 111: Is that trailing null included in the length of the field or is an artifact of the capturing process?

xml/dhcp_vendor_class.xml
```diff
@@ -0,0 +1,215 @@
...
+    <description>Linux</description>
+    <example os.version="4.14.78" os.arch="armv7l">dhcpcd-6.11.5:Linux-4.14.78:armv7l:Freescale</example>
+    <example os.version="4.19.155" os.arch="x86_64">dhcpcd-6.8.2:Linux-4.19.155-10581-g8bdb5ed8e80c:x86_64:GenuineIntel</example>
+    <param pos="0" name="os.family" value="Linux"/>
+    <param pos="0" name="os.product" value="Linux"/>
+    <param pos="1" name="os.version"/>
+    <param pos="2" name="os.arch"/>
+  </fingerprint>
+
+  <fingerprint pattern="^SAMSUNG Network Printer">
```
tsellers-r7, line 193: Here and elsewhere: If the example is representative of the whole contents of the field please go ahead and add the end of string match, `$`, to the regex. I'd much rather fail to match something (false negative) than match something we shouldn't (false positive).  We can always pickup false negatives by generating metrics on un-matched strings while automatically detecting false positives is much harder.

  That will also help us collect additional examples to ensure our regex' are robust.

xml/dhcp_vendor_class.xml
```diff
@@ -0,0 +1,215 @@
...
+    <param pos="0" name="hw.vendor" value="AXIS"/>
+    <param pos="1" name="hw.model"/>
+    <param pos="0" name="os.vendor" value="AXIS"/>
+    <param pos="2" name="os.version"/>
+  </fingerprint>
+
+  <fingerprint pattern="^PCoIP Endpoint$">
+    <description>PCoIP Endpoint Device</description>
+    <example>PCoIP Endpoint</example>
+    <param pos="0" name="hw.device" value="Remote Desktop"/>
```
tsellers-r7, line 167: I recommend that this be changed to `Thin Client` here and in `hw_device.txt` above. It makes it clearer which end of the service we are fingerprinting. If I see `Remote Desktop` I'm more likely to think of the server end of the service.

tsellers-r7 (changes requested):
Thanks for the PR @modoyle-r7 . Apologies for the extreme delay in response.   I've left some comments. We can discuss them further over slack.

xml/dhcp_vendor_class.xml
```diff
@@ -0,0 +1,215 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<fingerprints matches="dhcp_vendor_class" protocol="dhcp" database_type="service">
+  <!--
+    Fingerprint definitions that are matched against the string values in the
+    dhcp message vi_vendor_class field
+    The vi_vendor_class field can be found in client discover (1), request (3)
+    and inform (8) messages.
+  -->
```
tsellers-r7, line 8: If you don't mind, please add wording to indicate that this field is Option 60 as defined in RFC 2132 9.13.

PR #358, opened 2021-05-10
HTTPS related fingerprints

tsellers-r7:
## Description
This PR includes fingerprint updates based on recent data. Most changes revolve around HTML titles or certificates.

*NOTE*: I have attempted to normalize some of the VMware ESX / ESXi values around what already exists in the NTP and SNMP databases.  

- ESXi replaced ESX.  They were released side by side starting with version 4.1. All versions of 5.x (starting in 2010) were ESXi only.
- My preference would be to set the `os.family` to `vSphere` and the `os.product` to either `ESX` or `ESXi` as appropriate but I do not want to break any existing dependencies. 

## Types of changes
- Fingerprints

## Checklist:

### DIFF
--- a/xml/html_title.xml
+++ b/xml/html_title.xml
@@ -27,14 +27,32 @@
     <example>403 Forbidden</example>
   </fingerprint>
 
-  <fingerprint pattern="^404 Not Found$">
+  <fingerprint pattern="^(?:404 )?Not Found$">
     <description>404 Not Found - generic -- assert nothing.</description>
     <example>404 Not Found</example>
+    <example>Not Found</example>
   </fingerprint>
 
   <fingerprint pattern="^Invalid URL$">
     <description>Invalid URL - generic -- assert nothing.</description>
     <example>Invalid URL</example>
+  </fingerprint>
+
+  <fingerprint pattern="^ERROR: The request could not be satisfied$">
+    <description>Amazon CloudFront web load balancer endpoint</description>
+    <example>ERROR: The request could not be satisfied</example>
+    <param pos="0" name="service.vendor" value="Amazon"/>
+    <param pos="0" name="service.family" value="CloudFront"/>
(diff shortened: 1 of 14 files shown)


tsellers-r7:
@hdm  - FYI for VMware normalization (see comment) and cert improvements.

xml/html_title.xml
```diff
@@ -2291,7 +2465,8 @@
     <param pos="0" name="os.device" value="Lights Out Management"/>
     <param pos="0" name="os.vendor" value="Oracle"/>
     <param pos="0" name="os.family" value="ILOM"/>
-    <param pos="0" name="os.product" value="ILOM"/>
+    <param pos="0" name="os.product" value="Integrated Lights Out Manager firmware"/>
```
hdm, line 2468: Does this need the word "firmware" at the end?

xml/x509_issuers.xml
```diff
@@ -135,7 +204,8 @@
     <description>VMware ESXi w/Installer</description>
     <example>O=VMware Installer</example>
     <param pos="0" name="os.vendor" value="VMware"/>
-    <param pos="0" name="os.product" value="ESXi"/>
+    <param pos="0" name="os.family" value="VMware ESX/ESXi"/>
+    <param pos="0" name="os.product" value="VMware ESXi Server"/>
```
hdm, line 208: Any reason there is a vendor in the family/product field here? We don't put Cisco into the ASA variant above this.

tsellers-r7: I only used that value to be consistent with the existing values in the NTP database. My preference would be
`os.family` = `vSphere`
`os.product` = `ESXi`

rkirk-r7: `os.family` = `VMware ESX/ESXi`
`os.product` = `VMware ESXi Server`
these match Nexpose expectations so if you plan to change them let us know as we would need to introduce some mappings

rkirk-r7 (approved):

Issue #343, opened 2021-02-19
Deprecated CPEs


p0lr:
**Describe the bug**
I noticed deprecated CPEs in the XML fingerprints.

Looking at /xml/http_servers.xml
Definition for IIS 8.5 (for example) includes:
<param pos="0" name="service.cpe23" value="cpe:/a:microsoft:iis:8.5"/>
As stated in the other bug I opened, this is in cpe 2.2 format.  Converting it to CPE 2.3 as follows:

cpe:2.3:a:microsoft:iis:8.5:*:*:*:*:*:*:*

Submit it to NVD:
https://services.nvd.nist.gov/rest/json/cpes/1.0?cpeMatchString=cpe:2.3:a:microsoft:iis:8.5:*:*:*:*:*:*:*

Response is as follows:
{"resultsPerPage":20,"startIndex":0,"totalResults":0,"result":{"dataType":"CPE","feedVersion":"1.0","cpeCount":0,"feedTimestamp":"2021-02-19T03:40","cpes":[]}}

Essentially unrecognized.  There was a deprecation window where the service returned a "deprecatedBy" flag that told how to update the CPE.

Using the replacement CPE:
https://services.nvd.nist.gov/rest/json/cpes/1.0?cpeMatchString=cpe:2.3:a:microsoft:internet_information_services:8.5:*:*:*:*:*:*:*

Response:
{"resultsPerPage":20,"startIndex":0,"totalResults":1,"result":{"dataType":"CPE","feedVersion":"1.0","cpeCount":1,"feedTimestamp":"2021-02-19T03:43","cpes":[{"deprecated":false,"cpe23Uri":"cpe:2.3:a:microsoft:internet_information_services:8.5:*:*:*:*:*:*:*","lastModifiedDate":"2018-12-11T18:56Z","titles":[{"title":"Microsoft Internet Information Services (IIS) 8.5","lang":"en_US"}],"refs":[{"ref":"https://www.iis.net/","type":"Product"}],"deprecatedBy":[],"vulnerabilities":[]}]}}

If cpeCount is 0, there is a problem with the CPE.  Might also be good to keep an eye on the deprecatedBy response when performing validation checking.

You can pull down the entire CPE 2.3 dictionary for validation checking here: https://nvd.nist.gov/feeds/xml/cpe/dictionary/official-cpe-dictionary_v2.3.xml.gz

[...]

PR #361, opened 2021-05-12
CPE: stop using deprecated CPEs, split the remap YAML into sections

tsellers-r7:
## Description
This PR addresses an issue brought up by @p0lr in Issue #343. In that issue @p0lr points out that we're using deprecated CPE values and that the NIST database that we're already using indicates that the values are deprecated.

I've adjusted the matching logic in `update_cpes.py` so that these deprecated CPEs aren't added when we build our list of NIST issued CPEs.

I've also restructured `cpe-remap.yaml` so that we remap values are specific to a CPE type (`a`, `o`, `h`). This will allow us to remap certain older records that use the same `.product` value but for which the CPE would be different between `service.product` and `os.product`.

For example in the fingerprint below we couldn't have created a remap for `ILOM` under `oracle` because it would have then generated (or tried to generate) the same CPE for both `hw.product` and `os.product`.

```xml
<fingerprint pattern="^(?:Oracle\(R\) )?Integrated Lights Out Manager$">
    <description>Oracle iLOM</description>
    <example>Oracle(R) Integrated Lights Out Manager</example>
    <example>Integrated Lights Out Manager</example>
    <param pos="0" name="hw.device" value="Lights Out Management"/>
    <param pos="0" name="hw.vendor" value="Oracle"/>
    <param pos="0" name="hw.family" value="ILOM"/>
    <param pos="0" name="hw.product" value="ILOM"/>
    <param pos="0" name="os.device" value="Lights Out Management"/>
    <param pos="0" name="os.vendor" value="Oracle"/>
    <param pos="0" name="os.family" value="ILOM"/>
    <param pos="0" name="os.product" value="ILOM"/>
  </fingerprint>
```

With the new system we can remap them independently which will help us generate CPEs for existing values without having to change them.

## Motivation and Context
Correctness of results.

## How Has This Been Tested?
Execution of `cpe-remap.yaml`, rspec, etc.

## Types of changes

## Checklist:

### DIFF
--- a/identifiers/service_product.txt
+++ b/identifiers/service_product.txt
@@ -125,6 +125,7 @@ ESMTP
 EWS
 Ecelerity Mail Server
 Elastic Load Balancer
+Elastic Load Balancing
 EmWeb
 Email Appliance
 Email Security
@@ -277,7 +278,7 @@ MetaDirectory Server
 Metabase
 Metasploit
 MiniDLNA
-MiniUPnP
+MiniUPnPd
 MobaXterm
 MoinMoin
 Mongoose
@@ -389,6 +390,7 @@ Recursor
 Red Hat Directory Server
 Redmine
 Reflection
+Reflection for Secure IT
 ReflectionX
 RemoteView
 Resin
--- a/xml/html_title.xml
+++ b/xml/html_title.xml
@@ -44,6 +44,7 @@
     <param pos="0" name="service.vendor" value="Amazon"/>
     <param pos="0" name="service.family" value="CloudFront"/>
     <param pos="0" name="service.product" value="CloudFront Load Balancer"/>
+    <param pos="0" name="service.cpe23" value="cpe:/a:amazon:amazon_cloudfront:-"/>
   </fingerprint>
 
   <fingerprint pattern="^Microsoft Azure Web App - Error 404$">
@@ -965,11 +966,12 @@
     <param pos="0" name="hw.device" value="Lights Out Management"/>
     <param pos="0" name="hw.vendor" value="HP"/>
     <param pos="0" name="hw.product" value="iLO"/>
-    <param pos="0" name="hw.cpe23" value="cpe:/h:hp:integrated_lights_out:-"/>
+    <param pos="0" name="hw.cpe23" value="cpe:/h:hp:integrated_lights-out:-"/>
     <param pos="0" name="os.vendor" value="HP"/>
     <param pos="0" name="os.device" value="Lights Out Management"/>
     <param pos="0" name="os.family" value="iLO"/>
     <param pos="0" name="os.product" value="iLO"/>
+    <param pos="0" name="os.cpe23" value="cpe:/o:hp:integrated_lights-out_firmware:-"/>
   </fingerprint>
 
   <fingerprint pattern="^HP Integrated Lights-Out 2$">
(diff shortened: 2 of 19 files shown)


identifiers/service_product.txt
```diff
@@ -277,7 +278,7 @@ MetaDirectory Server
 Metabase
 Metasploit
 MiniDLNA
-MiniUPnP
+MiniUPnPd
```
tsellers-r7, line 281: There is actually a `MiniUPnPd` (daemon) and `MiniUPnPc` (client) so I changed this product value.

xml/html_title.xml
```diff
@@ -44,6 +44,7 @@
     <param pos="0" name="service.vendor" value="Amazon"/>
     <param pos="0" name="service.family" value="CloudFront"/>
     <param pos="0" name="service.product" value="CloudFront Load Balancer"/>
+    <param pos="0" name="service.cpe23" value="cpe:/a:amazon:amazon_cloudfront:-"/>
   </fingerprint>
```
tsellers-r7, line 48: This as well as the new CPE for Akamai Ghost and a couple of AWS services is a result of requesting NIST to issue them. Thanks NIST!

xml/html_title.xml
```diff
@@ -1695,10 +1710,11 @@
...
 
   <fingerprint pattern="^Login - OpenStack Dashboard$">
-    <description>OpenStack Dashboard</description>
+    <description>OpenStack Horizon Dashboard</description>
     <example>Login - OpenStack Dashboard</example>
     <param pos="0" name="service.vendor" value="OpenStack"/>
-    <param pos="0" name="service.product" value="Dashboard"/>
+    <param pos="0" name="service.product" value="Horizon"/>
```
tsellers-r7, line 1716: Incorrect product name but calling it out here in case anyone depends on it.

xml/html_title.xml
```diff
@@ -2492,7 +2526,7 @@
     <param pos="0" name="os.device" value="Lights Out Management"/>
     <param pos="0" name="os.vendor" value="Oracle"/>
     <param pos="0" name="os.family" value="ILOM"/>
-    <param pos="0" name="os.product" value="Integrated Lights Out Manager firmware"/>
+    <param pos="0" name="os.product" value="ILOM"/>
```
tsellers-r7, line 2529: @hdm  - Returning this value to what it was earlier this week. You'd asked about the need to add ` firmware` to the end of the value in that PR.  I didn't like it at the time but I didn't have decent options.  With the current PR this is now trivial.

xml/http_servers.xml
```diff
@@ -1364,13 +1364,13 @@
     <param pos="0" name="service.vendor" value="HP"/>
     <param pos="0" name="service.product" value="iLO"/>
     <param pos="0" name="service.family" value="iLO"/>
-    <param pos="0" name="service.cpe23" value="cpe:/a:hp:integrated_lights_out:-"/>
```
tsellers-r7, line 1367: This was deprecated w/o a replacement.

xml/x509_subjects.xml
```diff
@@ -1195,6 +1199,7 @@
     <example>CN=a248.e.akamai.net,O=Akamai Technologies\, Inc.,L=Cambridge,ST=Massachusetts,C=US</example>
     <param pos="0" name="service.vendor" value="Akamai"/>
     <param pos="0" name="service.product" value="GHost"/>
+    <param pos="0" name="service.cpe23" value="cpe:/a:akamai:akamaighost:-"/>
```
tsellers-r7, line 1202: This entry alone will result in CPEs for 9.5+ million endpoints..

tsellers-r7:
CC @hdm  @p0lr

xml/html_title.xml
rkirk-r7: Nexpose side does not depend on these

rkirk-r7 (approved):

Issue #364, opened 2021-07-26
Proposal: migrate host.id  to hw.serial_number for storing hardware serial numbers


tsellers-r7:
**Proposal**: Store hardware serial numbers in `hw.serial_number` instead of `host.id`.

In some cases recog is using `host.id` to store hardware serial numbers when they are captured from banners.  I would like to change this, where appropriate, to a _*new field*_ `hw.serial_number`. This would make it more clear as to what is being captured and move the data to a more appropriate context (`hw`).

https://github.com/rapid7/recog/search?q=host.id

In the cases where we have existing vendor specific serial numbers I would duplicate these values in `hw.serial_number` unless stakeholders indicate that these are not being used at all.

https://github.com/rapid7/recog/search?q=serial_number

Any feedback on this change would be greatly appreciated!

CC:  @hdm @dabdine

hdm:
+1 here, thanks!

rkirk-r7:
+1 too

tsellers-r7:
I have opened PR #366  to implement this change.

tsellers-r7:
The change has been landed. Thanks for the feedback here and elsewhere folks.

PR #366, opened 2021-07-27
Field migration: host.id -> hw.serial_number

tsellers-r7:
## Description
This PR implements the change discussed in Issue #364. 

The changes:
- changes `host.id` to `hw.serial_number` where the value is a hardware serial number
- changes `host.id` to `host.name` where the value references a host name
- in each case where there is a vendor specific serial number field such as `cisco.serial_number` that value is duplicated in `hw.serial_number` and a test added for it.

## Motivation and Context
The intent is to add clarity to the data that is captured. The field `host.id` was unclear and was being used to capture hostnames as well as serial numbers.

## How Has This Been Tested?
`rspec`, fingerprint example captures.

## Types of changes
- Field rename

## Checklist:

### DIFF
--- a/cpe-remap.yaml
+++ b/cpe-remap.yaml
@@ -309,6 +309,10 @@ mappings:
     hp:
       products:
         ilo: integrated_lights-out
+    kace:
+      vendor: dell
+      products:
+        k1000: kace_k1000_systems_management_appliance
     tandberg:
       vendor: cisco
     ubiquiti:
--- a/identifiers/os_product.txt
+++ b/identifiers/os_product.txt
@@ -116,7 +116,6 @@ IPReach
 IPSO
 IRIX
 Integrated Lights Out Manager
-Integrated Lights Out Manager firmware
 Irix
 Isilon OneFS OS
 JetDirect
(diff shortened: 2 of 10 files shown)


identifiers/os_product.txt
```diff
@@ -116,7 +116,6 @@ IPReach
 IPSO
 IRIX
 Integrated Lights Out Manager
-Integrated Lights Out Manager firmware
```
tsellers-r7, line 119: I rebuilt all of the `identifiers/` files. Apparently this value was no longer used.

lib/recog/nizer.rb
```diff
@@ -264,84 +264,3 @@ def self.best_service_match(matches)
 
 end
 end
-
```
tsellers-r7, line 267: Out of date comment replaced by the data in the `identifiers/` directory

cpe-remap.yaml
```diff
@@ -309,6 +309,10 @@ mappings:
     hp:
       products:
         ilo: integrated_lights-out
+    kace:
+      vendor: dell
+      products:
+        k1000: kace_k1000_systems_management_appliance
```
tsellers-r7, line 315: opportunistic CPE improvement

tsellers-r7:
cc @hdm  @dabdine

ekelly-rapid7 (approved):

rkirk-r7 (approved):

hdm:
I agree with changing to `hw.serial_number` as a consistent field, but there may be some adjustment needed on the Rumble side (if not Nexpose too).

Issue #138, opened 2017-06-15
Add tests to ensure portability of regex syntax


gschneider-r7:
There are some regexes using Ruby's `\h` syntax for hexadecimal values, but in other runtimes `\h` means horizontal whitespace. These regexes will not match correctly outside of Ruby. Instead we should be using `[a-f\d]` (unless someone has a better idea). 

There should be test coverage to ensure that the regexes are not relying on special Ruby-only features so we can avoid any issues with this in the future.

Examples:
https://github.com/rapid7/recog/blob/master/xml/ftp_banners.xml#L219
https://github.com/rapid7/recog/blob/master/xml/ftp_banners.xml#L1126
https://github.com/rapid7/recog/blob/master/xml/ftp_banners.xml#L1136
https://github.com/rapid7/recog/blob/master/xml/ftp_banners.xml#L1194

PR #367, opened 2021-07-28
Regex cross-language compatibility

tsellers-r7:
## Description
The intent of this PR is to improve cross language compatibility for the regex patterns used. 

The changes:

- Moves inline flags such as `(?i)` and `(?m)` to the beginning of the regex.  In Python these flags have global meaning and use of them anywhere other than the start of the pattern is deprecated. This does not change the function of these flags in current patterns for other languages.

- Replaces the use of POSIX character class indicators (`[[:alnum:]]`, `[[:digit:]`, and `[[:space:]`) with either the short backslash version (`\d`, `\s`) or a character range (`a-zA-Z0-9`).  The reason for this is that certain languages (Java, Python, Javascript) don't support the use of POSIX character classes.

## How Has This Been Tested?
`rspec`, 3rd party regex testors.

## Types of changes
Fingerprint construction

## Checklist:

### DIFF
--- a/xml/html_title.xml
+++ b/xml/html_title.xml
@@ -353,7 +353,7 @@
     <param pos="0" name="os.cpe23" value="cpe:/o:microsoft:windows:-"/>
   </fingerprint>
 
-  <fingerprint pattern="^(?i)Default (?:Parallels )?Plesk (?:Panel )?Page$">
+  <fingerprint pattern="(?i)^Default (?:Parallels )?Plesk (?:Panel )?Page$">
     <description>Plesk web hosting platform with no version</description>
     <example>Default Parallels Plesk Panel Page</example>
     <example>Default Parallels Plesk Page</example>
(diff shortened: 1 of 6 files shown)


tsellers-r7:
CC @hdm @pberry25 @dabdine

rkirk-r7 (approved):

hdm:
Seems great here! Any chance we can get a test case to scan new regexes for the incompatible variants? ([:digit:], non-leading ?i, etc).

tsellers-r7:
@hdm - I'm leaning towards an effort to separate the fingerprint databases into their own repo and then have language/tool specific tests that ensure that we know that patterns are valid in Ruby, Go, Java, etc and also pass the build in tests vs examples.

gschneider-r7:
See #138 for when tests for this can be accomplished :smiley:

gschneider-r7 (approved):

PR #369, opened 2021-08-02
Add explicit certainty for Assert Nothings

tsellers-r7:
## Description
This PR explicitly sets the certainty levels for certain `Assert Nothing` style fingerprint matches.  The intent is to convey that all other matches to a banner or endpoint should be given priority over these specific matches. For example, a http_server match may be too generic to have value but the x509 issuer may also match and we want that second one to win always.

## How Has This Been Tested?
`rspec`

## Checklist:

### DIFF
--- a/xml/dns_versionbind.xml
+++ b/xml/dns_versionbind.xml
@@ -17,30 +17,40 @@
   <fingerprint pattern="^$">
     <description>empty string -- assert nothing.</description>
     <example/>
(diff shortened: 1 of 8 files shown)


tsellers-r7:
CC @hdm @dabdine

hdm:
I think this should "0" and not "0.0" , since "0" already means "0%" (where 100 means 100%).

hdm:
Looks good to me! (deleted incorrect comment about 0 vs 0.0)

PR #400, opened 2022-01-26
Scripted update to add missing example attributes

dabdine:
## Description
Adds missing example attributes to `<example>` elements in recog fingerprint XML files, greatly reducing recog_verify warnings.

This is only round 1. Subsequent rounds will target some hand modifications to target fingerprints missing examples altogether. Additionally, there is one warning that is not fixed with this patch due to it being an edge case: The fingerprint regex conditionally matches a version, but the only example does not supply one. That will be fixed by hand as well in a subsequent PR.

The script that was used to make these changes is here:
https://gist.github.com/dabdine/6c641407e805ab33530ec67459a94c3a

## Motivation and Context
Clean up warnings so that more strict enforcement of quality can eventually be turned on.

## How Has This Been Tested?
Ran recog_verify

## Types of changes

- Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] I have updated the documentation accordingly (or changes are not required).
- [x] I have added tests to cover my changes (or new tests are not required).
- [x] All new and existing tests passed.

PR #404, opened 2022-01-28
Fix additional warnings due to untested attributes

dabdine:
## Description
Fixes these additional warnings that #400 did not address due to the examples missing content that could be parsed by the regular expression:

```
xml/http_servers.xml:1225: WARN: 'Sun Java System Application Server (formerly iPlanet Application Server, Sun ONE Application Server)' is missing an example that checks for parameter 'service.version' which is derived from a capture group
xml/operating_system.xml:59: WARN: 'Windows 2000' is missing an example that checks for parameter 'os.version' which is derived from a capture group
xml/operating_system.xml:228: WARN: 'Gentoo Linux' is missing an example that checks for parameter 'os.version' which is derived from a capture group
xml/operating_system.xml:663: WARN: 'IBM OSes' is missing an example that checks for parameter 'os.version' which is derived from a capture group
xml/operating_system.xml:679: WARN: 'HP OSes' is missing an example that checks for parameter 'os.version' which is derived from a capture group
xml/operating_system.xml:693: WARN: 'Juniper' is missing an example that checks for parameter 'os.version' which is derived from a capture group
xml/operating_system.xml:705: WARN: 'Cisco' is missing an example that checks for parameter 'os.version' which is derived from a capture group
xml/sip_user_agents.xml:286: WARN: 'Polycom RealPresence Trio Phones' is missing an example that checks for parameter 'host.mac' which is derived from a capture group
xml/smtp_banners.xml:541: WARN: 'MailEnable - Simple' is missing an example that checks for parameter 'host.name' which is derived from a capture group
```

Most of these (especially in operating_systems.xml) were made up to satisfy the regular expression, while others were incorporated using real (anonymized) data from internet scan engines.

After this patch, the only warnings that will remain are those that complain about fingerprints which have no test cases (`<example>` elements).

## Motivation and Context
Clean more warnings.

## How Has This Been Tested?
`bundle exec rake tests`

## Types of changes

## Checklist:

### DIFF
--- a/xml/operating_system.xml
+++ b/xml/operating_system.xml
@@ -683,7 +683,7 @@
     <description>HP OSes</description>
     <example os.product="HP-UX" os.family="HP-UX">HP-UX</example>
     <example os.product="OpenVMS" os.family="OpenVMS">OpenVMS</example>
-    <example os.product="OpenVMS" os.family="OpenVMS" os.version="1.0">OpenVMS 1.0</example>
+    <example os.product="OpenVMS" os.family="OpenVMS" os.version="6.2">OpenVMS 6.2</example>
     <param pos="0" name="os.vendor" value="HP"/>
     <param pos="1" name="os.family"/>
     <param pos="1" name="os.product"/>
--- a/xml/operating_system.xml
+++ b/xml/operating_system.xml
@@ -699,10 +699,11 @@
 
   <!-- Network equipment begin -->
 
-  <fingerprint pattern="^(?i:(?:Juniper\s?)?(Junos|Junos OS|ScreenOS)\s?(\d+?(?:\.\d+?)*?)?)$">
+  <fingerprint pattern="^(?i:(?:Juniper\s?)?(Junos|Junos OS|ScreenOS)\s?(\d+(?:\.\d+?)*(?:X\d{2})?)?)$">
     <description>Juniper</description>
     <example os.family="Junos" os.product="Junos">Junos</example>
-    <example os.family="Junos" os.product="Junos" os.version="1.0">Junos 1.0</example>
+    <example os.family="Junos" os.product="Junos" os.version="4.4">Junos 4.4</example>
+    <example os.family="Junos" os.product="Junos" os.version="12.1X44">Junos 12.1X44</example>
     <example os.family="ScreenOS" os.product="ScreenOS">ScreenOS</example>
     <param pos="0" name="os.vendor" value="Juniper"/>
     <param pos="1" name="os.family"/>
@@ -712,11 +713,14 @@
 
   <!-- This needs to be improved if it's not how one would generally present a Cisco OS version. -->
 
-  <fingerprint pattern="^(?i:(?:Cisco\s?)?(ASA|Adaptive Security Appliance|IOS|IOS-XE|IOS-XR|NX-OS|PIX-OS|SAN-OS)\s?(?:Version (\S+))?)$">
+  <fingerprint pattern="^(?i:(?:Cisco\s?)?(ASA|Adaptive Security Appliance|IOS|IOS-XE|IOS-XR|NX-OS|PIX-OS|SAN-OS)\s?(?:Version (\d+(?:\.\d+)*(?:\(\d+(?:\.\d+)*\)?)))?)$">
     <description>Cisco</description>
     <example os.family="ASA" os.product="ASA">Cisco ASA</example>
-    <example os.family="ASA" os.product="ASA" os.version="1.0">Cisco ASA Version 1.0</example>
     <example os.family="IOS" os.product="IOS">Cisco IOS</example>
+    <example os.family="ASA" os.product="ASA" os.version="7.0(1)">Cisco ASA Version 7.0(1)</example>
+    <example os.family="ASA" os.product="ASA" os.version="8.2(4.4)">Cisco ASA Version 8.2(4.4)</example>
+    <example os.family="ASA" os.product="ASA" os.version="8.3(2.25)">Cisco ASA Version 8.3(2.25)</example>
+    <example os.family="ASA" os.product="ASA" os.version="9.5(2.200)">Cisco ASA Version 9.5(2.200)</example>
     <param pos="0" name="os.vendor" value="Cisco"/>
     <param pos="1" name="os.family"/>
     <param pos="1" name="os.product"/>
--- a/xml/operating_system.xml
+++ b/xml/operating_system.xml
@@ -713,7 +713,7 @@
 
   <!-- This needs to be improved if it's not how one would generally present a Cisco OS version. -->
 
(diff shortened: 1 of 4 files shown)


xml/operating_system.xml
```diff
@@ -680,6 +683,7 @@
     <description>HP OSes</description>
     <example os.product="HP-UX" os.family="HP-UX">HP-UX</example>
     <example os.product="OpenVMS" os.family="OpenVMS">OpenVMS</example>
+    <example os.product="OpenVMS" os.family="OpenVMS" os.version="1.0">OpenVMS 1.0</example>
```
mkienow-r7, line 686: Is this a valid version number? All of the versions listed on [OpenVMS Releases](https://wiki.vmssoftware.com/OpenVMS#Releases) start with a "V".

mkienow-r7: I noticed the comment on the PR description that stated these were "made up to satisfy the regular expression".

mkienow-r7: We should ensure that all examples have provenance.

dabdine: Correct, I made this up -- though 1.0 is a valid version. Generally versions should be in numeric dotted notation for easier comparison ("is 1.2.3 less than 4.0"? is programmatically easier to ask than "is V1.2.3 less than 4.0?" -- which is done often for vulnerability testing), though I do see other places in recog that extract stuff like "V" before the version number. The exception to that rule is when you have old Cisco IOS versions like 12(T)13b (or whatever, I've really tried to erase those from my memory) where they don't really follow a dotted notation format. But if the purpose of "V" is only to denote that it is a version, there's no point in keeping it as part of the extracted version string IMHO.

dabdine: Correct, I made this up -- though 1.0 is a valid version. Generally versions should be in numeric dotted notation for easier comparison ("is 1.2.3 less than 4.0"? is programmatically easier to ask than "is V1.2.3 less than 4.0?" -- which is done often for vulnerability testing), though I do see other places in recog that extract stuff like "V" before the version number. The exception to that rule is when you have old Cisco IOS versions like 12(T)13b (or whatever, I've really tried to erase those from my memory) where they don't really follow a dotted notation format. But if the purpose of "V" is only to denote that it is a version, there's no point in keeping it as part of the extracted version string IMHO.

dabdine: Having issues today with github single comments & reviews 🤦

dabdine: Also note that "V1.0" wouldn't match the regular expression without modifying the regex to match a V. Just because it's listed in Wikipedia that way doesn't mean that was the originally intended way for it to be represented -- in fact I'd say it wasn't, since the original author (old schooler like myself or tas at rapid7) didn't include it

mkienow-r7: I agree the dotted notation portion makes sense for the parameter value. It does appear that the fingerprint regex is not correct. The table of version numbers that I linked was from a VMS Software wiki not from Wikipedia. I found documentation on [confirming the installation of the OpenVMS operating system](https://www.ibm.com/docs/en/ds8800?topic=attachment-confirming-installation-openvms-operating-system) that shows the version number reported with a leading "V".

```
$ show system
OpenVMS V7.3-2 on node DS20 5-FEB-2004 15:15:16.88 Uptime 1 01:45:09
```

dabdine: There is also documentation here:
https://marc.vos.net/books/vms/help/show/

That shows version 6.2 returning as:
```
$ SHOW SYSTEM
     OpenVMS 6.2 on node KRYPTN 14-DEC-1994 17:45:47:78 Uptime 2 21:53:59
```

However, regardless, this patch wasn't meant to make behavioral changes to any of the regular expressions. I don't think this regular expression was designed to process output from the `show system` command directly, as it is very limited in what it would match. Thus, I say we keep it as is, but happy to change 1.0 to 6.2.

dabdine: I'll add that I'd be fine adding a _conditional_ "V?" to the regular expression (not captured within the version capture group), but again, this regex doesn't appear to have been designed to specifically parse the output of `show system` (as it anchors the end of the string after the version, whereas the real output of `show system` continues on with a timestamp etc.).

mkienow-r7: Interesting find on the `SHOW SYSTEM` output. Thanks for making this one more flexible.

mkienow-r7:
Would you please take a look at the made up parameter values and ensure they have provenance?

xml/operating_system.xml
```diff
@@ -705,6 +715,7 @@
   <fingerprint pattern="^(?i:(?:Cisco\s?)?(ASA|Adaptive Security Appliance|IOS|IOS-XE|IOS-XR|NX-OS|PIX-OS|SAN-OS)\s?(?:Version (\S+))?)$">
     <description>Cisco</description>
     <example os.family="ASA" os.product="ASA">Cisco ASA</example>
+    <example os.family="ASA" os.product="ASA" os.version="1.0">Cisco ASA Version 1.0</example>
```
mkienow-r7, line 718: From a quick look at [Cisco ASA New Features by Release](https://www.cisco.com/c/en/us/td/docs/security/asa/roadmap/asa_new_features.html) it appears the first release was `ASA 7.0(1)` on May 31, 2005.

xml/operating_system.xml
```diff
@@ -693,6 +702,7 @@
   <fingerprint pattern="^(?i:(?:Juniper\s?)?(Junos|Junos OS|ScreenOS)\s?(\d+?(?:\.\d+?)*?)?)$">
     <description>Juniper</description>
     <example os.family="Junos" os.product="Junos">Junos</example>
+    <example os.family="Junos" os.product="Junos" os.version="1.0">Junos 1.0</example>
```
mkienow-r7, line 705: [Junos OS Dates & Milestones](https://support.juniper.net/support/eol/software/junos/) appears to show a limit history given the initial release of Junos OS was in 1998, but I didn't find anything quickly showing releases in 1998. Perhaps we use a more current version number from the list?

dabdine: Fixed the regular expression and added a couple examples.

xml/operating_system.xml
dabdine: Fixed the regular expression and added a couple of examples

dabdine:
The issue I have with operating_systems.xml in particular is that nothing is defined as to what the input should be for this file. Is it what's pulled from SNMP? Data from executing the equivalent of "show version" on the command line (where applicable)? The problem here is that, despite these changes adding more provenance to actual versions, they may not test real world scenarios.

I'll add that the Cisco fingerprint looks entirely broken if it is to be applied to the output of `show version`. The fingerprint was made to be generic, but won't parse complicated Cisco IOS versions like `12.2(33)SXI1`. If memory serves, in many of these cases some of the version info was received from SNMP.

mkienow-r7:
I appreciate you taking the time to enhance some of the fingerprints and adding more provenance. After spending more time looking at the fingerprints I agree that we need to make more sense of data source that to be used here and reassess this DB. 

PR #405, opened 2022-01-28
Enhance recog verify to report an untested parameter as a failure

mkienow-r7:
## Description
Enhances `recog_verify` to report an untested parameter as a failure instead of a warning.

## Motivation and Context
Ensure all fingerprints have examples that fully verify all capture group parameters.

## How Has This Been Tested?
* `rake tests`

## Types of changes

## Checklist:

### DIFF
--- a/features/data/tests_with_failures.xml
+++ b/features/data/tests_with_failures.xml
@@ -17,4 +17,10 @@
      <param pos="2" name="os.version" />
      <param pos="1" name="os.name" value="Bar" />
    </fingerprint>
+  <fingerprint pattern="^(\S+) ([\d.]+)$">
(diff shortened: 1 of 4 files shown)


dabdine:
LGTM.

Should we add an aruba test (in veatures/verify.feature + corresponding 'data' dir file) to validate the output as FAIL?

When I added the change to eliminate some of the warnings, I had to also update verify.feature to remove the WARN lines from the ProFTPd test case that would've otherwise failed. In retrospect, I should've separated them out to a separate test case and file in the data dir. That would've meant you'd have to have updated that file too.

tsellers-r7 (approved):

PR #406, opened 2022-02-01
Fix examples in architecture.xml, ftp_banners.xml, and ssh_banners.xml

dabdine:
## Description
Fixes fingerprint examples & regular expression issues in `ftp_banners.xml`, `architecture.xml`, and `ssh_banners.xml`. These files no longer yield any warnings whatsoever.

## Motivation and Context
Eliminate more warnings & fix bad regular expressions in some fingerprints.

## How Has This Been Tested?
`bundle exec rake tests`

## Types of changes

## Checklist:

### DIFF
--- a/xml/ftp_banners.xml
+++ b/xml/ftp_banners.xml
@@ -683,7 +683,7 @@ more text</example>
     <param pos="0" name="os.device" value="IP Camera"/>
   </fingerprint>
 
-  <fingerprint pattern="(?i)^AXIS (\S+) (?:(?:Mk II )?Video) (?:Encoder Blade|Encoder|Module|Server|Decoder) (\S+)">
+  <fingerprint pattern="(?i)^AXIS (\S+) (?:(?:Mk II )?Video) (?:Encoder(?: Blade)?|Module|Server|Decoder) (\S+)">
     <description>Axis Video encoders/servers</description>
     <example hw.product="Q7406" hw.version="5.01">AXIS Q7406 Video Encoder Blade 5.01 (Aug 01 2008) ready.</example>
     <example hw.product="241Q" hw.version="4.47.2">AXIS 241Q Video Server 4.47.2 (Dec 11 2008) ready.</example>
--- a/xml/ftp_banners.xml
+++ b/xml/ftp_banners.xml
@@ -202,7 +202,7 @@ example.com FTP server (Version:  Mac OS X Server) ready.</example>
 
   <fingerprint pattern="^ProFTPD (\d+\.[^\s]+) Server \(Linksys(W.+)\) \[(.+)\]$">
     <description>ProFTPD on a Linksys Wireless Access Point/Router</description>
-    <example service.version="1.3.0rc2" os.product="WRT350N" host.name="host">ProFTPD 1.3.0rc2 Server (LinksysWRT350N) [host]</example>
+    <example service.version="1.3.0rc2" os.product="WRT350N" hw.product="WRT350N" host.name="host">ProFTPD 1.3.0rc2 Server (LinksysWRT350N) [host]</example>
     <param pos="0" name="service.family" value="ProFTPD"/>
     <param pos="0" name="service.vendor" value="ProFTPD Project"/>
     <param pos="0" name="service.product" value="ProFTPD"/>
@@ -211,6 +211,8 @@ example.com FTP server (Version:  Mac OS X Server) ready.</example>
     <param pos="0" name="os.vendor" value="Linksys"/>
     <param pos="0" name="os.device" value="WAP"/>
     <param pos="2" name="os.product"/>
+    <param pos="0" name="hw.vendor" value="Linksys"/>
+    <param pos="2" name="hw.product"/>
     <param pos="3" name="host.name"/>
   </fingerprint>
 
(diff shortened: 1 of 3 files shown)


xml/ftp_banners.xml
```diff
@@ -693,9 +683,9 @@ more text</example>
     <param pos="0" name="os.device" value="IP Camera"/>
   </fingerprint>
 
-  <fingerprint pattern="(?i)^AXIS (\S+) (?:(?:Mk II )?Video) (?:Encoder|Encoder Blade|Module|Server|Decoder) (\S+)">
+  <fingerprint pattern="(?i)^AXIS (\S+) (?:(?:Mk II )?Video) (?:Encoder Blade|Encoder|Module|Server|Decoder) (\S+)">
     <description>Axis Video encoders/servers</description>
-    <example hw.product="Q7406" hw.version="Blade">AXIS Q7406 Video Encoder Blade 5.01 (Aug 01 2008) ready.</example>
+    <example hw.product="Q7406" hw.version="5.01">AXIS Q7406 Video Encoder Blade 5.01 (Aug 01 2008) ready.</example>
```
dabdine, line 688: cc @tsellers-r7 
https://github.com/rapid7/recog/pull/400#discussion_r794637958

xml/ftp_banners.xml
```diff
@@ -241,19 +243,6 @@ example.com FTP server (Version:  Mac OS X Server) ready.</example>
     <param pos="1" name="host.name"/>
   </fingerprint>
 
-  <fingerprint pattern="^ProFTPD (\d+\.[^\s]+) Server \(Linksys(.*)\) \[(.+)\]$">
```
dabdine, line 244: This looks like a duplicate of what's above. I can't find any examples that don't start with W (WRTxxx). Also, it fails if I keep it since the example in this one was matching the earlier fingerprint and causing `bundle exec rake tests` to fail.

xml/ftp_banners.xml
```diff
@@ -211,6 +211,8 @@ example.com FTP server (Version:  Mac OS X Server) ready.</example>
     <param pos="0" name="os.vendor" value="Linksys"/>
     <param pos="0" name="os.device" value="WAP"/>
```
dabdine, line 212: Is os.device a thing? I thought this was hw.device, but I see os.device everywhere.

dabdine:
Added one fix for ssh_servers.xml as well, but this file still has warnings to address.

xml/ftp_banners.xml
tsellers-r7: It _is_ a thing but it is used incorrectly more often than not. Also, in almost every case it should be paired with `hw.device`

xml/ftp_banners.xml
tsellers-r7: Oh, unexpected capture due to optional string ordering.  That's an interesting catch, thanks!

xml/ftp_banners.xml
tsellers-r7: The regex in both of the fingerprints remains unchanged since the initial commit 8 years ago. As you mentioned the only real difference is the `W` in the regex.

I've checked some of our datasets and I do NOT see any matches for this fingerprint in them.
I DO see matches for the `W` variant.

++ to seeing this FP removed.

xml/ssh_banners.xml
```diff
@@ -1732,6 +1732,7 @@
 
   <fingerprint pattern="^SSH Protocol Compatible Server SCS (.*)$">
     <description>Netscreen with version</description>
+    <example service.version="2.0">SSH Protocol Compatible Server SCS 2.0</example>
```
tsellers-r7, line 1735: I don't see this value in any of our datasets (I searched SSH and FTP data) but it has been documented on the internet and matches regex used in Nmap.

https://www.chiark.greenend.org.uk/~sgtatham/putty/wishlist/netscreen-command.html
https://rt.cpan.org/Public/Bug/Display.html?id=6256

I wonder if this was observed on an SFTP service or if it should have been in the SSH fingerprints file. Either way, it seems to be of little consequence.

xml/ftp_banners.xml
dabdine: any suggestions on what to do with this? You marked a few hw.device etc. in #400 that need to be corrected. We probably just need to audit the file specifically for all of those.

xml/ssh_banners.xml
dabdine: https://search.censys.io/search?resource=hosts&sort=RELEVANCE&per_page=25&virtual_hosts=EXCLUDE&q=services.ssh.endpoint_id.raw%3A%22SSH+Protocol+Compatible+Server+SCS+2.0%22

tsellers-r7: Thanks. yeah, we should make sure that we get a fingerprint for this in the SSH banners file.

dabdine: Not following -- this is ssh_banners.xml?

dabdine:
I went ahead and remediated all remaining SSH examples in ssh_banners.xml. This PR now closes out all issues on the three mentioned files.

xml/ftp_banners.xml
tsellers-r7: I'd add `hw.device` to this FP but honestly we need to perform an audit and fix them so either way is fine.

tsellers-r7 (approved):

mkienow-r7:

PR #408, opened 2022-02-01
Fix all examples in http_cookies.xml by adding/improving examples

dabdine:
## Description
Addresses all warnings by incorporating real data as examples into http_cookies.xml

## Motivation and Context
Eliminate warnings

## How Has This Been Tested?
`bundle exec rake tests`

## Types of changes

## Checklist:

### DIFF
--- a/xml/http_cookies.xml
+++ b/xml/http_cookies.xml
@@ -86,10 +86,25 @@
     <param pos="0" name="hw.device" value="VPN"/>
   </fingerprint>
 
-  <fingerprint pattern="^(Apache)=[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.([0-9]+);">
-    <description>Apache</description>
-    <param pos="1" name="cookie"/>
-    <param pos="2" name="system.time.micros"/>
+  <fingerprint pattern="^Apache=(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})\.([0-9]+(?:\.[0-9]+)?);">
+    <description>Apache with session ID containing IP and timestamp (timestamp can be micros, millis or seconds)</description>
+    <example host.ip="10.10.130.165" system.time="1643670182768255">Apache=10.10.130.165.1643670182768255; path=/</example>
+    <example host.ip="10.0.101.6" system.time="1643663969718158">Apache=10.0.101.6.1643663969718158; path=/; expires=Wed, 31-Jan-24 21:19:29 GMT; domain=.contoso.com</example>
+    <example host.ip="10.10.20.18" system.time="1643510579.1915">Apache=10.10.20.18.1643510579.1915; domain=foo.com; path=/; expires=Mon, 30-Jan-2023 02:42:58 GMT</example>
+    <example host.ip="10.23.219.241" system.time="1643541709604">Apache=10.23.219.241.1643541709604; path=/; expires=Tue, 19-Jan-2038 03:14:07 GMT</example>
+    <param pos="0" name="cookie" value="Apache"/>
+    <param pos="2" name="system.time"/>
+    <param pos="1" name="host.ip"/>
+    <param pos="0" name="service.vendor" value="Apache"/>
+    <param pos="0" name="service.family" value="Apache"/>
+    <param pos="0" name="service.product" value="HTTPD"/>
+    <param pos="0" name="service.cpe23" value="cpe:/a:apache:http_server:-"/>
+  </fingerprint>
+
+  <fingerprint pattern="^Apache=[0-9a-z]{8}\.[0-9a-z]{13};">
+    <description>Apache with opaque session ID</description>
+    <example>Apache=1148b9c3.5d6e61e36f2f9; path=/; domain=.foo.com</example>
+    <param pos="0" name="cookie" value="Apache"/>
     <param pos="0" name="service.vendor" value="Apache"/>
     <param pos="0" name="service.family" value="Apache"/>
     <param pos="0" name="service.product" value="HTTPD"/>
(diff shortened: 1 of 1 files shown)


xml/http_cookies.xml
```diff
@@ -86,10 +86,25 @@
     <param pos="0" name="hw.device" value="VPN"/>
   </fingerprint>
 
-  <fingerprint pattern="^(Apache)=[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.([0-9]+);">
-    <description>Apache</description>
-    <param pos="1" name="cookie"/>
-    <param pos="2" name="system.time.micros"/>
+  <fingerprint pattern="^Apache=(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})\.([0-9]+(?:\.[0-9]+)?);">
+    <description>Apache with session ID containing IP and timestamp (timestamp can be micros, millis or seconds)</description>
+    <example host.ip="10.10.130.165" system.time="1643670182768255">Apache=10.10.130.165.1643670182768255; path=/</example>
```
dabdine, line 91: This is a fun one:
1. The regular expression was storing the "system time" in `system.time.micros`, but the time stamp that this regular expression was matching comes back from Apache in seconds, milliseconds or microseconds. This seems overly configuration specific. I chose to store the value in `system.time` and leave it to the client application to interpret. A separate option was to explicitly parse microseconds/milliseconds/seconds as separate fingerprints, but this would slow down parsing and run into issues in the future when interpreting these timestamps as epoch timestamps (albeit, a bit far out). 
2. The regular expression didn't catch the "uuid-like" form of apache session cookies, which I added support for.

xml/http_cookies.xml
```diff
@@ -123,27 +138,20 @@
...
+    <example>WebLogicSession=QKRlJZbj0b948CrXnoQw8FNuSWvO6fXaJNadlcCWwA3qm6CtqD5a; path=/</example>
+    <param pos="0" name="cookie" value="WebLogicSession"/>
     <param pos="0" name="service.vendor" value="BEA"/>
     <param pos="0" name="service.family" value="WebLogic"/>
     <param pos="0" name="service.product" value="WebLogic"/>
     <param pos="0" name="service.cpe23" value="cpe:/a:bea:weblogic_server:-"/>
   </fingerprint>
 
-  <fingerprint pattern="^(BCSI-CSC[0-9A-Za-z]+)=">
+  <fingerprint pattern="^(BCSI-CS-[0-9A-Za-z]+)=">
```
dabdine, line 152: I could not find any examples--_anywhere_--which references `BCSI-CSC...` -- only `BCSI-CS-...`. So, I updated the regex with what I found from reality. If anyone has a clue whether this change will break something, please let me know so I can update it. 

I could also make this backward compatible just in case something breaks, but not sure what the value is if we can't prove that cookie actually exists.

xml/http_cookies.xml
```diff
@@ -365,6 +374,8 @@
 
   <fingerprint pattern="^IBMCBR=">
     <description>IBM WebSphere Load Balancer</description>
+    <!-- Replace with a valid example if one is discovered -->
+    <example>IBMCBR=fakevalue</example>
```
dabdine, line 378: Could not find a real value for this. Let me know if you find one.

mkienow-r7: I found the format, but not a concrete example in [WebSphere Application Server: Load Balancer Administration Guide - Version 6.0.1](https://public.dhe.ibm.com/software/webserver/appserv/library/v601/infocenter/edge/pdf/LBguide.pdf)

> Active cookie affinity is used to ensure that a client continues to be load balanced
to the same server for some period of time. This is accomplished by sending a
cookie to be stored by the clients browser. The cookie contains the cluster:port:rule
that was used to make the decision, the server that was load balanced to, and a
timeout timestamp for when the affinity is no longer valid. The cookie is in the
following format: IBMCBR=cluster:port:rule+server-time! The cluster:port:rule and
server information are encoded so the CBR configuration is not revealed.

dabdine: Noticed this as well, but I didn't bother trying to write something to conform as to not make the reader think that it actually is a valid value later.

xml/http_cookies.xml
dabdine: Actually, I think we _can_ and should split these into separate fingerprints per seconds/millis/micros. The integer lengths are far enough apart that it's improbable that these values will ever overlap (a regular expression matching a number of digits should be good enough precision to capture each timestamp).

tsellers-r7: Option 4: Decide we don't really care about the timestamp and ignore it  ;)

tsellers-r7:
@dabdine @mkienow-r7 - I'll try to see if I can find examples of some of the missing values in a recent dataset. It will be later today before I can do this.

xml/http_cookies.xml
tsellers-r7: A check of a couple studies vs 80 and 443 found only the `BCSI-CS-` version that you mentioned above. 

++ to this change.

xml/http_cookies.xml
tsellers-r7: No luck finding this when looking at a couple studies vs 80 and 443

xml/http_cookies.xml
```diff
@@ -517,8 +534,10 @@
     <param pos="0" name="service.product" value="Sage X3 Syracuse Web Server"/>
   </fingerprint>
 
-  <fingerprint pattern="^(gx_session_id|JROUTE)=">
+  <fingerprint pattern="^(gx_session_id|JROUTE)=" flags="REG_ICASE">
```
tsellers-r7, line 537: I looked across multiple datasets and in a couple places on the web and only found `GX_SESSION_ID`, not the lower case version.  We may wish to just change to that in the regex and remove the `REG_ICASE` flag.

xml/http_cookies.xml
dabdine: @tsellers-r7 I'm fine with that too. I'm still not super excited about expanding this into three regular expressions as that'll def slow it down... so changing my mind again, as I don't think the time adds enough value to offset the cost. I think we keep it as one. Happy to either remove the time capture group or leave as is.

tsellers-r7: I don't know any production use of the time value. I'm ok with removing it in order to improve this fingerprint.

CC: @mkienow-r7 @hdm @gschneider-r7 for comments needing to capture time and emit it in the recog results.

xml/http_cookies.xml
dabdine: will do this, thx for the 👀

xml/http_cookies.xml
dabdine: thx @mkienow-r7 and @tsellers-r7 -- will be keeping this as is.

xml/http_cookies.xml
mkienow-r7: I received confirmation via Slack from @hdm and @ekelly-rapid7 that we should be good to go removing `system.time` here.

mkienow-r7 (approved):

PR #409, opened 2022-02-07
Add phpMyAdmin fingerprints

mkienow-r7:
## Description
Adds 3 [phpMyAdmin](https://www.phpmyadmin.net/) fingerprints.

## Motivation and Context
Fingerprint more services.

## How Has This Been Tested?
* `rake tests`

## Types of changes

## Checklist:

### DIFF
--- a/xml/favicons.xml
+++ b/xml/favicons.xml
@@ -1962,4 +1962,12 @@
     <param pos="0" name="service.cpe23" value="cpe:/a:mitre:caldera:-"/>
   </fingerprint>
 
+  <fingerprint pattern="^531b63a51234bb06c9d77f219eb25553$">
+    <description>phpMyAdmin web interface for MySQL and MariaDB</description>
+    <example>531b63a51234bb06c9d77f219eb25553</example>
+    <param pos="0" name="service.vendor" value="phpMyAdmin"/>
+    <param pos="0" name="service.product" value="phpMyAdmin"/>
+    <param pos="0" name="service.cpe23" value="cpe:/a:phpmyadmin:phpmyadmin:-"/>
+  </fingerprint>
+
 </fingerprints>
--- a/xml/html_title.xml
+++ b/xml/html_title.xml
@@ -3738,17 +3738,19 @@
     <param pos="0" name="service.cpe23" value="cpe:/a:mitre:caldera:-"/>
   </fingerprint>
 
-  <fingerprint pattern="phpMyAdmin(?: ([\d.]+(?:-[a-zA-Z0-9]+)?(?:\+\d{8}\.[a-f0-9]{4,40})?))?">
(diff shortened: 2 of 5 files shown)


xml/html_title.xml
```diff
@@ -3738,4 +3738,18 @@
     <param pos="0" name="service.cpe23" value="cpe:/a:mitre:caldera:-"/>
   </fingerprint>
 
+  <fingerprint pattern="phpMyAdmin(?: (\d+\.\d+\.\d+(?:\.\d+)?(?:-[a-z0-9]+)?))?$">
+    <description>phpMyAdmin web interface for MySQL and MariaDB</description>
+    <example>phpMyAdmin</example>
+    <example service.version="2.10.0.2">phpMyAdmin 2.10.0.2</example>
+    <example service.version="5.0.0-rc1">phpMyAdmin 5.0.0-rc1</example>
+    <example service.version="5.0.4">10.10.10.10:8080 / db.foo.bar | phpMyAdmin 5.0.4</example>
```
dabdine, line 3746: should we extract 10.10.10.10 into a host.ip field like we do elsewhere? Same thing can be said for the host name below, though that's arguably less useful since it can be localhost or something internal (and it would also likely require splitting the regex which i'm not a fan of).

mkienow-r7: Yeah, I was thinking about doing that. The regex might be a little tricky in that it needs to capture both IPv4, IPv6 and hostnames dropping the `:port` portion.
The [title](https://docs.phpmyadmin.net/en/latest/config.html#page-titles) is loaded with information `'@HTTP_HOST@ / @VSERVER@ | @PHPMYADMIN@'` and also contains the MySQL server name.

From [What format strings can I use?](https://docs.phpmyadmin.net/en/latest/faq.html#what-format-strings-can-i-use)
```
@HTTP_HOST@
HTTP host that runs phpMyAdmin

@VSERVER@
Verbose MySQL server name if set, otherwise normal
```

xml/favicons.xml
```diff
@@ -1962,4 +1962,12 @@
     <param pos="0" name="service.cpe23" value="cpe:/a:mitre:caldera:-"/>
   </fingerprint>
 
+  <fingerprint pattern="^531b63a51234bb06c9d77f219eb25553$">
+    <description>phpMyAdmin web interface for MySQL and MariaDB</description>
+    <example>531b63a51234bb06c9d77f219eb25553</example>
+    <param pos="0" name="service.vendor" value="phpmyadmin"/>
```
tsellers-r7, line 1968: Here and elsewhere for `.vendor` and `.product`: This should be `phpMyAdmin`

mkienow-r7: Good catch!

tsellers-r7 (approved):

PR #382, opened 2021-10-29
Updates from RumbleDiscovery/recog fork

pbarry25:
## Description
Updates from the [RumbleDiscovery/recog](https://github.com/RumbleDiscovery/recog) fork. Mostly fingerprints, with addition of the ability to use filesystem-based examples for long examples.

## Motivation and Context
Looking to merge `RumbleDiscovery/recog` fork updates back to R7's repo, for community benefit.

## How Has This Been Tested?
- `recog_match`, `recog_verify`, `rake spec` as we went
- `rake tests` prior to PR creation here

## Types of changes
- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)

## Checklist:
- [x] I have updated the documentation accordingly (or changes are not required).
- [x] I have added tests to cover my changes (or new tests are not required).
- [x] All new and existing tests passed.

PR #410, opened 2022-02-07
Enable fingerprint xsd validation and CI automation

dabdine:
## Description
* Enables XML validation in recog_verify via XSD for fingerprint files using the already-existing fingerprints.xsd schema.
* Configures the XML extension in vscode to properly autocomplete and validate recog XML files if the extension is installed.
* Updates the background monitor to automatically validate via XSD as well

This does not change any behavior (yet) in Go or Java, however.

## Motivation and Context
Malformed XML is not handled well by recog_verify. In most cases, when a malformed document is produced, recog_verify will simply dump a stack trace. This change will now properly report validation errors as errors in the same format as example errors that were previously produced.

This should allow new PR submissions to pass more strict testing automatically, keeping the quality bar high while avoiding the burden on the reviewer for checking whether certain attributes & elements are present, and set to the right values.

## How Has This Been Tested?
`bundle exec rake tests`

## Types of changes

## Checklist:


dabdine:
Here's an example of an XML change (`_encoding="base65"`) which will break CI because base65 is not in the enumeration allowed for the `_encoding` attribute:

https://github.com/dabdine/recog/pull/1

Rakefile
```diff
@@ -14,12 +14,8 @@ end
 require 'cucumber'
 require 'cucumber/rake/task'
 
-def jruby?
```
dabdine, line 17: Backed this change out in favor of using cucumber's features/support/hooks.rb, which allows you to define annotations that can tag a scenario with code that executes before the test is run. In this case, I created two annotations that will instruct cucumber to skip the test. The result is functionally the same as what we had here, but the tests will be marked as skipped in the log output (before, the tests were not even read, so there was nothing in the cucumber run log about them).

bin/recog_verify
```diff
@@ -48,15 +55,45 @@ if ARGV.empty?
...
-    verifier = Recog::VerifierFactory.build(options, ndb)
-    verified = verifier.verify
-    failures += verifier.reporter.failure_count
-    warnings += verifier.reporter.warning_count
+    # Create a new reporter per XML file to hold context on success/warn/fails
+    reporter  = Recog::VerifyReporter.new(options, formatter, file)
+
+    begin
+      # Validate the XML database against the recog schema first, if requested
+      if options.schema_file
```
dabdine, line 74: This block will validate the XML schema and dump out errors prior to running `<example>` tests. If there are XML errors, the example tests will never run (as the XML is considered malformed anyway). Thus, it's possible to run this, get XML errors in the verify output log, then run it again and get errors with your fingerprints if you violate rules, like missing an example.

Also, we should require the example element at least once via the XSD file once all fingerprint files have them. That way those checks can be done simply through the XML schema rather than having recog_verify manually validate that examples are present.

lib/recog/verifier.rb
```diff
@@ -15,19 +15,19 @@ def verify
         fp.verify_params do |status, message|
           case status
           when :warn
-            reporter.warning "WARN: #{message}", fp.line
```
dabdine, line 18: I moved the WARN/FAIL tags for output into verify_reporter.rb instead, since that is what is actually logging the text anyway. 

Needed to do this so that we could share a formatter and reporter between the XML schema validator and recog's verifier (don't repeat yourself/DRY, yadda yadda).

bin/recog_verify
```diff
@@ -1,12 +1,15 @@
...
 require 'optparse'
 require 'ostruct'
 require 'recog'
-require 'recog/verifier_factory'
+require 'recog/formatter'
+require 'recog/verifier'
+require 'recog/verify_reporter'
 
-options = OpenStruct.new(color: false, detail: false, quiet: false, warnings: true)
+options = OpenStruct.new(color: false, detail: false, quiet: false, warnings: true, schema_file: nil)
```
dabdine, line 12: validation against the xsd is *not* on by default. You have to explicitly enable it. Figured this was the best option as to not require folks to now know where the xsd file is every time they want to use recog_verify.

bin/recog_verify
```diff
@@ -48,15 +55,45 @@ if ARGV.empty?
...
-    failures += verifier.reporter.failure_count
-    warnings += verifier.reporter.warning_count
+    # Create a new reporter per XML file to hold context on success/warn/fails
+    reporter  = Recog::VerifyReporter.new(options, formatter, file)
+
+    begin
+      # Validate the XML database against the recog schema first, if requested
+      if options.schema_file
+        errors = validate(file, options.schema_file)
+        if errors.length > 0
```
mkienow-r7, line 76: Minor: use `size` over `length`. See [rubocop/ruby-style-guide: map/find/select/reduce/include?/size](https://github.com/rubocop/ruby-style-guide#map-find-select-reduce-include-size)

bin/recog_verify
```diff
@@ -48,15 +55,45 @@ if ARGV.empty?
...
+  schema.validate(doc)
+end
+
 warnings = 0
 failures = 0
+formatter = Recog::Formatter.new(options, $stdout)
 ARGV.each do |arg|
   Dir.glob(arg).each do |file|
-    ndb = Recog::DB.new(file)
-    verifier = Recog::VerifierFactory.build(options, ndb)
```
mkienow-r7, line 56: I don't disagree with the removal of `VerifierFactory`, but would you share your thoughts behind the change?

xml/fingerprints.xsd
```diff
@@ -104,16 +104,23 @@
...
+  <xsd:simpleType name="encoding">
+    <xsd:restriction base="xsd:string">
+      <xsd:enumeration value="base64" />
+    </xsd:restriction>
+  </xsd:simpleType>
+
   <xsd:complexType name="example_element">
     <xsd:simpleContent>
       <xsd:extension base="xsd:string">
+        <xsd:attribute name="_encoding" type="encoding"/>
```
mkienow-r7, line 123: We should also validate the `_filename` attribute introduced in #382.

lib/recog/verifier.rb
mkienow-r7: Nice work pulling the label out and simplifying the `reporter` calls!

bin/recog_verify
```diff
@@ -35,6 +38,10 @@ option_parser = OptionParser.new do |opts|
     options.warnings = o
   end
 
+  opts.on("--schema-location SCHEMA_FILE", "Location of the Recog XSD schema file (default: ./xml/fingerprints.xsd). If not specified, validation will not be run.") do |schema_file|
```
mkienow-r7, line 41: Should we mention "default: ./xml/fingerprints.xsd" in the description if that isn't used by default?

mkienow-r7 (changes requested):

bin/recog_verify
```diff
@@ -48,15 +55,45 @@ if ARGV.empty?
   exit(1)
 end
 
+def validate(xmlfile_path, xsd_path)
+  schema = Nokogiri::XML::Schema(File.read(xsd_path))
```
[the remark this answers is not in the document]
dabdine: Nope, i'll update this

bin/recog_verify
dabdine: Removes a level of indirection and simplifies the code base. All it does is three things (so it doesn't really do anything special), and it's only used in one place. Additionally, I needed to reuse the formatter to report the XML errors which was previously only created within the factory.

bin/recog_verify
dabdine: Ah yeah. Initially I had it force schema checking by default, but decided to disable that (partly because I didn't want to assume where fingerprints.xsd would be). I'll update the documentation. Good catch.

bin/recog_verify
```diff
@@ -48,15 +55,45 @@ if ARGV.empty?
...
+formatter = Recog::Formatter.new(options, $stdout)
 ARGV.each do |arg|
   Dir.glob(arg).each do |file|
-    ndb = Recog::DB.new(file)
-    verifier = Recog::VerifierFactory.build(options, ndb)
-    verified = verifier.verify
-    failures += verifier.reporter.failure_count
-    warnings += verifier.reporter.warning_count
+    # Create a new reporter per XML file to hold context on success/warn/fails
+    reporter  = Recog::VerifyReporter.new(options, formatter, file)
```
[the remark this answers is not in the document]
dabdine: Will update it. Can we enforce this through some ruby standards checks that get run with CI?

xml/fingerprints.xsd
dabdine: I can place it here for completeness, but there technically is nothing to validate since it'll just be an xsd:string type (it isn't an enumeration, etc.). It passes now because example allows any attribute. _encoding was added because it's an explicit enumeration.

bin/recog_verify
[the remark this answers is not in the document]
dabdine: Question: This also means that we're loading the XML documents twice every time -- once to validate using the schema, and another time to validate the examples.

I briefly looked at this. The way I'd approach solving this problem would be to make a DB factory (this is actually a good place to use one) which has two methods: One that creates a new DB from a file, and one that creates a new DB from a pre-parsed XML document (output of Nokogiri::XML(xxx)). I stopped myself from making that change since it is more invasive, and not sure if we'd want to push that change upstream to recog-go and recog-java.

bin/recog_verify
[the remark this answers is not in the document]
mkienow-r7: Yes, I opened #47 after making that comment!

mkienow-r7 (approved):

PR #426, opened 2022-02-23
Add pfSense and Netgate pfSense Plus firewall fingerprints

mkienow-r7:
## Description
* Adds pfSense and Netgate pfSense Plus firewall fingerprints. Corrects `xml/favicons.xml` pfSense fingerprint since we are unable to assert hardware parameters based on the available data.

## Motivation and Context
More fingerprints!

## How Has This Been Tested?
* `./bin/recog_verify`
* `rake tests`

## Types of changes

## Checklist:

### DIFF
--- a/xml/html_title.xml
+++ b/xml/html_title.xml
@@ -3781,4 +3781,30 @@
     <param pos="1" name="host.name"/>
   </fingerprint>
 
+  <fingerprint pattern="^pfSense - Login$">
+    <description>pfSense Firewall</description>
+    <example>pfSense - Login</example>
+    <param pos="0" name="service.vendor" value="pfSense"/>
+    <param pos="0" name="service.product" value="pfSense"/>
+    <param pos="0" name="service.device" value="Firewall"/>
+    <param pos="0" name="service.component.vendor" value="nginx"/>
+    <param pos="0" name="service.component.product" value="nginx"/>
+    <param pos="0" name="service.component.cpe23" value="cpe:/a:nginx:nginx:-"/>
+    <param pos="0" name="os.vendor" value="pfSense"/>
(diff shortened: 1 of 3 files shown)


xml/html_title.xml
```diff
@@ -3770,4 +3770,24 @@
     <param pos="0" name="service.cpe23" value="cpe:/a:apache:solr:-"/>
   </fingerprint>
 
+  <fingerprint pattern="^pfSense - Login$">
+    <description>pfSense Firewall</description>
+    <example>pfSense - Login</example>
+    <param pos="0" name="service.vendor" value="pfSense"/>
+    <param pos="0" name="service.product" value="pfSense"/>
```
tsellers-r7, line 3777: I believe that the actual product here, per the `Server` header is `nginx`. I'm checking our data to be sure. I should be able to provide an update in an hour or so.

tsellers-r7: Here are the frequency counts values for the `Server` header for Sonar responses where the title is `pfSense - Login`. These were taken from an HTTP study vs `80/tcp` on Jan 23, 2022. I'm not sure about the differences between the top 5 values that made them not sort together. I'll rerun this with `LC_ALL=C` and see what happens. That being said, it provides the info we need.

```
1892 "nginx"
1891 "nginx"
1849 "nginx"
 747 "nginx"
 147 "nginx"
   1 "nginx/1.18.0"
   1 "nginx/1.18.0 (Ubuntu)"
   1 "nginx/1.16.1"
   1 "NADA"
```

mkienow-r7: Note that we agreed to supply this information via `service.component.*` in a separate discussion.

tsellers-r7 (approved):

PR #438, opened 2022-04-05
recog_standardize: handle missing files, detect removed identifiers, unify device type tracking

### LINKED ISSUE #384: recog_standard crashes when identifier files are missing.
**Describe the bug**
`bin/recog_standardize` crashes when trying to process a file that does not exist. 

Sometimes values are added to the identifier files and then they are removed from the fingerprints later without cleaning up the identifier files.  `bin/recog_standardize` doesn't remove orphaned entries. 

 If, instead, it were to create missing files then that would make it easy for contributors to ensure the accuracy of files in `identifiers/` by simply deleting all of the `txt` files and running the tool with the `-w` flag. 

Additionally, it would be nice to have a flag that just does this for us for any file that `recog_standardize` is processing.

**To Reproduce**

```shell
$ rm identifiers/vendor.txt 

$ ruby bin/recog_standardize -w xml/*
Traceback (most recent call last):
	2: from bin/recog_standardize:52:in `<main>'
	1: from bin/recog_standardize:10:in `load_identifiers'
bin/recog_standardize:10:in `readlines': No such file or directory @ rb_sysopen - /Users/something/git/recog/identifiers/vendor.txt (Errno::ENOENT)
```

**Expected behavior**
`recog_standardize` creates any missing files

TomSellers:
## Description
This PR fixes #384 and makes a few other changes.

Changes
1. Adds error handling so that if a particular identifiers file is missing the script will return an empty string so that the contents can be populated from the existing fingerprint database corpus.

1. Adds logic to detect and notify the user when values have been removed from an indicators list. This is useful when performing cleanups or widespread changes of values. No flag is required to enable this behavior, it works just like the previous detection of new values.

1. Unifies `*.device` identifiers in `indicators/device.txt` and removes the existing `os.device` and `hw.device` files. This has already helped identify a few issues which I have addressed.

Example output after deleting 3 valid entries from the `device.txt` and adding one bonus entry of `Sample Device Type Thingy`.

```shell
$ ruby bin/recog_standardize xml/*.xml -w
DEVICE REMOVED VALUE: Sample Device Type Thingy
DEVICE NEW VALUE: DSL Modem
DEVICE NEW VALUE: DSLAM
DEVICE NEW VALUE: DSU/CSU
$ echo $?                                
1

$ ruby bin/recog_standardize xml/*.xml   
$ echo $?                             
0


```

## How Has This Been Tested?
`ruby bin/recog_standardize xml/*.xml` and reviewing the changes

## Types of changes

## Checklist:


mkienow-r7:
What do you think about changes that would be compatible with the `recog_standardize` use in `tools/dev/hooks/pre-commit`? The new reset flag won't provide a focused alert for the user.

xml/snmp_sysdescr.xml
```diff
@@ -5672,11 +5672,11 @@ Copyright (c) 1995-2005 by Cisco Systems
    =======================================================================-->
 
   <fingerprint pattern="^PARADYNE T1 DSU/CSU; model: ([^;]+); S/W Release: ([^;]+);">
-    <description>Paradyne CSU/DSU</description>
+    <description>Paradyne DSU/CSU</description>
```
TomSellers, line 5675: Here and below. Simple typo as can been seen in the example. Changed to align with correct usage elsewhere.

xml/snmp_sysdescr.xml
```diff
@@ -6446,8 +6446,8 @@ Copyright (c) 1995-2005 by Cisco Systems
     <example os.product="GE Series " os.version="V5 R3.10404.0">HiPath Wireless Access Controller - GE Series , System Version V5 R3.10404.0</example>
     <example os.product="V2110" os.version="08.01.01.0251">HiPath Wireless Access Controller - V2110, System Version 08.01.01.0251</example>
     <param pos="0" name="os.vendor" value="Siemens"/>
-    <param pos="0" name="os.device" value="HiPath"/>
-    <param pos="0" name="os.family" value="WAP"/>
+    <param pos="0" name="os.family" value="HiPath"/>
+    <param pos="0" name="os.device" value="WAP"/>
```
TomSellers, line 6450: This was a typo that switched the values between the `family` and `device` params.

TomSellers:
@mkienow-r7 
I went ahead and merged `upstream/main` into this PR to ensure that there were no issues.
Glad to see `master` -> `main`, thanks for doing that.

mkienow-r7 (approved):
Nice work improving `recog_standardize` and cleaning up the identifiers. Thank you for the contribution @TomSellers!

PR #368, opened 2021-07-29
DoS fix / regex performance tuning

tsellers-r7:
## Motivation and Context
  
TL:DR - Fixed an unlikely DoS condition in a pattern in `xml/ssh_banners.xml` and improved performance of certain patterns.  Metrics included here.  This PR doesn't address all of the slower patterns but does get the worst of them.

In a PR ( https://github.com/rapid7/recog-java/pull/7 ) for `recog-java` @hudclark provided some data ( https://github.com/rapid7/recog-java/pull/7#issuecomment-829254286 ) that would cause one of our regexes to hang indefinitely.  That specific problem has already been fixed https://github.com/rapid7/recog/pull/353.

Since `recog` is generally used to process data from untrustworthy sources I wanted to make sure that this specific data and certain other patterns wouldn't cause indefinite hangs or otherwise have unacceptable performance characteristics when processed using any of `recog`'s other fingerprint patterns. In Project Sonar we process 100's of millions of banners on a regular basis so small performance improvements can add up.

Findings:
 - one case of regex that resulted in an indefinite hang when processing large numbers of spaces
 - 50+ cases where performance could be up to **50x** slower than it should be when processing certain strings.

## Testing methodology

Script that processes a specific test string with every fingerprint pattern currently in `recog`. Since the process can be incredibly fast the script timed how long it took to perform this process 100,000 times per pattern. The script then emitted results sorted by duration with the highest duration first. 

There are 3,879 patterns currently in `recog`. Using @hudclark's test string I found that 3,338 took less that 0.1 seconds to process. The longest duration was nearly 9 seconds. Using that as my benchmark I decided that anything over 2 seconds should probably be reviewed. There were 151 of these in this dataset alone.  I did **not** address all of them because later tests revealed patterns that took up to **62.2** seconds and thus took priority.
  
  
  
## First test case - hudclark's 

The data that hudclark provided was string ~1,500 characters long that consisted solely of digits.  It pointed out some patterns that took 2x to 4x longer than I'd like.  These were quickly overshadowed by later tests.

**Original results**
<details>
  <summary>Click to expand results!</summary>
  
```text
8.9967840140    smtp_banners.xml : Some unknown mail server on OpenVMS
7.8108736030    smtp_banners.xml : Postfix - Ubuntu, Mail-in-a-Box package
6.1197626160    smtp_banners.xml : Exim - with version string and optional timestamp
5.8996496320    apache_os.xml    : Sun Cobalt RaQ (Red Hat based Linux)
[...]

PR #476, opened 2022-08-31
More DHCP-based fingerprints

dmoinescu-r7:
## Description
Added more fingerprints from observed DHCP vendor_class strings:
* Huawei Android
* Cisco IP Phones
* Cisco Wireless Phones
* Cisco Aironet Wireless Access Points
* Cisco Catalyst Wireless Access Points
* Cisco Integrated Access Points
* Cisco Wireless Controllers

## Motivation and Context
These fingerprints can be used for identifying network devices based on the DHCP fields they advertise

## How Has This Been Tested?
Ran `./bin/recog_verify xml/dhcp_vendor_class.xml`

## Types of changes

- New fingerprints

## Checklist:

### DIFF
--- a/xml/dhcp_vendor_class.xml
+++ b/xml/dhcp_vendor_class.xml
@@ -218,7 +218,7 @@
     <param pos="0" name="hw.device" value="VoIP"/>
     <param pos="0" name="hw.vendor" value="Cisco"/>
     <param pos="1" name="hw.model"/>
-    <param pos="0" name="hw.certainty" value="0.9"/>
+    <param pos="0" name="hw.certainty" value="0.8"/>
     <param pos="0" name="os.vendor" value="Cisco"/>
   </fingerprint>
 
@@ -228,7 +228,7 @@
     <param pos="0" name="hw.device" value="VoIP"/>
     <param pos="0" name="hw.vendor" value="Cisco"/>
     <param pos="1" name="hw.model"/>
-    <param pos="0" name="hw.certainty" value="0.9"/>
+    <param pos="0" name="hw.certainty" value="0.8"/>
     <param pos="0" name="os.vendor" value="Cisco"/>
   </fingerprint>
 
@@ -242,7 +242,7 @@
     <param pos="0" name="os.vendor" value="Cisco"/>
   </fingerprint>
 
-  <fingerprint pattern="^Cisco AP (c(?:1540|1550|1560|1700|1815|1830|1850|2700|2800|3700|3800|701|702))$" certainty="0.9">
+  <fingerprint pattern="^Cisco AP (c(?:1540|1550|1560|1700|1815|1830|1850|2700|2800|3700|3800|701|702))$" certainty="0.8">
     <description>Cisco Aironet Wireless Access Point</description>
     <example hw.model="c1830">Cisco AP c1830</example>
     <param pos="0" name="hw.device" value="WAP"/>
@@ -251,7 +251,7 @@
     <param pos="0" name="os.vendor" value="Cisco"/>
   </fingerprint>
 
-  <fingerprint pattern="^Cisco AP (C9115AX|C9120AX)" certainty="0.9">
+  <fingerprint pattern="^Cisco AP (C9115AX|C9120AX)" certainty="0.8">
     <description>Cisco Catalyst 9100 Series Wireless Access Point</description>
     <example hw.model="C9115AX">Cisco AP C9115AX</example>
     <param pos="0" name="hw.device" value="WAP"/>
@@ -260,7 +260,7 @@
     <param pos="0" name="os.vendor" value="Cisco"/>
   </fingerprint>
 
-  <fingerprint pattern="^Cisco (AP80[123])$" certainty="0.9">
+  <fingerprint pattern="^Cisco (AP80[123])$" certainty="0.8">
     <description>Cisco 800 Series Integrated Access Point</description>
     <example hw.model="AP801">Cisco AP801</example>
     <example hw.model="AP802">Cisco AP802</example>
@@ -271,7 +271,7 @@
     <param pos="0" name="os.vendor" value="Cisco"/>
   </fingerprint>
 
-  <fingerprint pattern="^Cisco (AIR-CTVM-K9)$" certainty="0.9">
+  <fingerprint pattern="^Cisco (AIR-CTVM-K9)$" certainty="0.8">
     <description>Cisco Virtual Wireless Controller</description>
     <example hw.model="AIR-CTVM-K9">Cisco AIR-CTVM-K9</example>
     <param pos="0" name="hw.device" value="Wireless Controller"/>
@@ -280,7 +280,7 @@
     <param pos="0" name="os.vendor" value="Cisco"/>
   </fingerprint>
 
-  <fingerprint pattern="^Cisco (AIR-CT55\d\d-K9)$" certainty="0.9">
+  <fingerprint pattern="^Cisco (AIR-CT55\d\d-K9)$" certainty="0.8">
     <description>Cisco 5500 Series Wireless Controller</description>
     <example hw.model="AIR-CT5508-K9">Cisco AIR-CT5508-K9</example>
     <param pos="0" name="hw.device" value="Wireless Controller"/>
--- a/xml/dhcp_vendor_class.xml
+++ b/xml/dhcp_vendor_class.xml
@@ -212,20 +212,24 @@
     <param pos="0" name="os.family" value="Windows"/>
   </fingerprint>
 
-  <fingerprint pattern="^(?i)Cisco Systems Inc. Wireless Phone (.*)" certainty="0.9">
+  <fingerprint pattern="^(?i)Cisco Systems Inc. Wireless Phone (\d{4})$">
     <description>Cisco Wireless Phone</description>
     <example hw.model="7925">Cisco Systems Inc. Wireless Phone 7925</example>
     <param pos="0" name="hw.device" value="VoIP"/>
     <param pos="0" name="hw.vendor" value="Cisco"/>
     <param pos="1" name="hw.model"/>
+    <param pos="0" name="hw.certainty" value="0.9"/>
+    <param pos="0" name="os.vendor" value="Cisco"/>
   </fingerprint>
 
-  <fingerprint pattern="^(?i)Cisco Systems, Inc. IP Phone (.*)" certainty="0.9">
+  <fingerprint pattern="^(?i)Cisco Systems, Inc. IP Phone (.*)">
     <description>Cisco IP Phone</description>
     <example hw.model="CP-7911G">Cisco Systems, Inc. IP Phone CP-7911G</example>
     <param pos="0" name="hw.device" value="VoIP"/>
     <param pos="0" name="hw.vendor" value="Cisco"/>
     <param pos="1" name="hw.model"/>
+    <param pos="0" name="hw.certainty" value="0.9"/>
+    <param pos="0" name="os.vendor" value="Cisco"/>
   </fingerprint>
 
   <fingerprint pattern="^Cisco AP (c(?:1130|1140|1240|1250|1260|1600|2600|3500|3600))$" certainty="0.9">
@@ -235,6 +239,7 @@
     <param pos="0" name="hw.device" value="WAP"/>
     <param pos="0" name="hw.vendor" value="Cisco"/>
     <param pos="1" name="hw.model"/>
+    <param pos="0" name="os.vendor" value="Cisco"/>
   </fingerprint>
 
   <fingerprint pattern="^Cisco AP (c(?:1540|1550|1560|1700|1815|1830|1850|2700|2800|3700|3800|701|702))$" certainty="0.9">
@@ -243,6 +248,7 @@
     <param pos="0" name="hw.device" value="WAP"/>
     <param pos="0" name="hw.vendor" value="Cisco"/>
     <param pos="1" name="hw.model"/>
+    <param pos="0" name="os.vendor" value="Cisco"/>
   </fingerprint>
 
   <fingerprint pattern="^Cisco AP (C9115AX|C9120AX)" certainty="0.9">
@@ -251,6 +257,7 @@
     <param pos="0" name="hw.device" value="WAP"/>
     <param pos="0" name="hw.vendor" value="Cisco"/>
     <param pos="1" name="hw.model"/>
+    <param pos="0" name="os.vendor" value="Cisco"/>
   </fingerprint>
 
   <fingerprint pattern="^Cisco (AP80[123])$" certainty="0.9">
@@ -261,6 +268,7 @@
     <param pos="0" name="hw.device" value="WAP"/>
     <param pos="0" name="hw.vendor" value="Cisco"/>
     <param pos="1" name="hw.model"/>
+    <param pos="0" name="os.vendor" value="Cisco"/>
   </fingerprint>
 
   <fingerprint pattern="^Cisco (AIR-CTVM-K9)$" certainty="0.9">
@@ -269,13 +277,15 @@
     <param pos="0" name="hw.device" value="Wireless Controller"/>
     <param pos="0" name="hw.vendor" value="Cisco"/>
     <param pos="1" name="hw.model"/>
+    <param pos="0" name="os.vendor" value="Cisco"/>
   </fingerprint>
 
-  <fingerprint pattern="^Cisco (AIR-CT55..-K9)$" certainty="0.9">
+  <fingerprint pattern="^Cisco (AIR-CT55\d\d-K9)$" certainty="0.9">
     <description>Cisco 5500 Series Wireless Controller</description>
     <example hw.model="AIR-CT5508-K9">Cisco AIR-CT5508-K9</example>
     <param pos="0" name="hw.device" value="Wireless Controller"/>
     <param pos="0" name="hw.vendor" value="Cisco"/>
     <param pos="1" name="hw.model"/>
+    <param pos="0" name="os.vendor" value="Cisco"/>
   </fingerprint>
 </fingerprints>
--- a/xml/dhcp_vendor_class.xml
+++ b/xml/dhcp_vendor_class.xml
@@ -178,6 +178,7 @@
     <param pos="0" name="os.vendor" value="Google"/>
     <param pos="0" name="os.family" value="Linux"/>
     <param pos="0" name="os.product" value="Android"/>
+    <param pos="0" name="os.cpe23" value="cpe:/o:google:android:-"/>
   </fingerprint>
 
   <fingerprint pattern="^dhcpcd-(?:[\d\.]+):Linux-([\d\.]+).*:(\S*):">
--- a/xml/dhcp_vendor_class.xml
+++ b/xml/dhcp_vendor_class.xml
@@ -171,6 +171,15 @@
     <param pos="0" name="os.cpe23" value="cpe:/o:google:android:{os.version}"/>
   </fingerprint>
 
+  <fingerprint pattern="HUAWEI:android:" certainty="0.9">
+    <description>Huawei Android Device</description>
+    <example>HUAWEI:android:ABC</example>
+    <param pos="0" name="hw.vendor" value="Huawei"/>
+    <param pos="0" name="os.vendor" value="Google"/>
+    <param pos="0" name="os.family" value="Linux"/>
+    <param pos="0" name="os.product" value="Android"/>
+  </fingerprint>
+
   <fingerprint pattern="^dhcpcd-(?:[\d\.]+):Linux-([\d\.]+).*:(\S*):">
     <description>Linux</description>
     <example os.version="4.14.78" os.arch="armv7l">dhcpcd-6.11.5:Linux-4.14.78:armv7l:Freescale</example>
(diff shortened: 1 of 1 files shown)


xml/dhcp_vendor_class.xml
```diff
@@ -203,4 +212,70 @@
     <param pos="0" name="os.family" value="Windows"/>
   </fingerprint>
 
-</fingerprints>
\ No newline at end of file
+  <fingerprint pattern="^(?i)Cisco Systems Inc. Wireless Phone (.*)" certainty="0.9">
```
modoyle-r7, line 215: Include $ end of string

xml/dhcp_vendor_class.xml
```diff
@@ -203,4 +212,70 @@
...
+
+  <fingerprint pattern="^(?i)Cisco Systems, Inc. IP Phone (.*)" certainty="0.9">
+    <description>Cisco IP Phone</description>
+    <example hw.model="CP-7911G">Cisco Systems, Inc. IP Phone CP-7911G</example>
+    <param pos="0" name="hw.device" value="VoIP"/>
+    <param pos="0" name="hw.vendor" value="Cisco"/>
+    <param pos="1" name="hw.model"/>
+  </fingerprint>
+
+  <fingerprint pattern="^Cisco AP (c(?:1130|1140|1240|1250|1260|1600|2600|3500|3600))$" certainty="0.9">
```
modoyle-r7, line 231: Do you have examples for all of these? Or expect matches based on research of available products?
I generally only add matches where I've examples.

xml/dhcp_vendor_class.xml
```diff
@@ -203,4 +212,70 @@
...
+
+  <fingerprint pattern="^Cisco (AIR-CTVM-K9)$" certainty="0.9">
+    <description>Cisco Virtual Wireless Controller</description>
+    <example hw.model="AIR-CTVM-K9">Cisco AIR-CTVM-K9</example>
+    <param pos="0" name="hw.device" value="Wireless Controller"/>
+    <param pos="0" name="hw.vendor" value="Cisco"/>
+    <param pos="1" name="hw.model"/>
+  </fingerprint>
+
+  <fingerprint pattern="^Cisco (AIR-CT55..-K9)$" certainty="0.9">
```
modoyle-r7, line 274: Maybe match on digits if you expect digits, else have another example.

xml/dhcp_vendor_class.xml
```diff
@@ -203,4 +212,70 @@
...
\ No newline at end of file
+  <fingerprint pattern="^(?i)Cisco Systems Inc. Wireless Phone (.*)" certainty="0.9">
+    <description>Cisco Wireless Phone</description>
+    <example hw.model="7925">Cisco Systems Inc. Wireless Phone 7925</example>
+    <param pos="0" name="hw.device" value="VoIP"/>
+    <param pos="0" name="hw.vendor" value="Cisco"/>
+    <param pos="1" name="hw.model"/>
+  </fingerprint>
+
+  <fingerprint pattern="^(?i)Cisco Systems, Inc. IP Phone (.*)" certainty="0.9">
```
modoyle-r7, line 223: Should the certainty be hw.certainty as there's nothing here about OS.

xml/dhcp_vendor_class.xml
dmoinescu-r7: Yes, those model numbers reflect strings we actually recorded.

xml/dhcp_vendor_class.xml
dmoinescu-r7: OK, I can do that. My reasoning was that no matter what comes after, it's still a wireless phone, even if the model is 7925 or 7925G or whatever  (saves us from having to add specific models in the future)

xml/dhcp_vendor_class.xml
dmoinescu-r7: I figured that a simpler regex would be preferable to a more complex (read: less readable) one (and `..` looks more readable than `\d\d`, imho). But yeah, sure, I can change it.

xml/dhcp_vendor_class.xml
```diff
@@ -171,6 +171,15 @@
     <param pos="0" name="os.cpe23" value="cpe:/o:google:android:{os.version}"/>
   </fingerprint>
 
+  <fingerprint pattern="HUAWEI:android:" certainty="0.9">
```
mkienow-r7, line 174: Can we add a start of string anchor here? What is the "ABC" value in the example and can it be placed into the regex so that we can also have an end of string anchor?

xml/dhcp_vendor_class.xml
mkienow-r7: While tighter regex sometimes require updates, unbounded matches may result in unexpected matches and captures as well as possible performance issues (see #368).

xml/dhcp_vendor_class.xml
```diff
@@ -203,4 +212,70 @@
...
+
+  <fingerprint pattern="^Cisco AP (c(?:1540|1550|1560|1700|1815|1830|1850|2700|2800|3700|3800|701|702))$" certainty="0.9">
+    <description>Cisco Aironet Wireless Access Point</description>
+    <example hw.model="c1830">Cisco AP c1830</example>
+    <param pos="0" name="hw.device" value="WAP"/>
+    <param pos="0" name="hw.vendor" value="Cisco"/>
+    <param pos="1" name="hw.model"/>
+  </fingerprint>
+
+  <fingerprint pattern="^Cisco AP (C9115AX|C9120AX)" certainty="0.9">
```
mkienow-r7, line 248: Can we add an end of string anchor here?

xml/dhcp_vendor_class.xml
dmoinescu-r7: Actual DHCP strings include what appears to be a random string, sometimes made up of 3 characters, sometimes of more:

- HUAWEI:android:ANE
- HUAWEI:android:FIG-L11
- HUAWEI:android:FIG-L21
- HUAWEI:android:LYA

xml/dhcp_vendor_class.xml
dmoinescu-r7: done

xml/dhcp_vendor_class.xml
```diff
@@ -203,4 +212,80 @@
...
+    <description>Cisco Wireless Phone</description>
+    <example hw.model="7925">Cisco Systems Inc. Wireless Phone 7925</example>
+    <param pos="0" name="hw.device" value="VoIP"/>
+    <param pos="0" name="hw.vendor" value="Cisco"/>
+    <param pos="1" name="hw.model"/>
+    <param pos="0" name="hw.certainty" value="0.8"/>
+    <param pos="0" name="os.vendor" value="Cisco"/>
+  </fingerprint>
+
+  <fingerprint pattern="^(?i)Cisco Systems, Inc. IP Phone (.*)">
```
mkienow-r7, line 225: 1. Escape the period in `Inc\.` otherwise we'll match any character in that position.
2. Tighten the regex rather than allowing unbounded matches that may result in unexpected matches and captures as well as possible performance issues. Is there a reason we couldn't use something like `(CP-\d+G?)` instead of `(.*)`? Note, I derived the pattern from this [existing sip banners fingerprint](https://github.com/rapid7/recog/blob/8383440f954a4511463fd256d62b546ce51075cd/xml/sip_banners.xml#L48).
3. Is there any reason not to have an end of string anchor?

xml/dhcp_vendor_class.xml
mkienow-r7: Do you have any idea what the string represents? Could it be made a part of the pattern and allow us to add an end of string anchor?

xml/dhcp_vendor_class.xml
```diff
@@ -203,4 +212,80 @@
...
+  <fingerprint pattern="^Cisco (AIR-CTVM-K9)$" certainty="0.8">
+    <description>Cisco Virtual Wireless Controller</description>
+    <example hw.model="AIR-CTVM-K9">Cisco AIR-CTVM-K9</example>
+    <param pos="0" name="hw.device" value="Wireless Controller"/>
+    <param pos="0" name="hw.vendor" value="Cisco"/>
+    <param pos="1" name="hw.model"/>
+    <param pos="0" name="os.vendor" value="Cisco"/>
+  </fingerprint>
+
+  <fingerprint pattern="^Cisco (AIR-CT55\d\d-K9)$" certainty="0.8">
```
mkienow-r7, line 283: Please simplify to match exactly two digits using `\d{2}`.

xml/dhcp_vendor_class.xml
```diff
@@ -203,4 +212,80 @@
...
+    <example hw.model="CP-7911G">Cisco Systems, Inc. IP Phone CP-7911G</example>
+    <param pos="0" name="hw.device" value="VoIP"/>
+    <param pos="0" name="hw.vendor" value="Cisco"/>
+    <param pos="1" name="hw.model"/>
+    <param pos="0" name="hw.certainty" value="0.8"/>
+    <param pos="0" name="os.vendor" value="Cisco"/>
+  </fingerprint>
+
+  <fingerprint pattern="^Cisco AP (c(?:1130|1140|1240|1250|1260|1600|2600|3500|3600))$" certainty="0.9">
+    <description>Cisco Aironet Wireless Access Point (obsolete)</description>
```
mkienow-r7, line 236: Should we combine this pattern with the "Cisco Aironet Wireless Access Point" pattern below? It is the same product so I'm not sure them simply being obsolete warrants a separate fingerprint. At some point in time the models in the other fingerprint will also become obsolete.

mkienow-r7 (changes requested):

xml/dhcp_vendor_class.xml
```diff
@@ -203,4 +212,80 @@
     <param pos="0" name="os.family" value="Windows"/>
   </fingerprint>
 
-</fingerprints>
\ No newline at end of file
+  <fingerprint pattern="^(?i)Cisco Systems Inc. Wireless Phone (\d{4})$">
```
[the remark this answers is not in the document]
dmoinescu-r7: done

xml/dhcp_vendor_class.xml
dmoinescu-r7: I added a size limit and an end-of-string anchor. I don't see any benefit in making the model regexp more complicated (something like `CP-?\d{4}(?:G|G-GE)?`), since that would decrease readability and performance at the same time.

xml/dhcp_vendor_class.xml
dmoinescu-r7: done

xml/dhcp_vendor_class.xml
dmoinescu-r7: I believe some customers would find it useful to have obsolete WAPs pointed out since they represent a major security risk, hence the separation. But if the question is rhetorical, I'll combine the two categories.

TomSellers: In my experience, few consumers of Recog are actually using the `description` field in a user facing way.  When at R7 I primarily used it for metrics generation and as commentary describing how this fingerprint differs from other similar fingerprints (for example, `w/o version string`, `with hostname`, etc. 

Additionally, the list of "obsolete" products changes over time. IMO, that logic should be left to the consuming application as opposed to pushing the maintenance burden to Recog.

xml/dhcp_vendor_class.xml
TomSellers: >  My reasoning was that no matter what comes after, it's still a wireless phone, even if the model is 7925 or 7925G 

It's actually a bit more of an issue than that. This may silently match a new variant that is of interest to us in the future. 

For example, `Cisco Systems Inc. Wireless Phone 990x running NewPhoneOS 10.1` would be matched but we'd want to know about this new banner so we can add specific coverage for it and assert `os.*` attributes.

xml/dhcp_vendor_class.xml
```diff
@@ -203,4 +212,83 @@
...
 
-</fingerprints>
\ No newline at end of file
+  <fingerprint pattern="^(?i)Cisco Systems Inc\. Wireless Phone (\d{4})$">
+    <description>Cisco Wireless Phone</description>
+    <example hw.model="7925">Cisco Systems Inc. Wireless Phone 7925</example>
+    <param pos="0" name="hw.device" value="VoIP"/>
+    <param pos="0" name="hw.vendor" value="Cisco"/>
+    <param pos="1" name="hw.model"/>
+    <param pos="0" name="hw.certainty" value="0.8"/>
```
TomSellers, line 221: `hw.product` here should likely be `Unified Wireless IP Phone {hw.model}`

From the NIST CPE list
```
h:cisco:unified_wireless_ip_phone_7920
h:cisco:unified_wireless_ip_phone_7921g
h:cisco:unified_wireless_ip_phone_7925g
```

Similarly, `os.product` here should likely be `Unified Wireless IP Phone {hw.model} Firmware`

```
o:cisco:unified_wireless_ip_phone_7920_firmware
```

xml/dhcp_vendor_class.xml
```diff
@@ -203,4 +212,83 @@
...
+    <param pos="0" name="hw.device" value="WAP"/>
+    <param pos="0" name="hw.vendor" value="Cisco"/>
+    <param pos="1" name="hw.model"/>
+    <param pos="0" name="os.vendor" value="Cisco"/>
+  </fingerprint>
+
+  <fingerprint pattern="^Cisco AP (c(?:1540|1550|1560|1700|1815|1830|1850|2700|2800|3700|3800|701|702))$" certainty="0.8">
+    <description>Cisco Aironet Wireless Access Point</description>
+    <example hw.model="c1830">Cisco AP c1830</example>
+    <param pos="0" name="hw.device" value="WAP"/>
```
TomSellers, line 251: Here and elsewhere for Aironet:

**_IMO_**,  The value for `hw.model` should not contain the `c` prefix. Cisco doesn't use it in [human consumable marketing](https://www.cisco.com/c/en/us/products/collateral/wireless/aironet-1830-series-access-points/datasheet-c78-735582.html)  and it enables use to generate CPEs for the devices.

Example hardware CPES:
```
h:cisco:aironet_1815t
h:cisco:aironet_1815w
h:cisco:aironet_1830
h:cisco:aironet_1830e
h:cisco:aironet_1830i
```

Example OS CPE
```
o:cisco:aironet_1800_firmware
o:cisco:aironet_1800i_firmware
o:cisco:aironet_1810_firmware
o:cisco:aironet_1810w_firmware
```

Assuming the example of `Cisco AP c1830` I recommend `hw.model` be `1830` adding the following attributes:

```
    <param pos="0" name="hw.product" value="Aironet {hw.model}"/>
    <param pos="0" name="os.product" value="Aironet {hw.model} Firmware"/>
```

xml/dhcp_vendor_class.xml
```diff
@@ -203,4 +212,83 @@
...
+    <param pos="0" name="os.vendor" value="Cisco"/>
+  </fingerprint>
+
+  <fingerprint pattern="^Cisco AP (C9115AX|C9120AX)$" certainty="0.8">
+    <description>Cisco Catalyst 9100 Series Wireless Access Point</description>
+    <example hw.model="C9115AX">Cisco AP C9115AX</example>
+    <param pos="0" name="hw.device" value="WAP"/>
+    <param pos="0" name="hw.vendor" value="Cisco"/>
+    <param pos="1" name="hw.model"/>
+    <param pos="0" name="os.vendor" value="Cisco"/>
```
TomSellers, line 263: Ugh, thanks Cisco for tossing the Catalyst branding on something else ..

Similar commentary to the above, here are some CPEs for guidance. 

```
h:cisco:catalyst_9115_ap
h:cisco:catalyst_9117_ap
h:cisco:catalyst_9120_ap
h:cisco:catalyst_9130_ap

o:cisco:catalyst_9115_ap_firmware
o:cisco:catalyst_9117_ap_firmware
o:cisco:catalyst_9120_ap_firmware
o:cisco:catalyst_9130_ap_firmware
```

xml/dhcp_vendor_class.xml
```diff
@@ -203,4 +212,83 @@
...
+    <example hw.model="AP801">Cisco AP801</example>
+    <example hw.model="AP802">Cisco AP802</example>
+    <example hw.model="AP803">Cisco AP803</example>
+    <param pos="0" name="hw.device" value="WAP"/>
+    <param pos="0" name="hw.vendor" value="Cisco"/>
+    <param pos="1" name="hw.model"/>
+    <param pos="0" name="os.vendor" value="Cisco"/>
+  </fingerprint>
+
+  <fingerprint pattern="^Cisco (AIR-CTVM-K9)$" certainty="0.8">
```
TomSellers, line 277: A capture isn't required here, you can just set the value statically.

xml/dhcp_vendor_class.xml
dmoinescu-r7: updated

xml/dhcp_vendor_class.xml
dmoinescu-r7: fixed

xml/dhcp_vendor_class.xml
dmoinescu-r7: updated

xml/dhcp_vendor_class.xml
dmoinescu-r7: updated

xml/dhcp_vendor_class.xml
dmoinescu-r7: Removed the obsolete-Aironet fingerprint (merged into Aironet)

xml/dhcp_vendor_class.xml
dmoinescu-r7: `update_cpes.py` now reports errors for all strings that embed `{hw.model}`:

> xml/dhcp_vendor_class.xml: ERROR: Product unified_wireless_ip_phone_{hw.model} from vendor cisco invalid for CPE h and no mapping.

Is this something something to worry about? Can it be fixed?

TomSellers: It is expected with the current tooling.

xml/dhcp_vendor_class.xml
```diff
@@ -203,4 +213,93 @@
...
+    <example hw.model="3911">Cisco systems, Inc. IP Phone CP3911</example>
+    <param pos="0" name="hw.device" value="VoIP"/>
+    <param pos="0" name="hw.vendor" value="Cisco"/>
+    <param pos="1" name="hw.model"/>
+    <param pos="0" name="hw.product" value="Unified SIP Phone {hw.model}"/>
+    <param pos="0" name="os.vendor" value="Cisco"/>
+    <param pos="0" name="os.product" value="Unified SIP Phone {hw.model} Firmware"/>
+  </fingerprint>
+
+  <fingerprint pattern="^Cisco Systems, Inc\. IP Phone CP-(\d{4}G?)(?:-GE)?$" certainty="0.8">
```
mkienow-r7, line 239: I was taking a look at this fingerprint again after looking over the NVD CPEs and it appears the capture should include the value after the dash as a part of the model number.

For example, the value "Cisco Systems, Inc. IP Phone CP-7941G-GE" would result in the CPE `h:cisco:unified_ip_phone_7941g`, however, the model is "7941G-GE" and NVD contains the CPE `h:cisco:unified_ip_phone_7941g-ge`.

dmoinescu-r7: As far as I can tell, the CPE contains the following entries for 7941:
```
h:cisco:unified_ip_phone_7941g
h:cisco:unified_ip_phone_7941g-ge

o:cisco:unified_ip_phone_firmware_7941g
o:cisco:unified_ip_phone_7941g_firmware
```

So both the G and GE models are present, but firmware only for G. I've no idea what the correct approach is (for instance, can anything bad happen if we refer to an inexistent firmware model?), so I'm totally happy to take guidance.

mkienow-r7: @dmoinescu-r7 I understand the confusion. The CPEs from NVD are frequently added based on existing CVEs and sometimes reflect different naming decisions based on the individual that added them to the NVD database. In addition, understanding the reason for differences might require additional research into particular model(s) when possible. It appears we've identified an odd edge case with the 7941g model hardware and firmware. I would error on the side of emitting values that match the known model number. When we know we have the `7941g-ge` model we should emit values that match known CPEs with that model number. If we were to learn that the `7941g-ge` is simply a hardware revision and uses the `7941g` firmware we might carve off the `-ge` models into their own fingerprint allowing us more flexibility over formatting the params for those models. In general, one uses NVD data to guide formatting the various params. I would advise selecting the format that matches the greatest number in the NVD database. @TomSellers do you have anything you would like to add here?

Looking closer at the full list of `cisco:unified_ip_phone_` CPE it appears the fingerprint pattern could be extended to match two models that include `nr` after the numeric model.

```
h:cisco:unified_ip_phone_8851nr
h:cisco:unified_ip_phone_8865nr

o:cisco:unified_ip_phone_8851nr_firmware
o:cisco:unified_ip_phone_8865nr_firmware
```

From the perspective of new fingerprints there would be no failures caused by a reference to a nonexistent firmware model. It is simply data enrichment, but it is best to figure this out when the fingerprint is first added. As you discovered earlier the `update_cpes.py` script doesn't currently support strings that embed `{hw.model}`, therefore, it won't emit the `*.cpe23` params. Formatting the product params correctly now, will allow them to be emitted in the future when the tool is enhanced.

We should schedule a call this week to discuss the joys of CPE in greater detail. :)

xml/dhcp_vendor_class.xml
```diff
@@ -203,4 +213,94 @@
...
+    <example hw.model="3911">Cisco systems, Inc. IP Phone CP3911</example>
+    <param pos="0" name="hw.device" value="VoIP"/>
+    <param pos="0" name="hw.vendor" value="Cisco"/>
+    <param pos="1" name="hw.model"/>
+    <param pos="0" name="hw.product" value="Unified SIP Phone {hw.model}"/>
+    <param pos="0" name="os.vendor" value="Cisco"/>
+    <param pos="0" name="os.product" value="Unified SIP Phone {hw.model} Firmware"/>
+  </fingerprint>
+
+  <fingerprint pattern="^Cisco Systems, Inc\. IP Phone CP-(\d{4}(?:G|G-GE)?)$" certainty="0.8">
```
mkienow-r7, line 239: I mentioned this on our call today, but wanted to document in the PR. We should consider matching the models that include `nr` after the numeric model number (there are 4 `cisco:unified_ip_phone_` CPEs that include the model number).

xml/dhcp_vendor_class.xml
dmoinescu-r7: updated

xml/dhcp_vendor_class.xml
dmoinescu-r7: updated to allow for trailing NR

mkienow-r7 (approved):
Thank you for the contribution @dmoinescu-r7!

PR #479, opened 2022-09-14
Add Python WSGIServer fingerprint

mkienow-r7:
## Description
Adds fingerprint for Python's WSGIServer ([wsgiref.simple_server](https://docs.python.org/3/library/wsgiref.html#module-wsgiref.simple_server)), a simple WSGI HTTP server.

## Motivation and Context
Improved coverage

## How Has This Been Tested?
* `bundle exec ./bin/recog_verify xml/http_servers.xml`
* `rake tests`

## Types of changes

## Checklist:

### DIFF
--- a/xml/http_servers.xml
+++ b/xml/http_servers.xml
@@ -2623,6 +2623,17 @@
     <param pos="0" name="service.cpe23" value="cpe:/a:encode:uvicorn:-"/>
   </fingerprint>
 
+  <fingerprint pattern="^WSGIServer/(\d+(?:\.\d+)*)\s*C?Python/(\d+(?:\.\d+)*)$">
+    <description>WSGIServer (wsgiref.simple_server) – a simple WSGI HTTP server for Python</description>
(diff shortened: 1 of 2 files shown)


xml/http_servers.xml
```diff
@@ -2623,6 +2623,17 @@
     <param pos="0" name="service.cpe23" value="cpe:/a:encode:uvicorn:-"/>
   </fingerprint>
 
+  <fingerprint pattern="^WSGIServer/((?:\d+\.)*\d+)\s*C?Python/((?:\d+\.)*\d+)$">
```
cbarnard-r7, line 2626: Matches `WSGIServer/0.2                          CPython/3.10.5` but I think that's fine.

cbarnard-r7 (approved):
LGTM

xml/http_servers.xml
mkienow-r7: It should match `WSGIServer/0.2 CPython/3.10.5` as that's one of the examples. :)

cbarnard-r7: Oops, I copied the wrong example. I think it was a similar multiple dot situation.

PR #491, opened 2022-09-29
DHCP vendor_class Xerox Printer updates

modoyle-r7:
DHCP vendor_class Xerox Printers

Update fingerprint based on new examples, mainly case insensitive matches.

Update for Typ=Printer (uppercase P)
Case insensitive for Mfg, Typ, Mod, Ser and Loc
Support 10 digit serial numbers with example
Added Typ=AIO


mkienow-r7 (approved):
Thank you for the contribution @modoyle-r7!

PR #501, opened 2022-10-27
New matches for Apple Boot Service Discovery Protocol

modoyle-r7:
Device descriptions from:
https://support.apple.com/en-gb/HT201634
https://support.apple.com/en-gb/HT201862
https://support.apple.com/en-ie/HT201300
https://support.apple.com/en-ie/HT201581

Added hw_family MacBook Air
Added various Mac to hw_products


mkienow-r7:
It appears that not all `hw.product` values included have a remap entry in `cpe-remap.yaml` (See [current Apple hardware product remap entries](https://github.com/rapid7/recog/blob/e952c3cdf18ff1f194f800ea0d7756cb19545c66/cpe-remap.yaml#L334-L351)). By adding the remap entries `update_cpes.py` should be able to emit a `hw.cpe23` value.

mkienow-r7 (approved):

PR #508, opened 2022-11-07
DHCP fingerprint for Compaq Remote Insight

dmoinescu-r7:
## Description
Fingerprint for Compaq Remote Insight based on observed DHCP traffic

## How Has This Been Tested?
Ran `recog_verify`, `recog_standardize` and `update_cpes`

## Types of changes

## Checklist:

### DIFF
--- a/cpe-remap.yaml
+++ b/cpe-remap.yaml
@@ -376,6 +376,9 @@ mappings:
     hp:
       products:
         ilo: integrated_lights-out
+        ilo_2: integrated_lights-out_2
+        ilo_3: integrated_lights-out_3
+        ilo_4: integrated_lights-out_4
     kace:
       vendor: dell
       products:
--- a/xml/dhcp_vendor_class.xml
+++ b/xml/dhcp_vendor_class.xml
@@ -412,6 +412,24 @@
     <param pos="0" name="hw.product" value="iDRAC"/>
   </fingerprint>
 
+  <fingerprint pattern="^CPQRIB3$">
+    <!-- CPQRI stands for Compaq Remote Insight -->
+
+    <description>Compaq/HP Remote Insight/Integrated Lights-Out</description>
+    <example>CPQRIB3</example>
+    <param pos="0" name="hw.device" value="Lights Out Management"/>
+    <param pos="0" name="hw.vendor" value="HP"/>
+    <param pos="0" name="hw.product" value="iLO 3"/>
+    <param pos="0" name="hw.certainty" value="0.5"/>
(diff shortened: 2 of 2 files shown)


cpe-remap.yaml
```diff
@@ -376,6 +376,10 @@ mappings:
     hp:
       products:
         ilo: integrated_lights-out
+        ilo_2: integrated_lights-out_2
+        ilo_3: integrated_lights-out_3
+        ilo_4: integrated_lights-out_4
+        ilo_5: integrated_lights-out_5
```
mkienow-r7, line 382: Was this block created in a sense copying and modifying what exists in the operating system remap?

The following HP hardware CPEs exist in the NVD dictionary, but there isn't an entry for `h:hp:integrated_lights-out_5`.
```
h:hp:integrated_lights-out
h:hp:integrated_lights-out_2
h:hp:integrated_lights-out_3
h:hp:integrated_lights-out_4
```

In addition, there are only fingerprints with `hw.product` values for "iLO 3" and "iLO 4" currently in use. I think one could argue for keeping the hardware `ilo_2` remap since the operating system remap has a value for `ilo_5` and I didn't see that in use. The remapping is really used when we have an active need to remap a value rather than building out all possible remappings.

dmoinescu-r7: > Was this block created in a sense copying and modifying what exists in the operating system remap?

Yes. Prior to adding those remaps, `update_cpes.py` complained that `Product ilo_3 from vendor hp invalid for CPE h and no mapping`, which was very puzzling (why not an equivalent error for `o`?). After a bit of head-scratching, I discovered the remaps for `o:hp:ilo_*` and I replicated them into `h:hp:`.

On the `ilo_5` question, I was pretty sure that I'd seen an `ilo_5` hardware entry, but on closer inspection, turns out that what I'd actually seen was this:
```
<cpe-item name="cpe:/o:hp:integrated_lights-out_5:2.44">
```
which is the only `/o` entry that doesn't include `_firmware`, so I mistook it for a hardware entry (I suspect it was meant to be be a hardware one and the `/o` bit is a copy-paste error, but whatever).

I'll remove the remapping for `ilo_5`. If you feel we should also remove the `ilo_2` remapping on account of its not being used, fine with me.

dmoinescu-r7: >I'll remove the remapping for `ilo_5`
Done.

mkienow-r7 (approved):
Thank you for the contribution @dmoinescu-r7!

PR #511, opened 2022-11-11
New DHCP fingerprints for Cisco VoIP Devices

modoyle-r7:
CISCO SPA112
CISCO SPA122
CISCO SPA232D
CISCO SPA8800

Added individual fingerprints for Cisco SPA VoIP gateway devices, from captured vendor_class strings.


xml/dhcp_vendor_class.xml
```diff
@@ -367,6 +367,52 @@
     <param pos="0" name="os.vendor" value="Cisco"/>
   </fingerprint>
 
+  <fingerprint pattern="^CISCO SPA8800$">
+    <description>Cisco SPA8800 VoIP Gateway</description>
+    <example>CISCO SPA8800</example>
+    <param pos="0" name="hw.vendor" value="Cisco"/>
+    <param pos="0" name="hw.device" value="VoIP"/>
+    <param pos="0" name="hw.product" value="SPA8800 IP Telephony Gateway"/>
+    <param pos="0" name="hw.cpe23" value="cpe:/h:cisco:spa8800_ip_telephony_gateway:-"/>
```
mkienow-r7, line 376: It's interesting that the `/h` and the `/o` don't match for the spa8800, but they do for the spa8000. 
```
h:cisco:spa8000_8-port_ip_telephony_gateway
h:cisco:spa8800_ip_telephony_gateway
```
```
o:cisco:spa8000_8-port_ip_telephony_gateway_firmware
o:cisco:spa8800_8-port_ip_telephony_gateway_firmware
```

mkienow-r7 (approved):
Thank you for the contribution @modoyle-r7!

xml/dhcp_vendor_class.xml
```diff
@@ -367,6 +367,52 @@
...
+    <param pos="0" name="os.vendor" value="Cisco"/>
+    <param pos="0" name="os.product" value="SPA8800 8-Port IP Telephony Gateway Firmware"/>
+    <param pos="0" name="os.cpe23" value="cpe:/o:cisco:spa8800_8-port_ip_telephony_gateway_firmware:-"/>
+  </fingerprint>
+
+  <fingerprint pattern="^CISCO SPA112$">
+    <description>Cisco SPA112</description>
```
dmoinescu-r7, line 383: Shouldn't this be a bit more enlightening about what the device actually is? (IIUC, ATA stands for Analog Telephone Adapter)

dmoinescu-r7 (changes requested):

xml/dhcp_vendor_class.xml
mkienow-r7: I took a quick look an existing SIP fingerprint for the SPA112 and it didn't have an expansion of SPA, however, it can be helpful to provide additional details in the description. These details could be added in a follow-up PR. From some quick searching it appears SPA might be a "shared port adapter". Although, in this case the name of the [SPA112](https://www.cisco.com/c/en/us/products/unified-communications/spa112-2-port-phone-adapter/index.html) appears to be "Cisco SPA112 2-Port Phone Adapter" and makes no reference to "shared port adapter" and only "port adapter".

modoyle-r7: Sorry Dan, didn't see this till after the PR was merged.
Yes, as Matthew identified, there is an existing fingerprint for the SPA112 device in sip_banners, so I used the same description. 2 fingerprints for a single device with different descriptions looked confusing to me.

PR #517, opened 2022-12-09
Sortlines on all identifiers/*.txt

todb-r7:
## Description

Presumably, the intent is to keep these alphabetized so new people can pop in and add new entries in a sane way. Looks like the alphabetization has gotten a little loose.

If there's a performance reason to keep these strategically semi-sorted, then please reject this change.

## Motivation and Context
Make it easier for people to sense where to insert their new entries.

The only change here is alphabetization. I didn't do anything like add/remove newlines, spellcheck, or sneak in new entries.

## How Has This Been Tested?

In absolutely no way has this been tested beyond my human eyeballs.

## Types of changes

## Checklist:

### DIFF
--- a/identifiers/device.txt
+++ b/identifiers/device.txt
@@ -1,41 +1,41 @@
+Access Control
 ADSL Modem
 ADSL Router
-ATM DSL Unit
-AV Receiver
-Access Control
 Alarm Panel
 Appliance
+ATM DSL Unit
 Audio Encoder
+AV Receiver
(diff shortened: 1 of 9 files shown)


mkienow-r7:
The `identifiers/*` files should not be manually edited. These are managed by `./bin/recog_standardize`.

todb-r7:
This is a bad idea, forget it. See #438, there's a sorting in place already.

PR #537, opened 2023-02-02
Add Exchange 2007 POP3 Quit Message banner pattern

cfleming-R7:
## Description
Add a banner pattern for a quit message for Microsoft Exchange Server 2007

## Motivation and Context
We observed a POP3 quit message exposing the underlying service, in this case Microsoft Exchange Server 2007. A pattern has been added to match against that banner

## How Has This Been Tested?
- rake tests
- bundle exec ./bin/recog_verify xml/pop_banners.xml

## Types of changes

## Checklist:

### DIFF
--- a/xml/pop_banners.xml
+++ b/xml/pop_banners.xml
@@ -152,6 +152,19 @@
     <param pos="0" name="os.cpe23" value="cpe:/o:microsoft:windows_server_2003:-"/>
   </fingerprint>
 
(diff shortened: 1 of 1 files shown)


mkienow-r7:
I don't believe `xml/pop_banners.xml` is the correct place for quit messages since that is not really the banner. For "SMTP response lines to the QUIT command" we have `xml/smtp_quit.xml` so this would belong in a new database `xml/pop_quit.xml`.

PR #547, opened 2023-02-23
mirror doc changes from recog-ruby

TomSellers:
## Description
This PR mirrors README changes from https://github.com/rapid7/recog-ruby/pull/12
Please see that PR for discussion.

it also prevents the `_tmp.` attributes from being added to the `identifiers/fields.txt` file.

NOTE: Tests involving `features/data/successful_tests.xml` and the `recog-ruby` implementation SHOULD fail until that PR is landed.

PR #552, opened 2023-03-02
Cross language compatibility: case insensitivity flag

TomSellers:
## Description
This PR moves the inline flag `(?i)` (case insensitivity)  to the beginning of the regex. In Python this flags has global meaning and use of it anywhere other than the start of the pattern is deprecated. This does not change the function of this flag in current patterns for other languages.

See previous work in https://github.com/rapid7/recog/pull/367

This was caught by internal testing in `recog-go` and the fixes are required in order for us to implement and test the changes from the following PRs into `recog-go`:
- https://github.com/rapid7/recog-ruby/pull/12
- https://github.com/rapid7/recog/pull/547

## How Has This Been Tested?
`bundle exec rake tests`
`ruby bin/recog_verify xml/*.xml | grep -v WARN | grep -v SUMMARY`

## Types of changes>
- Content

## Checklist:

### DIFF
--- a/xml/dhcp_vendor_class.xml
+++ b/xml/dhcp_vendor_class.xml
@@ -8,7 +8,7 @@
     and inform (8) messages.
   -->
 
-  <fingerprint pattern="^[Mm]fg=(?:Fuji)?(?i:Xerox);[Tt]yp=(?:MFP|AIO|[Pp]rinter);[Mm]od=(?:Xerox )?(\S+) ([a-zA-Z0-9]+).*;[Ss]er=([A-Z0-9]{9,10})(?:;[Ll]oc=.*)?$">
+  <fingerprint pattern="(?i)^[Mm]fg=(?:Fuji)?Xerox;[Tt]yp=(?:MFP|AIO|[Pp]rinter);[Mm]od=(?:Xerox )?(\S+) ([a-zA-Z0-9]+).*;[Ss]er=([A-Z0-9]{9,10})(?:;[Ll]oc=.*)?$">
     <description>Xerox Multifunction Printer</description>
     <example hw.family="VersaLink" hw.model="C405" hw.serial_number="ABC123456">Mfg=Xerox;Typ=MFP;Mod=VersaLink C405;Ser=ABC123456;Loc=Print Room</example>
     <example hw.family="AltaLink" hw.model="C8055" hw.serial_number="1AB234567">Mfg=Xerox;Typ=MFP;Mod=Xerox AltaLink C8055 Multifunction Printer;Ser=1AB234567;Loc=Print Room2</example>
--- a/xml/dhcp_vendor_class.xml
+++ b/xml/dhcp_vendor_class.xml
@@ -8,7 +8,7 @@
     and inform (8) messages.
   -->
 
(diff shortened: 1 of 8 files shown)


xml/imap_banners.xml
```diff
@@ -142,7 +142,6 @@
     <example>Dovecot (Debian) ready.</example>
     <example>[CAPABILITY IMAP4rev1 SASL-IR LOGIN-REFERRALS ID ENABLE IDLE LITERAL+ AUTH=PLAIN AUTH=LOGIN] Dovecot (Debian) ready.</example>
     <param pos="0" name="service.vendor" value="Dovecot"/>
-    <param pos="0" name="service.vendor" value="Dovecot"/>
```
TomSellers, line 145: Not flag related, but duplicate `param` use detected by tests in `recog-go`.

mkienow-r7:
> This was caught by internal testing in recog-go ...

The hope behind the [Go Verify workflow](https://github.com/rapid7/recog/blob/c25eda8b5ec4e8b04311f41b66760222c43a27a8/.github/workflows/verify.yml#L70-L89) was to detect issues between the various recog language implementation. What sort of internal testing discovered this issue that was not discoverable when verifying the fingerprints against their examples via [cmd/recog_verify/main.go](https://github.com/runZeroInc/recog-go/blob/main/cmd/recog_verify/main.go)?

xml/dhcp_vendor_class.xml
```diff
@@ -8,7 +8,7 @@
     and inform (8) messages.
   -->
 
-  <fingerprint pattern="^[Mm]fg=(?:Fuji)?(?i:Xerox);[Tt]yp=(?:MFP|AIO|[Pp]rinter);[Mm]od=(?:Xerox )?(\S+) ([a-zA-Z0-9]+).*;[Ss]er=([A-Z0-9]{9,10})(?:;[Ll]oc=.*)?$">
+  <fingerprint pattern="(?i)^[Mm]fg=(?:Fuji)?Xerox;[Tt]yp=(?:MFP|AIO|[Pp]rinter);[Mm]od=(?:Xerox )?(\S+) ([a-zA-Z0-9]+).*;[Ss]er=([A-Z0-9]{9,10})(?:;[Ll]oc=.*)?$">
```
mkienow-r7, line 11: With case insensitivity applied to the whole pattern the explicit case matching is no longer necessary.
* `[Mm]`
* `[Tt]`
* `[Pp]`
* `[Ss]`
* `[Ll]`

TomSellers: Good catch, fixed.

TomSellers:
@mkienow-r7 

> The hope behind the [Go Verify workflow](https://github.com/rapid7/recog/blob/c25eda8b5ec4e8b04311f41b66760222c43a27a8/.github/workflows/verify.yml#L70-L89) was to detect issues between the various recog language implementation. What sort of internal testing discovered this issue that was not discoverable when verifying the fingerprints against their examples via [cmd/recog_verify/main.go](https://github.com/runZeroInc/recog-go/blob/main/cmd/recog_verify/main.go)?

Testing in the Go version of `recog_verify` should be the same as the ruby version so that we can test consistently.

The test for this condition is in `fingerprints_test.go` which would similar to the `spec` tests.

https://github.com/runZeroInc/recog-go/blob/1f868cde9c4b2e110f99d1114d1df4e530842f90/fingerprints_test.go#L51-L58

mkienow-r7 (approved):
Thank you for the contribution @TomSellers!

PR #559, opened 2023-03-22
Rename SSL-VPN *.device parameter values to be more generic

mkienow-r7:
## Description
Renames "SSL-VPN" `*.device` parameter values to the more generic value "VPN".

## Motivation and Context
Provide concise device classification

## How Has This Been Tested?
* `rake tests`
* `bundle exec ./bin/recog_verify xml/dhcp_vendor_class.xml xml/html_title.xml xml/http_cookies.xml`

## Types of changes

## Checklist:

### DIFF
--- a/identifiers/device.txt
+++ b/identifiers/device.txt
@@ -78,7 +78,6 @@ Router
 SD-WAN Appliance
 SIP Device
 SIP Gateway
-SSL-VPN
 Scanner
 Security Appliance
 Sensor
--- a/xml/dhcp_vendor_class.xml
+++ b/xml/dhcp_vendor_class.xml
@@ -947,7 +947,7 @@
   <fingerprint pattern="^JNPR\.IVE$">
     <description>Juniper SSL VPN</description>
     <example>JNPR.IVE</example>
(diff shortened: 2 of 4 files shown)


adfoster-r7 (approved):
The change looks good to me in isolation :+1:

---

From a quick git blame it looks like the original `SSL-VPN` entry was added by HDM a few years ago - so it seems like an intentional decision to differentiate these values https://github.com/rapid7/recog/blame/bbb61f11ddc4ca246498c26baacf0a2e58bdf8b2/xml/http_cookies.xml#L248-L249

But it looks like this distinction hasn't been made for other entries such as the ANsession fingerprint

```
  <fingerprint pattern="^ANsession\d+=(\S+);">
    <description>Array Networks Secure Access Gateway / **SSL VPN**</description>
    <example cookie="IPMI">ANsession0002262072457555=IPMI; path=/;secure</example>
    <param pos="1" name="cookie"/>
    <param pos="0" name="service.vendor" value="Array Networks"/>
    <param pos="0" name="service.family" value="Secure Access Gateway"/>
  </fingerprint>
```

PR #563, opened 2023-04-14
Add FatPipe Networks MPVPN, IPVPN, WARP, SDWAN fingerprints

mkienow-r7:
## Description
Adds 5 fingerprints for FatPipe Networks MPVPN, IPVPN, WARP and SDWAN products.

### Notes
Fingerprinted various product versions (7.1.2r186p12, 9.1.2r156, 9.1.2r161p26, etc.)

* `/fpui/img/favicon.ico` (login)
```
$ file favicon.ico
favicon.ico: MS Windows icon resource - 1 icon, 16x16, 16 colors, 4 bits/pixel
$ md5 favicon.ico
MD5 (favicon.ico) = eddb2b697c8c3dab5a9572de564976d1
```
* `favicon.ico` (remote configuration)
```
$ file favicon.ico
favicon.ico: MS Windows icon resource - 1 icon, 16x16
$ md5 favicon.ico
MD5 (favicon.ico) = b0f000f8041a6f1d2c366972b1dbad69
```

## Motivation and Context
Improved coverage

## How Has This Been Tested?
* `bundle exec ./bin/recog_verify xml/favicons.xml xml/html_title.xml`
* `rake tests`

## Types of changes

## Checklist:

### DIFF
--- a/xml/favicons.xml
+++ b/xml/favicons.xml
@@ -2346,4 +2346,25 @@
     <param pos="0" name="service.product" value="R1Soft Server Backup Manager"/>
   </fingerprint>
 
+  <fingerprint pattern="^(?:eddb2b697c8c3dab5a9572de564976d1|b0f000f8041a6f1d2c366972b1dbad69)$">
+    <description>FatPipe Networks device web interface</description>
+    <!--
+      favicon.ico from multiple products and builds:
+        * FatPipe MPVPN versions: 9.1.2r161p26, 9.1.2r164p5
+        * FatPipe IPVPN versions: 9.1.2r129, 10.1.2r60p89
+        * FatPipe WARP versions: 9.1.2r161p19, 9.1.2r161p26, 9.1.2r185
+        * FatPipe SDWAN versions: unknown
+    -->
+    <example>eddb2b697c8c3dab5a9572de564976d1</example>
+    <!--
+      favicon.ico from multiple products and builds:
+        * FatPipe MPVPN version 9.1.2r161p26 to 9.1.2r164p5
+        * FatPipe IPVPN versions: 9.1.2r129
+        * FatPipe WARP versions: 9.1.2r161p19, 9.1.2r161p26, 9.1.2r185
+    -->
+    <example>b0f000f8041a6f1d2c366972b1dbad69</example>
+    <param pos="0" name="os.vendor" value="FatPipe Networks"/>
+    <param pos="0" name="hw.vendor" value="FatPipe Networks"/>
+  </fingerprint>
+
 </fingerprints>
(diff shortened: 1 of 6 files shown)


xml/html_title.xml
```diff
@@ -4196,4 +4196,59 @@
     <param pos="0" name="hw.device" value="Broadband Router"/>
   </fingerprint>
 
+  <!-- FatPipe Networks fingerprints -->
+
+  <fingerprint pattern="^FatPipe MPVPN(?:.nbsp;\| Log in| Remote Configuration)$">
```
adfoster-r7, line 4201: In terms of future-proofing a regex like this, i.e. if they update their page to have a different suffix - such as `FatPipe MPVPN - VPN Security | Login` or similar; would a partial match suffice?

```suggestion
  <fingerprint pattern="^FatPipe MPVPN.*$">
```

Or is that considered an anti-pattern?

mkienow-r7: I'm a fan of future flexibility, but the mindset handed off when I joined the project is that the fingerprint should be as specific as possible and only add flexibility where you know it needs to exist, for example, if there were a hostname in the title, etc.

Perform a quick google `intitle:"FatPipe MPVPN"` and you'll see examples that would result in a false positive match with the `.*` change.
```
FatPipe MPVPN - Product Overview
FatPipe MPVPN 2U 10Gbps Application Delivery Controller
FatPipe MPVPN Reviews & Product Details
```

xml/favicons.xml
```diff
@@ -2346,4 +2346,12 @@
     <param pos="0" name="service.product" value="R1Soft Server Backup Manager"/>
   </fingerprint>
 
+  <fingerprint pattern="^(?:eddb2b697c8c3dab5a9572de564976d1|b0f000f8041a6f1d2c366972b1dbad69)$">
+    <description>FatPipe Networks device web interface</description>
+    <example>eddb2b697c8c3dab5a9572de564976d1</example>
```
adfoster-r7, line 2351: Not a blocker - I wish we could add additional metadata here for what versions these are from, i.e. from your PR description you call out version `7.1.2r186p12` etc

that's something I would've found useful when I was adding new fingerprints for drupal - as I wasn't sure what versions were already fingerprinted when I was adding the fingerprints for newer versions

adfoster-r7 (approved):

xml/favicons.xml
mkienow-r7: That's a valid point. I've been in the habit of putting those details in the PR since there wasn't much in the way of documenting those as comments in the fingerprints. If we add that detail as comments to favicon fingerprints should we be adding them to all of the other fingerprints?

I'd like to consider even a future revision of the XML schema that includes this data.

PR #596, opened 2023-11-20
LEDO-125: Adding Jenkinsfile to support EKS job runs

dtwaddle2-r7:
## Description
Adding a Jenkinsfile to describe the job run by Jenkins in EKS. Tries to closely mimic the previous job, including some old debugging lines. Will need to determine if this entire process is still required.

Initially attempted to reuse an existing pipeline, however decided that due to deviations and additional behaviours required - decided to just create a custom pipeline here.

https://issues.corp.rapid7.com/browse/LEDO-125

## Motivation and Context
Transferring the release job over to Jenkins EKS.

## How Has This Been Tested?
Ran the new job multiple time, including the release process.

## Types of changes


Jenkinsfile
```diff
@@ -0,0 +1,142 @@
...
+                                curl --silent --show-error \
+                                    -H "Content-Type: application/json" \
+                                    -H "Accept: application/vnd.github.v3+json" \
+                                    -H "Authorization: token \${GH_TOKEN}" \
+                                    -d '{"tag_name":"'"v${VERSION}"'", "generate_release_notes": true, "name": "'"v${VERSION} - \$(date +"%Y.%m.%d")"'"}' \
+                                    https://api.github.com/repos/rapid7/recog/releases > recog-content-releases-response.json
+                            """
+                        }
+                    }
+                }
```
dtwaddle2-r7, line 77: Curious whether there are any thoughts on this and the following stage `process release response`.
I think the old job used this setup to verify the release process was functioning as expected. Initialising the release and then verifying the response, before sending the actual content.

I'm not too familiar with Github releases - particularly through this endpoint. Though curious if there are any opinions on whether this may be necessary? Or perhaps it could be simplified down into a single request.

Jenkinsfile
```diff
@@ -0,0 +1,142 @@
...
+                            RELEASE_HTML_URL=sh(script: 'cat recog-content-releases-response.json | jq -r .html_url', returnStdout: true).trim()
+                            echo "[*] Release v${VERSION} available at: ${RELEASE_HTML_URL}"
+
+                            // Note: This returns an `upload_url` key corresponding to the endpoint for uploading release assets.
+                            // This key is a [hypermedia resource](https://docs.github.com/rest/overview/resources-in-the-rest-api#hypermedia).
+                            // All URLs are expected to be proper RFC 6570 URI templates.
+                            UPLOAD_URL=sh(script: 'cat recog-content-releases-response.json | jq -r .upload_url | cut -f 1 -d "{"', returnStdout: true).trim()
+                        }
+                    }
+                }
```
dtwaddle2-r7, line 110: Tried to keep this consistent with the original job - including most of the debug lines. I was considering simplifying this - particularly it shouldn't be necessary if we combine the release requests. Though can definitely keep them if it may be helpful.

Jenkinsfile
inokii: @dtwaddle2-r7 That is correct. The `Process release response` stage is to verify there were not errors in the previous `Initialise Release` stage.

See the comment on 82-83:
```
// Note: Inspect the response for error message since curl exits with code 0
// for the "Validation failed" response with status code 422 Unprocessable Entity
```

inokii: Please note that it is also performing more work than verifying the release. It processes that response to get values that are used in other stages such as `UPLOAD_URL`.

dtwaddle2-r7: Ah, thank you for clarifying. I initially thought it may be possible to combine this initial release with the one used to submit the zip content. That perhaps this approach was done for additional validation / debugging.
Though yeah if it has to be separated, then this makes sense. Thanks.

inokii: You might be able to combine the work with other stages. Thinking back I may have broken it down to better determine the cause of a failure. If curl happened to exit with a non-zero then the `Initialise Release` step would fail and you would know that was due to a network or auth issue. If the job failed in `Process release response` it would based on the API response. See: [HTTP response status codes for "Create a release"](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release--status-codes).

PR #624, opened 2024-08-28
VC-2125 Add sip-banner record for Telepresence with higher certainty

atorovik-r7:
## Description
A detailed description of your changes.
Updated sip-banner record for Cisco Telepresence to reflect the current product naming, vendor, family. Make regex more strict to match only versions starting with X and TC. 
SIP protocol example:

SIP/2.0 403 Forbidden
Via: SIP/2.0/UDP nm;branch=foo;received=224.85.86.107;rport=26810;ingress-zone=DefaultZone
Call-ID: 50000
CSeq: 42 OPTIONS
From: <sip:nm@nm>;tag=root
To: <sip:nm2@nm2>;tag=dba04991a2a1090f
Server: TANDBERG/4145 (X15.0.3)

## How Has This Been Tested?
Tested with bin/recog_verify

## Types of changes

## Checklist:

### DIFF
--- a/xml/sip_banners.xml
+++ b/xml/sip_banners.xml
@@ -197,6 +197,24 @@
     <param pos="1" name="hw.product"/>
   </fingerprint>
 
+
+  <fingerprint pattern="^(TANDBERG\/(\d+)) \(((X|TC)\d+\S+).*\)$">
+    <description>Cisco/Tandberg TelePresence</description>
+    <example os.version="TC7.0.2.aecf2d9" hw.product="TANDBERG/519">TANDBERG/519 (TC7.0.2.aecf2d9)</example>
+    <example os.version="X12.5.2" hw.product="TANDBERG/4137">TANDBERG/4137 (X12.5.2 (TEST SW))</example>
+    <example os.version="X8.2.1" hw.product="TANDBERG/4130">TANDBERG/4130 (X8.2.1)</example>
+    <example os.version="TC5.1.4.295090" hw.product="TANDBERG/516">TANDBERG/516 (TC5.1.4.295090)</example>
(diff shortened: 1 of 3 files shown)


xml/sip_banners.xml
```diff
@@ -178,19 +178,17 @@
...
   </fingerprint>
 
-  <fingerprint pattern="^(TANDBERG/(\d+)) \((\S+).*\)$">
-    <description>Cisco/Tandberg TelePresence</description>
-    <example os.version="TC7.0.2.aecf2d9" tandberg.model="519" hw.product="TANDBERG/519">TANDBERG/519 (TC7.0.2.aecf2d9)</example>
-    <example os.version="X12.5.2" tandberg.model="4137" hw.product="TANDBERG/4137">TANDBERG/4137 (X12.5.2 (TEST SW))</example>
-    <example os.version="X8.2.1" hw.product="TANDBERG/4130" tandberg.model="4130">TANDBERG/4130 (X8.2.1)</example>
-    <example os.version="XC2.2.1-b2bua-1.0" hw.product="TANDBERG/4353" tandberg.model="4353">TANDBERG/4353 (XC2.2.1-b2bua-1.0)</example>
-    <example os.version="TC5.1.4.295090" hw.product="TANDBERG/516" tandberg.model="516">TANDBERG/516 (TC5.1.4.295090)</example>
-    <example os.version="TCNC5.1.4.295090" hw.product="TANDBERG/517" tandberg.model="517">TANDBERG/517 (TCNC5.1.4.295090)</example>
```
mlaskowski-r7, line 188: I'm wondering if it wouldn't make sense to keep the old examples as well? Looks like they have somewhat different format. This way we would be sure we are backward compatible version-wise. 🤔 Wdyt?

PR #626, opened 2024-09-13
Add Cisco CE fingerprinter

atorovik-r7:
## Description
A detailed description of your changes.
Add fingerprinting for Cisco TelePresence Cisco Collaboration Endpoint. Add new record to recog snmp. Update regex to only detect versions starting with ce.

## Motivation and Context
Explanation of why these changes are being proposed, including any links to other relevant issues or pull requests.

## How Has This Been Tested?
A clear and concise description of your changes were tested.

## Types of changes

## Checklist:

### DIFF
--- a/xml/snmp_sysdescr.xml
+++ b/xml/snmp_sysdescr.xml
@@ -1564,7 +1564,7 @@
     <param pos="2" name="hw.product"/>
   </fingerprint>
 
-  <fingerprint pattern="SoftW:\s+(ce\d+\S*)$">
+  <fingerprint pattern="SoftW:\s+(ce\d+\S*)">
     <description>Cisco TelePresence Collaboration Endpoint (RoomOS)</description>
     <example os.version="ce9.15.0.11.aec227943ed">SNMP:
   Uptime: 840410
--- a/xml/snmp_sysdescr.xml
+++ b/xml/snmp_sysdescr.xml
@@ -1564,7 +1564,7 @@
     <param pos="2" name="hw.product"/>
   </fingerprint>
 
-  <fingerprint pattern="^ce\d+\S*$">
+  <fingerprint pattern="^(ce\d+\S*)$">
     <description>Cisco/Tandberg TelePresence</description>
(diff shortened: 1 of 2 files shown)


atorovik-r7:
I'm a bit confused about the versions, 

in the example ce9.15.0.d97b604b745 : should be extract just ce9.15.0 or the whole string?

xml/snmp_sysdescr.xml
```diff
@@ -1564,6 +1564,20 @@
     <param pos="2" name="hw.product"/>
   </fingerprint>
 
+  <fingerprint pattern="^(ce\d+\S*)$">
```
mlaskowski-r7, line 1567: We may want to try test it out locally. The SNMP output would look like below and it seems to be one string with newlines. If so then we'll need to adjust the regex. 
```
SNMPv2-MIB::sysDescr.0 = STRING: Cisco Codec
SoftW: ce9.15.0.d97b604b745
MCU: Cisco Webex Room Kit
Date: 2020-10-26
S/N: ...
```
A different, simpler, example for Cisco Meraki, which is a one-liner:
`SNMPv2-MIB::sysDescr.0 = STRING: Meraki MX84 Cloud Managed Router`
and similarly complex for Cisco IOS:
```
SNMPv2-MIB::sysDescr.0 = STRING: Cisco IOS Software, C800 Software (C800-UNIVERSALK9-M), Version 15.6(3)M4, RELEASE SOFTWARE (fc1)
Technical Support: http://www.cisco.com/techsupport
Copyright (c) 1986-2018 by Cisco Systems, Inc.
Compiled Mon 05-Feb-18 11:06 by prod_rel_team
```

martinvotruba-r7: Agree, we should make the regular expression more specific. The example also should be updated to contain the actual output of the SNMP sysDescr.

martinvotruba-r7:
I think we should fingerprint the whole version string and leave the dealing with the hash part on the consumer of the result.

martinvotruba-r7 (approved):

PR #662, opened 2026-03-04
Fix nested repeat operator warnings in Ruby 3.3

svm1048:
This PR resolves the nested repeat operator warnings thrown by Ruby 3.3 during fingerprinting. 

Ruby 3.3's regex engine is much stricter and warns when redundant operators are used (e.g., `(?:[a-z]+)?`). I've audited the XML databases and updated these redundant `+)?` patterns to their mathematical equivalent `*)` (zero or more) to silence the warnings while preserving the exact same fingerprinting logic.

Tested locally with `bundle exec rspec` (195,555 examples, 0 failures).

Resolves [rapid7/metasploit-framework issue #20121](https://github.com/rapid7/metasploit-framework/issues/20121) (RegEx Replacement Warning in smb_version)

PR #682, opened 2026-08-23
Fix nested repeat warning

kx7m2qd:
## Description
Fixes the "nested repeat operator '+' and '?' was replaced with '*'" warning
that Ruby emits when compiling the Cisco RV Series fingerprint pattern in
`xml/snmp_sysdescr.xml`.

`(?:[A-Z]+)?` (a non-capturing group containing "one or more" wrapped in
"optional") is redundant with `(?:[A-Z]*)` ("zero or more") - both match the
exact same set of strings. Since it's a non-capturing group, there's no
nil-vs-empty-string capture semantics to worry about here (unlike `(X+)?`
vs `(X*)` inside a *capturing* group, which does change behavior).

## Motivation and Context
This picks up where #662 left off. That PR proposed a blanket `(.+)?` ->
`(.*)` substitution across many patterns in the XML database, but review
found that (a) most of those patterns don't actually produce the warning
under Ruby 3.3, and (b) for capturing groups, `(.+)?` -> `(.*)` isn't
semantically safe - it changes an unmatched capture from `nil` to `""`,
which breaks branching logic elsewhere in the recog-ruby gem (see
rapid7/recog-ruby, separate PR incoming for that).

So rather than reuse that diff, I re-audited every `pattern=` attribute in
`xml/` directly against Ruby's warning output. Only one pattern in the
entire database actually warns - this one. Everything else is left
untouched.

Relates to rapid7/metasploit-framework#20121 (RegEx Replacement Warning in
smb_version) and supersedes #662.

## How Has This Been Tested?
- Wrote a script that loads every `pattern` attribute from all 52 files in
  `xml/` via `Regexp.new` under `ruby -w` and checks stderr for the
  "nested repeat operator" warning. Only `xml/snmp_sysdescr.xml`'s Cisco RV
  pattern triggers it, both before and after cross-checking against Ruby
  3.3.8 behavior reported in review on #662.
- Verified the old and new regex produce identical `MatchData#captures` for
  all examples in the fingerprint (`RV320`, `RV345-K9`) plus a synthetic
  `RV320W` case, confirming the fix doesn't change matching behavior.
- Ran `bin/recog_verify xml/snmp_sysdescr.xml` locally, all examples still
  pass.

## Types of changes

## Checklist:


.gitignore
cdelafuente-r7, line 1: It looks like these changes are specific to your local setup. We generally avoid committing `.gitignore` changes to this repository unless necessary. Could you please revert this file?

dwelch-r7:
@kx7m2qd would you mind squashing your commits into a single one? think some of the commits were made in error/included things that shouldn't be in the git history

kx7m2qd:
@dwelch-r7  its done please review it once

Issue #416, opened 2022-02-09
Enforce all fingerprints to require examples


dabdine:
**Is your feature request related to a problem? Please describe.**
Fingerprint elements in fingerprint XML files should always contain examples. Work has been done in previous PRs to settle many broken `<example>` tags. However, several other `<fingerprint>` XML elements are missing `<example>` elements altogether. We should enforce that examples are always present.

**Describe the solution you'd like**
1. Update all fingerprint files to have an example, even if the data must be mocked (with an XML comment indicating as such).
2. Update fingerprints.xsd after #410 is merged so that `minOccurs` on `<example>` is 1 instead of 0: https://github.com/rapid7/recog/blob/master/xml/fingerprints.xsd#L106. With this simple change, once #410 lands the XML schema validation logic in recog_verify should automatically flag any missing example elements as an error (thus, giving developers immediate feedback & ensuring CI PR builds fail if one is not included).

**Describe alternatives you've considered**
None.

**Additional context**
Clearing out all warnings & errors from the existing XML fingerprint files should help greatly in enforcing strict standards for new fingerprints that are added, and ultimately ensure recog is kept high quality.

PR #685, opened 2026-09-19
Add examples to fixed-literal fingerprints that had none

MRX-72:
recog_verify warns about fingerprints that ship without a test case. Eight of them match a fixed string, so the example is just the literal from the pattern - the same thing existing fingerprints like ^Transmission$ already do with <example>Transmission</example>. This adds those examples.

In http_servers.xml: NetWare HTTP Stack, HTTP/1.0 (HP JetDirect), YAMAHA-RT, JRun Web Server, Xerox_MicroServer/Xerox11, D-Link MiniAVServer. In snmp_sysdescr.xml: Castelle FaxPress Product and the Intel NetportExpress print server. Each pattern is anchored with no captures or metacharacters, so the matching input is exactly the literal between the anchors.

recog_verify passes afterwards with eight fewer "no test cases" warnings and no failures, and recog_standardize leaves the additions as-is.

Relates to #416


MRX-72:
@awanve-r7 friendly ping — this small fingerprint-examples addition is ready for review whenever you have a moment.

